PR #182, opened 2016-01-11
Removed 'expect: false' duplicate

cesarandreslopez:
When running in the docker container:

  mvn -am -pl web/war -Dgrunt.target=development compile

Was getting:

´´´´
Running "eslint:development" (eslint) task
Warning: Cannot read config file: /opt/visallo-source/web/war/src/main/webapp/test/.eslintrc
Error: duplicated mapping key at line 17, column 18:
        expect: false
                     ^ Use --force to continue.

Aborted due to warnings.
´´´´


sfeng88:
This change has already been applied: d0124c2bceb3973478a7bde2c912ce925e23e99d.

Thank you for your contribution!

PR #187, opened 2016-03-03
In Firefox: Fix detail pane not opening for objects selected after an edge was deleted

joeybrk372:
- [ ] @sfeng88 @rygim 
- [x] @srfarley @joeferner @rajanpardipuramv5 
- [ ] @jharwig


web/war/src/main/webapp/js/data/withObjectSelection.js
```diff
@@ -444,7 +444,8 @@ define([], function() {
                             dataRequest('edge', 'store', { edgeIds: edgesToRequest }) :
                             Promise.resolve([])
                         ).then(function(edges) {
-                            edgesById = _.indexBy(_.compact(edges), 'id');
+                            edges = _.compact(edges);
```
EvanOxfeld: What can cause `edges` to be a sparse array? I see we call `dataRequest('edge', 'store')` above -- does this mean that edges stored in the client store are in a sparse array? If yes, why did you opt to compact the result of querying the store rather than maintaining edges in the store as a dense array?

joeybrk372: We were already compacting `edges` to fill `edgesById` it seemed minimally invasive that we would only take the same values for edges as well. Although I agree it does sound like the sparse array is the root cause of the issue. I will take another pass at it once I get some time.

PR #188, opened 2016-03-03
Only install node once using the front end plugin and use that installation for all other npm installs

joeferner:
- [x] @EvanOxfeld @joeybrk372

### DIFF
--- a/web/war/src/main/webapp/Makefile
+++ b/web/war/src/main/webapp/Makefile
@@ -1,14 +1,23 @@
-.PHONY: clean all test
+.PHONY: clean all test version
+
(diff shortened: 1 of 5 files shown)


EvanOxfeld:
LGTM :+1:

web/war/src/main/webapp/Makefile
```diff
@@ -1,14 +1,22 @@
-.PHONY: clean all test
+.PHONY: clean all test version
+
+NODEDIR=~/.m2/repository/node/v4.2.1/node/
```
sfeng88: Any way we can not hardcode the version number for the dir name here?

joeferner: done

PR #190, opened 2016-03-03
Compile Plugin JavaScript

jharwig:
- [x] @EvanOxfeld @joeybrk372

Previously plugin js was passed through untouched. This PR will minify and verify JavaScript on web app startup.

* Plugin files minified (comments/whitespace removed)
* Show error messages in log if bad plugin js found
* Cached result in memory
* Start minify on app start (but don't block web app)
* Source maps

### DIFF
--- a/web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
+++ b/web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
@@ -25,73 +25,77 @@ import java.util.logging.Level;
 
 
 public class JavascriptResourceHandler implements RequestResponseHandler {
-    private static final VisalloLogger LOGGER = VisalloLoggerFactory.getLogger(JavascriptResourceHandler.class);
+    private static final VisalloLogger LOGGER = VisalloLoggerFactory.getLogger(JavascriptResourceHandlerErrorManager.class);
+
     private static final int EXECUTOR_CONCURRENT = 3;
     private static final long EXECUTOR_IDLE_THREAD_RELEASE_SECONDS = 5;
-    private static final ExecutorService compilationExecutor;
+    private static final ThreadPoolExecutor compilationExecutor = new ThreadPoolExecutor(
+            EXECUTOR_CONCURRENT,
+            EXECUTOR_CONCURRENT,
+            EXECUTOR_IDLE_THREAD_RELEASE_SECONDS,
+            TimeUnit.SECONDS,
+            new LinkedBlockingQueue());
 
     static {
-        ThreadPoolExecutor e = (ThreadPoolExecutor) new ThreadPoolExecutor(
-                EXECUTOR_CONCURRENT,
-                EXECUTOR_CONCURRENT,
-                EXECUTOR_IDLE_THREAD_RELEASE_SECONDS,
-                TimeUnit.SECONDS,
-                new LinkedBlockingQueue());
-
-        e.allowCoreThreadTimeOut(true);
-
-        compilationExecutor = e;
+        compilationExecutor.allowCoreThreadTimeOut(true);
     }
 
     private String jsResourceName;
     private String jsResourcePath;
     private boolean enableSourceMaps;
-    private String sourceMap;
-    private String originalJavascript;
-    private String compiledJavascript;
-    private Long compiledLastModified;
-    private Future<String> compilationTask;
+    private Future<CachedCompilation> compilationTask;
+    private volatile CachedCompilation previousCompilation;
 
     public JavascriptResourceHandler(final String jsResourceName, final String jsResourcePath, boolean enableSourceMaps) {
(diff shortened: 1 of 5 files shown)


web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
```diff
@@ -0,0 +1,175 @@
...
+            public String call() throws Exception {
+                return compileIfNecessary();
+            }
+        });
+    }
+
+    @Override
+    public void handle(HttpServletRequest request, HttpServletResponse response, HandlerChain chain) throws Exception {
+
+        if (compilationTask.isDone()) {
```
joeferner: I think you can just call `get()` here, it will wait only if necessary.

web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
```diff
@@ -0,0 +1,175 @@
...
+
+        compilationExecutor = e;
+    }
+
+    private String jsResourceName;
+    private String jsResourcePath;
+    private boolean enableSourceMaps;
+    private String sourceMap;
+    private String originalJavascript;
+    private String compiledJavascript;
```
joeferner: This may need to be `volatile` along with any other variable coming from the `Future`. The alternative and probably nicer way would be to stick all the returned variables into a inner class and return that from the `Future`'s `get` method.

web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    private long getLastModified() throws IOException {
+        URL url = this.getClass().getResource(jsResourceName);
+        return url.openConnection().getLastModified();
+    }
+
+    private String compileIfNecessary() throws IOException {
+        long lastModified = getLastModified();
+        if (compiledLastModified == null || compiledLastModified != lastModified || compiledJavascript == null) {
```
joeferner: should this check be moved to line 71 and resubmit the task. It doesn't look like `compileIfNecessary` ever gets called twice.

web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
```diff
@@ -0,0 +1,175 @@
...
+                sourceMap = sb.toString();
+            }
+
+            return output;
+        }
+        return null;
+    }
+}
+
+class JavascriptResourceHandlerErrorManager extends BasicErrorManager {
```
joeferner: I didn't even know you can do this (two top level classes in the same file), I had to do a little researching and it sounds like you probably shouldn't do this (http://stackoverflow.com/questions/2336692/java-multiple-class-declarations-in-one-file). Can you move this class as a `static` inner class of `JavascriptResourceHandler` or stick it in a new file.

web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
jharwig: I check if it's done, because after it's done i want to check if a recompile is necessary using the last modified date.

web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
jharwig: It get's called on every request (and on init)

web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
jharwig: Yeah, can't have to public top level classes. Sure, I'll change to static inner

web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
jharwig: Okay, yeah that would be nicer.

web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
joeferner: I see now

EvanOxfeld:
While I still question the requirement to compile plugins at runtime, and I'm not sure if we'll have difficulties in the near future when webpack (+ babel?) + uglify bundles and minifies Visallo core at build time and webpack (+ babel?) bundles plugins at build time then closure compiler minifies plugins at runtime, other than @joeferner's suggestions implementation looks good. :+1:

jharwig:
@EvanOxfeld We'll revisit when plugins are compiled with web pack. We might (probably) want to disable this runtime compress somehow. Either by configuration from the plugin or auto detect minified files somehow.

jharwig:
@joeferner Made changes according to comments

PR #193, opened 2016-03-03
Change GraphPropertyWorkerTestBase to be more realistic and use visallo visibility when running the GPW under test

joeferner:
- [x] @EvanOxfeld @joeybrk372


sfeng88:
Was this test failing before this change?

joeferner:
This is a base class currently only used by `com.visallo.imageMetadataExtractor.ImageMetadataGraphPropertyWorkerTest` but I have another use case on my project which also uses it and it requires the authorizations to be more in line with how Visallo actually works.

PR #196, opened 2016-03-04
Setup JMX

rajanpardipuramv5:

### DIFF
--- a/web/server/jetty-server/pom.xml
+++ b/web/server/jetty-server/pom.xml
@@ -27,6 +27,12 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-jmx</artifactId>
+            <version>${jetty.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


sfeng88:
Why do we need this?

srfarley:
@sfeng88 This is very helpful for seeing what's happening internally. We used it to look at session counts, etc. Jetty supplies MBeans, and so do some of our enterprise plugins. Use `jconsole` (included with the JDK) to connect to the local Jetty server.

sfeng88:
So this is something we would only use locally not in a deployed environment?

srfarley:
No, we used it in a deployed Tomcat also. But with Tomcat, you just supply system properties to enable JMX: https://tomcat.apache.org/tomcat-7.0-doc/monitoring.html#Enabling_JMX_Remote

sfeng88:
How would we use this for a deployed jetty? We don't actually use this class when we deploy to Jetty.

srfarley:
For Jetty, you set it up through configuration: http://www.eclipse.org/jetty/documentation/current/jmx-chapter.html

PR #201, opened 2016-03-07
Runtime closure-compile

jharwig:
- [x] @EvanOxfeld @joeybrk372

The previous compile-js PR was reverted since the closure-compiler packages guava inside the jar. This version of guava conflicted with accumulo expecting guava 14. This references a new repackaged closure-compiler that shades the guava (and other) dependencies.

* [x] Add WebConfiguration parameter to disable (on by default / not disabled)
* [x] Travis build and maven release of closure-compiler

### DIFF
--- a/root/pom.xml
+++ b/root/pom.xml
@@ -58,6 +58,7 @@
         <elasticsearch.version>1.4.4</elasticsearch.version>
         <evo.inflector.version>1.0.1</evo.inflector.version>
         <guice.version>3.0</guice.version>
+        <visallo.closure.compiler.version>1.0.0</visallo.closure.compiler.version>
         <hadoop.version>2.6.0-cdh5.4.2</hadoop.version>
         <twill.version>0.4.0-incubating</twill.version>
         <hadoop-test.version>2.6.0-mr1-cdh5.4.2</hadoop-test.version>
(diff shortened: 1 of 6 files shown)


EvanOxfeld:
If you'd like to disable plugin JS minification, add `pluginDevMode=true` to your configuration.

srfarley:
I manually deployed: https://mvn.v5analytics.com/content/groups/public/org/visallo/visallo-closure-compiler/1.0.0/

Travis CI builds without deploying: https://travis-ci.org/v5analytics/closure-compiler

root/pom.xml
```diff
@@ -58,6 +58,7 @@
         <elasticsearch.version>1.4.4</elasticsearch.version>
         <evo.inflector.version>1.0.1</evo.inflector.version>
         <guice.version>3.0</guice.version>
+        <visallo.closure.compiler.version>1.0.0</visallo.closure.compiler.version>
```
EvanOxfeld: @jharwig Would it make sense to lockstep use the same version as Google? It would be faster than looking up the closure compiler version in our pom.

Ref: http://search.maven.org/#search%7Cga%7C1%7Cclosure%20compiler

jharwig: Sure, we could do that

EvanOxfeld:
I'm looking into why closure compiler doesn't like the Google Analytics plugin. If this pull request is merged in the meantime you can get around the GA plugin error with the new config setting.

sfeng88:
Thanks @EvanOxfeld! Do we want to merge this first and then fix the version later?

EvanOxfeld:
@sfeng88 Yes, this is ready to merge.

PR #204, opened 2016-03-08
only push to the work queue repository if there gpws that are interes…

rygim:
- [x] @EvanOxfeld @joeybrk372

…ted (or if we can't tell)


core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
```diff
@@ -248,13 +252,38 @@ public void pushGraphPropertyQueue(
...
+        if(this.graphPropertyRunner == null || !this.graphPropertyRunner.isStarted()){
+            //we are probably on a server and want to submit it to the architecture
+            return true;
+        }
+
+        Property property = element.getProperty(propertyKey, propertyName);
+
+        for(GraphPropertyWorker worker : this.graphPropertyRunner.getAllGraphPropertyWorkers()){
+            try {
+                if (worker.isHandled(element, property)) {
```
sfeng88: We need to also check `isDeleteHandled`, `isHiddenHandled`, and `isUnhiddenHandled` to make sure no GPWs needs to do work.

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
```diff
@@ -80,6 +84,7 @@ public void prepare(User user, GraphPropertyWorkerInitializer repository) {
         setUser(user);
         setAuthorizations(this.userRepository.getAuthorizations(user));
         prepareWorkers(repository);
+        this.getWorkQueueRepository().setGraphPropertyRunner(this);
```
joeferner: Can the `WorkQueueRepository` not just get injected with `GraphPropertyRunner`?

core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
```diff
@@ -248,13 +252,38 @@ public void pushGraphPropertyQueue(
...
+                    return true;
+                }
+            } catch(Throwable t){
+                LOGGER.warn("Error checking to see if workers will handle graph property message", t);
```
joeferner: Should the error case be to submit.  `return true;`

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
rygim: It can't because it would introduce a cyclic dependency https://github.com/v5analytics/visallo/blob/fe97037631c558061341b260c4214e2a63e90d79/core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java#L55

core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
rygim: good call

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
rygim: I do agree though, it looks awful

core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
rygim: See changes

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
```diff
@@ -49,6 +51,7 @@
     private Configuration configuration;
     private VisibilityTranslator visibilityTranslator;
     private AtomicLong lastProcessedPropertyTime = new AtomicLong(0);
```
srfarley: While you're at it, could you please delete `lastProcessedPropertyTime` and its getter method? The method doesn't appear to be used anywhere.

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
joeferner: Ah, I see now. I've been doing this in the past https://github.com/v5analytics/visallo/blob/master/core/core/src/main/java/org/visallo/core/model/user/UserRepository.java#L212 but in this case I think the setter as you have it makes sense since `GraphPropertyRunner` may not even be instantiated when using YARN.

LGTM

srfarley: Since `WorkQueueRepository` is mainly concerned with querying the `GraphPropertyWorker`s, can we pass the list of GPWs to it instead of the `GraphPropertyRunner`? That would break the dependency completely.

joeferner: Or move lines https://github.com/v5analytics/visallo/pull/204/files#diff-d77299e1d847199c39e0c5cca870a27bR270 to https://github.com/v5analytics/visallo/pull/204/files#diff-d77299e1d847199c39e0c5cca870a27bR294 (`WorkQueueRepository` 270-294) to `GraphPropertyRunner` exposed as a `isHandled` method. Then `WorkQueueRepository` doesn't really need to know much of anything on how `GraphPropertyRunner` works.

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
rygim: The if statement got to be too long and a little confusing so I changed it to a traditional branching structure.

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
rygim: @joeferner I like that, I will move it.

@srfarley While we can just do that, I would then need to prepare the graph property workers which would require duplicated code for preparing everything.  I am going to move things as @joeferner suggested which seems to make sense.

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
rygim: and I am not going to change removing the lastProcessedPropertyTime in this pr because that is a breaking api change and needs to be evaluated by each team separately

sfeng88:
@srfarley @rajanpardipuramv5 @joeferner: Are you guys ok with this PR?

srfarley:
:+1:

joeferner:
LGTM, squash it and merge it

PR #205, opened 2016-03-08
Defocus found paths on dismissal of path activity process

joeybrk372:
- [x] @EvanOxfeld @joeybrk372

### DIFF
--- a/web/war/src/main/webapp/js/activity/activity.js
+++ b/web/war/src/main/webapp/js/activity/activity.js
@@ -184,6 +184,16 @@ define([
         this.removeTask = function(taskId) {
             this.removedTasks[taskId] = true;
             var task = this.tasksById[taskId];
+            var handler = handlersByType[task.type];
+
+            if (handler.onRemove) {
+                if (_.isFunction(handler.onRemove)) {
+                    handler.onRemove.call(this);
+                } else {
+                    console.error('handler.onRemove expected a function, instead got: ', handler.onRemove);
+                }
+            }
+
             delete this.tasksById[taskId];
             this.tasks.splice(this.tasks.indexOf(task), 1);
         };
--- a/web/war/src/main/webapp/js/activity/handlers.js
+++ b/web/war/src/main/webapp/js/activity/handlers.js
@@ -40,6 +40,9 @@ define([], function() {
                     });
                 });
             },
+            onRemove: function() {
+                this.trigger('defocusPaths');
+            },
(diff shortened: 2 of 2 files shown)


sfeng88:
Does this work when the edges are not on the graph?

web/war/src/main/webapp/js/activity/activity.js
```diff
@@ -184,6 +184,13 @@ define([
         this.removeTask = function(taskId) {
             this.removedTasks[taskId] = true;
             var task = this.tasksById[taskId];
+            var handler = handlersByType[task.type];
+
+            if (_.has(handler, 'remove')) {
+                var removeHandler = _.bind(handler.remove, this);
```
jharwig: just need to `handler.onRemove.call(this)`

web/war/src/main/webapp/js/activity/activity.js
```diff
@@ -184,6 +184,13 @@ define([
         this.removeTask = function(taskId) {
             this.removedTasks[taskId] = true;
             var task = this.tasksById[taskId];
+            var handler = handlersByType[task.type];
+
+            if (_.has(handler, 'remove')) {
```
jharwig: change to `onRemove`

web/war/src/main/webapp/js/activity/handlers.js
```diff
@@ -40,6 +40,9 @@ define([], function() {
                     });
                 });
             },
+            remove: function() {
```
jharwig: Change to `onRemove`

jharwig:
@sfeng88 Yes, because it's just a call to defocus. This is nicer in that it removes on dismiss instead of waiting for the user to click in the graph to remove (which sends the same event).

web/war/src/main/webapp/js/activity/activity.js
EvanOxfeld: I tend to favor vanilla JS unless the library adds value in some way e.g. `if(handler.remove) {`.

jharwig: Yup, I prefer handler.remove over _.has, but _.isFunction provides some value here.

EvanOxfeld: @jharwig I didn't see your comment when I was writing  -- agreed, `_.isFunction` is a good way to go here as well. I recommend logging an error if `handler.onRemove` exists but is not a function.

PR #211, opened 2016-03-09
add version to visallo pom

sfeng88:
- [ ] @srfarley @rajanpardipuramv5 @joeferner
- [ ] @EvanOxfeld @joeybrk372

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,6 +11,7 @@
     <artifactId>visallo</artifactId>
     <packaging>pom</packaging>
     <name>Visallo</name>
+    <version>1.1-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -11,6 +11,7 @@
     <artifactId>visallo</artifactId>
     <packaging>pom</packaging>
     <name>Visallo</name>
+    <version>1.1-SNAPSHOT</version>
```
srfarley: This shouldn't be necessary. It inherits version from the parent.

sfeng88: We need to add this in so maven release plugin works.

PR #213, opened 2016-03-10
Fix fullscreen audio / video when modelUpdated event occurs

jharwig:
- [ ] @srfarley @rajanpardipuramv5 @joeferner
- [ ] @EvanOxfeld @joeybrk372

Fixes layout component renderer when updating the model and reusing a dom element that was a layout component and will now be a flightjs component.

Also update audio/video to not display player unless properties to play content exist.

### DIFF
--- a/web/war/src/main/webapp/js/detail/audio/audio.js
+++ b/web/war/src/main/webapp/js/detail/audio/audio.js
@@ -19,13 +19,38 @@ define([
         })
 
         this.after('initialize', function() {
-            this.$node.html('<div class="audio-preview"></div>')
-
-            var model = this.attr.model;
-
-            AudioScrubber.attachTo(this.select('previewSelector'), {
-                rawUrl: F.vertex.raw(model)
-            });
+            this.model = this.attr.model;
+            this.on('updateModel', function(event, data) {
+                this.model = data.model;
+                this.render();
+            })
+            this.render();
         });
+
+        this.isAudioReady = function() {
+            return (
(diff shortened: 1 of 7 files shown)


web/war/src/main/webapp/js/detail/audio/audio.js
```diff
@@ -19,13 +19,38 @@ define([
...
+        };
 
-            AudioScrubber.attachTo(this.select('previewSelector'), {
-                rawUrl: F.vertex.raw(model)
-            });
-        });
+        this.render = function() {
+            if (!this.isAudioReady()) {
+                this.rendered = false;
+                return this.$node.empty();
```
EvanOxfeld: Can we display a user friendly message, particularly during audio processing?

web/war/src/main/webapp/js/detail/video/video.js
```diff
@@ -29,36 +29,65 @@ define([
...
-                        self.updateCurrentTranscript(0);
-                    });
+            this.model = this.attr.model;
+            this.render();
+        });
+
+        this.render = function() {
+            if (!this.isVideoReady()) {
+                this.videoRendered = false;
+                return this.$node.empty();
```
EvanOxfeld: Can we display a user friendly message, particularly during processing? This is particularly important if video entities are identified as such prior to the completion of ffmpeg.

web/war/src/main/webapp/js/detail/audio/audio.js
jharwig: I'm not sure if we know that the video is processing. A plugin that processes the video might not be installed

PR #231, opened 2016-03-17
Multiple ontology upload

sfeng88:
- [x] @EvanOxfeld @joeybrk372

Things to be aware of:
- Last ontology upload wins
- Look at the VertexiumOntologyRepositoryTest to see assumptions and what happens when you change/delete certain annotations


core/core/src/main/java/org/visallo/core/model/ontology/OntologyProperties.java
```diff
@@ -36,4 +37,30 @@
     public static final JsonSingleValueVisalloProperty POSSIBLE_VALUES = new JsonSingleValueVisalloProperty("http://visallo.org#possibleValues");
     public static final BooleanSingleValueVisalloProperty DELETEABLE = new BooleanSingleValueVisalloProperty("http://visallo.org#deleteable");
     public static final BooleanSingleValueVisalloProperty UPDATEABLE = new BooleanSingleValueVisalloProperty("http://visallo.org#updateable");
-}
+
+    public static final ImmutableList<String> CHANGEABLE_PROPERTIES_IRI = ImmutableList.of(
```
srfarley: Use `List` as the variable type here because it's a best practice to use the most abstract type possible. In this case, you might consider using `Set` since there should be no repeats and order isn't important.

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepository.java
```diff
@@ -565,37 +568,57 @@ public OntologyProperty addPropertyTo(
...
+        if (boost != null) {
+            OntologyProperties.BOOST.setProperty(builder, boost, VISIBILITY.getVisibility());
+        }
+        if (displayName != null && !displayName.trim().isEmpty()) {
+            OntologyProperties.DISPLAY_NAME.setProperty(builder, displayName.trim(), VISIBILITY.getVisibility());
+        }
+        if (displayType != null && !displayType.trim().isEmpty()) {
+            OntologyProperties.DISPLAY_TYPE.setProperty(builder, displayType, VISIBILITY.getVisibility());
+        }
+        if (propertyGroup != null && !propertyGroup.trim().isEmpty()) {
```
joeferner: What if property group is removed? Should this remove the property?

joeferner: Nevermind I think I see. You call `deleteChangeableProperties` first?  This is going to cause quite a bit of churn and contention if two web servers are starting up.

core/core/src/main/java/org/visallo/core/model/ontology/OntologyProperties.java
```diff
@@ -36,4 +37,30 @@
...
+            DISPLAY_TYPE.getPropertyName(),
```
srfarley: Is it really necessary to call `getPropertyName()` on each of these? Why not make it a collection of `VisalloPropertyBase` and map them to property names when used? It's a one-liner:

`CHANGEABLE_PROPERTIES_IRI.stream().map(VisalloPropertyBase::getPropertyName).collect(Collect.toCollection())

core/plugins/model-vertexium/src/test/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepositoryTest.java
```diff
@@ -0,0 +1,230 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static org.junit.Assert.*;
+
+public class VertexiumOntologyRepositoryTest {
```
srfarley: I'm surprised you don't need `@RunWith(MockitoJUnitRunner.class)`

core/plugins/model-vertexium/src/test/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepositoryTest.java
```diff
@@ -0,0 +1,230 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static org.junit.Assert.*;
+
```
srfarley: I'm surprised you don't need `@RunWith(MockitoJUnitRunner.class)` here.

sfeng88: Tests passed for me.

core/plugins/model-vertexium/src/test/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepositoryTest.java
```diff
@@ -0,0 +1,230 @@
...
+    private static final String TEST_CHANGED_OWL = "test_changed.owl";
+    private static final String TEST_IRI = "http://visallo.org/test";
+    private VertexiumOntologyRepository ontologyRepository;
+    private Authorizations authorizations;
+    private Graph graph;
+    private AuthorizationRepository authorizationRepository;
+    @Mock
+    private Configuration configuration;
+    @Mock
+    protected Injector injector;
```
srfarley: I don't see where `injector` is used.

core/plugins/model-vertexium/src/test/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepositoryTest.java
srfarley: That's good. I just wish I knew what magic was making that happen. I know `injector` isn't used. Is it possible that nothing calls `configuration`?

sfeng88: `injector` was left over code. Need to mock `configuration` to pass into the `VertexiumOntologyRepository` configuration.

core/core/src/main/java/org/visallo/core/model/ontology/OntologyProperties.java
sfeng88: I would rather keep it as is instead of collecting the property names every time we are deleting properties.

I'll change the property name.

PR #233, opened 2016-03-18
throw a better error if timezone is not found

joeferner:

### DIFF
--- a/web/war/src/main/webapp/js/util/formatters.js
+++ b/web/war/src/main/webapp/js/util/formatters.js
@@ -643,8 +643,11 @@ define([
                     withOffsetForDate = Date.now();
                 }
 
-                var momentZone = moment.tz.zone(name),
-                    offset = momentZone.offset(withOffsetForDate) * -1,
+                var momentZone = moment.tz.zone(name);
+                if (!momentZone) {
+                    throw new Error('Could not find timezone "' + name + '"');
+                }
+                var offset = momentZone.offset(withOffsetForDate) * -1,
                     tzInfo = {
                         tzOffset: offset,
                         tzAbbr: momentZone.abbr(withOffsetForDate),


web/war/src/main/webapp/js/util/formatters.js
```diff
@@ -643,8 +643,11 @@ define([
                     withOffsetForDate = Date.now();
                 }
 
-                var momentZone = moment.tz.zone(name),
-                    offset = momentZone.offset(withOffsetForDate) * -1,
+                var momentZone = moment.tz.zone(name);
+                if (!momentZone) {
+                    throw new Error('Could not find timezone "' + name + '"');
```
srfarley: Should we default to UTC instead of throwing an error?

joeferner: I think if you pass in a name you should get an error. If name is null it might be nice to be UTC but I think that might be a little strange, if you want UTC pass 'UTC'.

srfarley: I'm thinking about the user experience here. Are any components going to catch the error and recover gracefully?

joeferner: If we don't throw an error will we ever notice its broken?

srfarley: It looks like the timezone name originally comes from https://www.npmjs.com/package/jstimezonedetect, so I think it's fine to throw an error.

EvanOxfeld: I agree with @joeferner. It's the component's responsibility to validate input and, in the case of a free form timezone input, surface an error.

srfarley: @EvanOxfeld I agree, and it's not free-form in this case, so we should be OK.

srfarley:
@jharwig Is this OK to merge?

PR #235, opened 2016-03-18
Map fixes and new method for radius search

joeybrk372:
New Radius Search Method
- Removed context menu to load results within a radius on right/ctrl+click of the map.
- Geolocation properties now have a "Search by Radius..." extra that will open the map and start radius selection

Bug Fixes
- markers no longer snap back to original geolocation after editing/publishing new geolocation properties
- markers now appear when adding a geolocation property to an already existing workspace vertex
- markers now leave and enter clusters properly on updates
- map now handles case where a geolocation property is deleted


srfarley:
@joeybrk372 I tried out this brach. These are nice improvements. Should we change "Search by Radius" to "Search on Map" or "Search by Radius on Map" to make it more clear that the map will open?

web/war/src/main/webapp/js/map/map.js
```diff
@@ -307,9 +308,34 @@ define([
...
+                    })
+                } else if (feature.id.indexOf(vertex.id) !== -1) {
+                    previousFeatures.push(feature);
+                }
+
+            });
 
-            if (!geoLocations || geoLocations.length === 0) return;
+            if (!geoLocations || geoLocations.length === 0) {
+                if (previousFeatures) {
```
EvanOxfeld: This should be `previousFeatures.length` -- previousFeatures is initialized to `[]` on line 312, and an empty array is truthy.

web/war/src/main/webapp/js/map/map.js
```diff
@@ -380,6 +412,8 @@ define([
 
                                 if (markers.length) {
                                     markers = self.findOrCreateMarkers(map, vertex);
+                                } else if (!markers.length && inWorkspace) {
+                                    markers = self.findOrCreateMarkers(map, vertex);
```
EvanOxfeld: `markers = self.findOrCreateMarkers(map, vertex);` appears 3 different times. Can the assignment of `markers` be simplified?

joeybrk372:
Good point @srfarley, changed to Search on Map

srfarley:
:+1:

srfarley:
@jharwig Are you OK with merging?

PR #240, opened 2016-03-23
Term Mention Fixes

joeybrk372:
- Fix resolve as new not creating a new entity(instead it was adding reference to first resolved term)
- Allow unresolve of published entity(consistent with detected objects)
- Resolve as property dropdown will now teardown after submission instead of being stuck loading
- Picking fullscreen option on resolved property/comment goes to the owning entity.


EvanOxfeld:
Looks good. :+1: With this fix, does unresolving the term mention remove the edge?

joeybrk372:
@EvanOxfeld No, I didn't run across that issue due to my environment setup.

PR #244, opened 2016-03-25
Support property field value changes after the initial rendering

srfarley:
This is necessary when using property fields outside of the details pane.
I also added missing semicolons.


web/war/src/main/webapp/js/fields/withPropertyField.js
```diff
@@ -63,7 +63,12 @@ define([
             this.on('fieldRendered', function handler() {
                 this.off('fieldRendered', handler);
                 rendered();
-            })
+            });
+
+            this.on('setValue', function(event, value) {
```
jharwig: Do you want to `event.stopPropagation` here? Just thinking that for compound fields, and when triggering setValue on a compound child, the event will propagate to the compound and try to call the function with the childs value.

srfarley: Good catch. Done.

PR #245, opened 2016-03-25
enable subclasses to define custom behavior after jetty starts

dsingley:


srfarley:
@dsingley What problem does this solve that `ApplicationBootstrapInitializer` doesn't? This won't be called unless you're running our Jetty class, whereas an `ApplicationBootstrapInitializer` will be invoked when Visallo is deployed to any web server as a "war".

dsingley:
this will allow desktop installations to load data after in-process components are available

srfarley:
@dsingley Were you able to get it working with the `ApplicationBootstrapInitializer`?

PR #247, opened 2016-03-25
enqueue imported RDF .nt elements for GPW

dsingley:
- [x] @EvanOxfeld @joeybrk372

### DIFF
--- a/common/rdf/src/main/java/org/visallo/common/rdf/RdfTripleImportHelper.java
+++ b/common/rdf/src/main/java/org/visallo/common/rdf/RdfTripleImportHelper.java
@@ -10,6 +10,8 @@ import org.vertexium.type.GeoPoint;
 import org.visallo.core.exception.VisalloException;
 import org.visallo.core.model.graph.GraphRepository;
 import org.visallo.core.model.properties.VisalloProperties;
+import org.visallo.core.model.workQueue.Priority;
+import org.visallo.core.model.workQueue.WorkQueueRepository;
 import org.visallo.core.security.VisibilityTranslator;
 import org.visallo.core.user.User;
 import org.visallo.core.util.VisalloDate;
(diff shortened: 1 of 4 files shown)


common/rdf/src/main/java/org/visallo/common/rdf/RdfTripleImportHelper.java
```diff
@@ -115,9 +144,17 @@ private void importRdfTriple(
 
             ++lineNum;
         }
+
+        if (workQueueRepository == null) {
```
joeferner: When would this happen?

common/rdf/src/main/java/org/visallo/common/rdf/RdfTripleImportHelper.java
```diff
@@ -69,9 +69,35 @@ public RdfTripleImportHelper(
         this.visibilityTranslator = visibilityTranslator;
     }
 
+    @Inject
+    public void setWorkQueueRepository(WorkQueueRepository workQueueRepository) {
```
joeferner: Why isn't this part of the constructor?

common/rdf/src/main/java/org/visallo/common/rdf/RdfTripleImportHelper.java
```diff
@@ -69,9 +69,35 @@ public RdfTripleImportHelper(
...
+        this.workQueueRepository = workQueueRepository;
+    }
+
+    @Deprecated
+    public void importRdfTriple(
```
joeferner: Who uses this? Can it be removed?

joeferner:
Should this be optional?

 If 5 lines contribute to a vertex should we somehow wait to queue until they all go in? 

Graph flush needs to be called before queuing which is going to effect performance.

common/rdf/src/main/java/org/visallo/common/rdf/RdfTripleImportHelper.java
srfarley: :+1: I think there should always be a queue present.

joeferner:
LGTM after the changes

PR #258, opened 2016-04-04
Enhanced "Possible Values" property search filter

srfarley:
- [x] @joeferner @mwizeman 

This changes the search filter for properties defined with "visallo:possibleValues" in the ontology such that more than one of the possible values can be selected in the UI. This executes an "IN" query in Vertexium.

restrictvaluesmulti

### DIFF
--- a/web/war/src/main/webapp/js/fields/restrictValuesMulti.js
+++ b/web/war/src/main/webapp/js/fields/restrictValuesMulti.js
@@ -1,0 +1,106 @@
+define([
+    'flight/lib/component',
+    './restrictValues',
+    'util/vertex/formatters',
+    './withPropertyField',
+    'hbs!./restrictValuesMultiTpl'
+], function(defineComponent, RestrictValues, F, withPropertyField, template) {
+    'use strict';
+
+    return defineComponent(RestrictValuesMultiField, withPropertyField);
+
+    function RestrictValuesMultiField() {
+
+        function compare(v1, v2) {
+            return v1.toLowerCase().localeCompare(v2.toLowerCase());
+        }
+
+        this.defaultAttrs({
+            removeSelector: 'ul.values li .remove-value'
+        });
+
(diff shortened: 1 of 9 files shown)


srfarley:
Note there seems to be a Vertexium bug with the total hits value being returned: https://github.com/v5analytics/vertexium/issues/73

web/war/src/main/webapp/js/fields/restrictValuesMulti.js
```diff
@@ -0,0 +1,104 @@
...
+                propertyId: this.attr.property.title,
+                value: this.getValue(),
+                metadata: _.isFunction(this.getMetadata) && this.getMetadata() || {},
+                options: {}
+            });
+        };
+
+        this.setValue = function(value) {
+            var self = this,
+                sortedValues = _.isArray(value) ? value.sort() : (value && value.length > 0 ? [value] : []);
```
EvanOxfeld: I find nested ternary expressions difficult to read. What's the intention of the inner ternary? To operate on strings?

srfarley: Yes, it covers the case where a single string is passed. I'm going to investigate whether or not that's actually needed now.

srfarley: There's no need to accept anything expect arrays, so I simplified.

web/war/src/main/webapp/js/fields/restrictValuesMulti.js
```diff
@@ -0,0 +1,102 @@
...
+                options: {}
+            });
+        };
+
+        this.setValue = function(values) {
+            var self = this;
+            this.values = [];
+            this.$node.find('ul.values').empty();
+
+            values.sort().forEach(function(value) {
```
jharwig: This case sensitive sort. Use _sortBy

PR #269, opened 2016-04-11
Fix header image disappearing after modifying properties

joeybrk372:

### DIFF
--- a/web/war/src/main/webapp/js/detail/headerImage/image.js
+++ b/web/war/src/main/webapp/js/detail/headerImage/image.js
@@ -38,6 +38,7 @@ define([
             this.on('filecomplete', this.onUploadComplete);
             this.on('fileerror', this.onUploadError);
             this.on(document, 'verticesUpdated', this.onVerticesUpdated);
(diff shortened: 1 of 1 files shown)


web/war/src/main/webapp/js/detail/headerImage/image.js
```diff
@@ -348,5 +349,17 @@ define([
             }
             this.$node.removeClass('uploading');
         };
+
+        this.onUpdateModel = function(event, data) {
+           var hasCustomImage = _.some(data.model.edgeLabels, function(label) {
+               return label === 'http://visallo.org/dev#entityHasImageRaw'
```
EvanOxfeld: This can be simplified with [`_.contains`](http://underscorejs.org/#contains). (Or when we add transpilation from ES6 to ES5 to Visallo, we can use the built-in [Array.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes).)

srfarley:
@jharwig Is this OK to merge?

PR #277, opened 2016-04-15
Fix IE11 loading issue when trying to do feature detection of XmlHttp…

joeferner:
- [x] @srfarley @mwizeman 

…Request upload feature.  IE11 does not like "if (xmlhttpreq.upload)" syntax

After making this change Visallo comes up and is somewhat usable in IE11


joeferner:
Microsoft Bug: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/107521/

EvanOxfeld:
Looks good to me until we fix uploading in general in IE11. Longer term we may want to console log once rather than always swallow these errors.

PR #283, opened 2016-04-19
Fix date picker field changing date on backspace

joeybrk372:


web/war/src/main/webapp/js/fields/date.js
```diff
@@ -63,23 +63,36 @@ define([
                 self.triggerFieldUpdated();
             });
 
+            this.on('keydown', function(event) {
+                if (event.which === 8) {
```
srfarley: What does "8" mean? Is there a constant defined anywhere?

joeybrk372: 8 is the keycode for backspace. Could potentially use `$.ui.keyCode.BACKSPACE` here, I made this to be consistent with other `event.which === ` instances in the component

joeferner: Could we at least have a

```
var KEY_BACKSPACE = 8;
```

at the top

PR #284, opened 2016-04-19
If server responds with 200 and an empty string don't treat it as an error

joeferner:

### DIFF
--- a/web/war/src/main/webapp/js/data/web-worker/util/ajax.js
+++ b/web/war/src/main/webapp/js/data/web-worker/util/ajax.js
@@ -80,9 +80,9 @@ define(['util/promise'], function(Promise) {
                     try {
                         r.upload.removeEventListener('progress', progressHandler);
                     } catch(e) {}
-                    var text = r.status === 200 && r.responseText;
 
-                    if (text) {
+                    if (r.status === 200) {
+                        var text = r.responseText;
                         if (isJson) {
                             try {
                                 var json = JSON.parse(text);


web/war/src/main/webapp/js/data/web-worker/util/ajax.js
```diff
@@ -80,9 +80,9 @@ define(['util/promise'], function(Promise) {
...
                         r.upload.removeEventListener('progress', progressHandler);
                     } catch(e) {}
-                    var text = r.status === 200 && r.responseText;
 
-                    if (text) {
+                    if (r.status === 200) {
+                        var text = r.responseText;
                         if (isJson) {
                             try {
                                 var json = JSON.parse(text);
```
srfarley: Is it possible for `parse` to throw an error or return a non-JSON result if `text` doesn't contain valid JSON?

joeferner: Yes parse will throw errors on anything non json

srfarley: If the server returns an empty string (with status 200), then `text` doesn't contain valid JSON and `parse` will throw an error. Isn't that what this PR is trying to avoid? What am I missing?

joeferner: Sorry no, this pr fixes empty string (non json)

EvanOxfeld:
What's the use case for returning a 200 with no content? Typically if there's no content the server responds with a 204 status code and a get or a post returns content.

joeferner:
Get property value. Maybe this should accept all 2xx response codes?

PR #287, opened 2016-04-21
add a registerJavaScript call to skip closure compilation for misbehaving js files

joeferner:
- [x] @srfarley @mwizeman 

We are using quill.js in a plugin which has closure compiler errors. This PR allows us to skip closure compiling for that particular js file.

It also fixes compiler warnings in `JavascriptResourceHandler`

### DIFF
--- a/web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
+++ b/web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
@@ -20,15 +20,12 @@ import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Enumeration;
 import java.util.List;
 import java.util.concurrent.*;
 import java.util.logging.Level;
 
 
 public class JavascriptResourceHandler implements RequestResponseHandler {
-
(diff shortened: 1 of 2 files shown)


web/web-base/src/main/java/org/visallo/web/JavascriptResourceHandler.java
```diff
@@ -59,12 +56,7 @@ public JavascriptResourceHandler(final String jsResourceName, final String jsRes
         this.enableSourceMaps = enableSourceMaps;
         this.closureExternResourcePath = closureExternResourcePath;
 
-        compilationTask = compilationExecutor.submit(new Callable<CachedCompilation>() {
-            @Override
-            public CachedCompilation call() throws Exception {
-                return compileIfNecessary(null);
-            }
-        });
+        compilationTask = compilationExecutor.submit(() -> compileIfNecessary(null));
```
srfarley: 👍 lambda

EvanOxfeld:
This change allows plugin authors to circumvent closure compiler. Is that something we want to allow? Would your plugin satisfy closure compiler if you specified an extern file? /cc @jharwig

jharwig:
Yes, providing an extern file would make it work. Probably uses some inferred global. Just need an externs file with `var Global;` in it. 

But we might also want this feature. To prevent double minify for some plugins.

PR #288, opened 2016-04-21
Export/Restore User Password Salt and Hash

mwizeman:

In the event that we have to restore user accounts, this new UserAdmin tool will allow us to restore user passwords properly. The `export-passwords` command will print a table with username, password salt, and password hash to the screen. These value can then be fed back into the `update-password` command in order to restore a user's password to its previous value.


core/core/src/main/java/org/visallo/core/model/user/UserRepository.java
```diff
@@ -93,6 +93,8 @@ protected UserRepository(
 
     public abstract void setPassword(User user, String password);
```
joeferner: Should we make this not abstract and call `setPassword(User, byte[], byte[])` with the correct info and remove the implementations in the derived classes?

tools/user-admin/src/main/java/org/visallo/tools/UserAdmin.java
```diff
@@ -67,6 +69,9 @@ protected JCommander parseArguments(String[] args) {
             case CMD_ACTION_SET_DISPLAYNAME_EMAIL:
                 this.args = new SetDisplayNameEmailArgs();
                 break;
+            case CMD_ACTION_EXPORT_PASSWORDS:
```
joeferner: can we make this export users and export all known fields including privileges?

joeferner:
how do I import users?

core/core/src/main/java/org/visallo/core/model/user/UserRepository.java
mwizeman: Fixed with 4742f445170aafbf9b028a859dbc304f8e91844a

tools/user-admin/src/main/java/org/visallo/tools/UserAdmin.java
mwizeman: If you use `UserAdmin list -t`, you'll get a table view of the user data minus the password and authorizations fields. That combined with the output of this new tool will give you the list you're looking for.

tools/user-admin/src/main/java/org/visallo/tools/UserAdmin.java
```diff
@@ -137,17 +166,7 @@ private int create(CreateUserArgs args) {
...
-        List<User> sortedUsers = new ArrayList<>();
-        while (true) {
-            List<User> users = toList(getUserRepository().find(skip, limit));
-            if (users.size() == 0) {
-                break;
-            }
-            sortedUsers.addAll(users);
-            skip += limit;
-        }
+        List<User> sortedUsers = loadUsers();
```
srfarley: Use `sortedUsers.stream().sorted(...)`

https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#sorted-java.util.Comparator-

tools/user-admin/src/main/java/org/visallo/tools/UserAdmin.java
```diff
@@ -379,6 +385,29 @@ private int maxWidth(Object o, int max) {
...
 
+    private List<User> loadUsers() {
+        return loadUsers(null);
+    }
+
+    private List<User> loadUsers(UserStatus filter) {
+        int skip = 0;
+        int limit = 100;
+        List<User> allUsers = new ArrayList<>();
+        while (true) {
```
srfarley: This can be done more succinctly with streams. See `org.visallo.core.util.StreamUtil` to easily convert `Iterable` to `Stream`.

tools/user-admin/src/main/java/org/visallo/tools/UserAdmin.java
mwizeman: Fixed in 6d05d3f6cc51742c280748a13a0308fc40fa9744

tools/user-admin/src/main/java/org/visallo/tools/UserAdmin.java
mwizeman: Refactored a bit to simplify it, but we can't convert the iterables to a stream since we don't know how many users there are. In order for us to do that, we need a countUsers method on the UserRepository.

PR #289, opened 2016-04-21
Redesign properties to be more data dense

jharwig:

*Changes:*
* Move visibility to right (if enough space)
* Hide info button until hover
* Move info button to in between name and value since it's hidden, they will all be in the same place
* Fix property info edge cases when no metadata, and better button formatting.
* Fix new property info popovers not showing correct buttons when privileges change

info
diff


web/war/src/main/webapp/js/detail/properties/properties.js
```diff
@@ -25,7 +25,7 @@ define([
         NO_GROUP = '${NO_GROUP}',
 
         // Property td types
-        GROUP = 0, NAME = 1, VALUE = 2,
+        GROUP = 0, NAME = 1, VALUE = 2, HIDDEN_COLLAPSE,
```
srfarley: Does `HIDDEN_COLLAPSE` need a value?

jharwig: Good catch, was working because it was just matching on undefined

PR #293, opened 2016-04-26
Don't block editing to original workspace name

joeybrk372:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley


web/war/src/main/webapp/js/workspaces/form/form.js
```diff
@@ -223,7 +223,8 @@ define([
                 return;
             }
 
-            if (_.contains(this.attr.workspaceTitlesLowercase, val.toLowerCase())) {
+            if (_.contains(this.attr.workspaceTitlesLowercase, val.toLowerCase()) &&
+                val.toLowerCase() !== this.attr.currentWorkspaceTitle) {
```
jharwig: shouldn't it be `this.attr.currentWorkspaceTitle.toLowerCase()`?

PR #297, opened 2016-04-27
Fix vertex history route to handle StreamingPropertyValues

joeferner:
- [x] @srfarley @joeferner
- [ ] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley

### DIFF
--- a/core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
+++ b/core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
@@ -5,6 +5,7 @@ import org.vertexium.property.StreamingPropertyValue;
 import org.vertexium.type.GeoPoint;
 import org.vertexium.type.GeoRect;
 import org.vertexium.util.IterableUtils;
+import org.visallo.core.exception.VisalloException;
 import org.visallo.core.ingest.video.VideoFrameInfo;
 import org.visallo.core.ingest.video.VideoPropertyHelper;
 import org.visallo.core.model.properties.MediaVisalloProperties;
@@ -13,12 +14,15 @@ import org.visallo.core.model.workspace.Dashboard;
 import org.visallo.core.model.workspace.DashboardItem;
 import org.visallo.web.clientapi.model.*;
 
+import java.io.IOException;
+import java.io.InputStream;
 import java.util.*;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
 public class ClientApiConverter extends org.visallo.web.clientapi.util.ClientApiConverter {
     public static final EnumSet<FetchHint> SEARCH_FETCH_HINTS = EnumSet.of(FetchHint.PROPERTIES, FetchHint.PROPERTY_METADATA, FetchHint.IN_EDGE_LABELS, FetchHint.OUT_EDGE_LABELS);
+    private static final int HISTORICAL_PROPERTY_MAX_SPV_SIZE = 2000;
 
     public static ClientApiTermMentionsResponse toTermMentionsResponse(Iterable<Vertex> termMentions, String workspaceId, Authorizations authorizations) {
         ClientApiTermMentionsResponse termMentionsResponse = new ClientApiTermMentionsResponse();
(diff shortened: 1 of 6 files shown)


core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -202,7 +208,11 @@ public static ClientApiHistoricalPropertyResults toClientApi(Iterable<Historical
         for (Metadata.Entry entry : hpv.getMetadata().entrySet()) {
             result.metadata.put(entry.getKey(), toClientApiValue(entry.getValue()));
         }
-        result.value = toClientApiValue(hpv.getValue());
+        Object value = hpv.getValue();
+        if (value instanceof StreamingPropertyValue) {
+            value = ((StreamingPropertyValue) value).readToString();
```
srfarley: If this is a very large value, do we still want to read the whole string?

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -191,6 +191,12 @@ private static ClientApiProperty findProperty(List<ClientApiProperty> clientApiP
     public static ClientApiHistoricalPropertyResults toClientApi(Iterable<HistoricalPropertyValue> historicalPropertyValues) {
         ClientApiHistoricalPropertyResults result = new ClientApiHistoricalPropertyResults();
         for (HistoricalPropertyValue historicalPropertyValue : historicalPropertyValues) {
+            if (historicalPropertyValue.getValue() instanceof StreamingPropertyValue) {
+                StreamingPropertyValue spv = (StreamingPropertyValue) historicalPropertyValue.getValue();
+                if (spv.getValueType() != String.class) {
+                    continue;
+                }
+            }
             result.events.add(toClientApi(historicalPropertyValue));
```
srfarley: Will the property be completely excluded from history if it's not a string? If so, can we put a placeholder in for value, something like "Non-displayable data (1234 bytes)"?

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
joeferner: I guess we can choose an arbitrary cut off?

srfarley: Yes, but we'll have to either add a `readToString(int max)` method to `StreamingPropertyValue` or handle the `InputStream` here.

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
joeferner: done

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
joeferner: done

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -202,13 +206,38 @@ public static ClientApiHistoricalPropertyResults toClientApi(Iterable<Historical
...
         result.propertyName = hpv.getPropertyName();
         result.propertyVisibility = hpv.getPropertyVisibility().getVisibilityString();
         return result;
     }
 
+    private static String readStreamingPropertyValueForHistory(StreamingPropertyValue spv) {
+        if (spv.getValueType() == String.class) {
+            return readStreamingPropertyValueStringForHistory(spv);
+        } else {
+            return String.format("Non-displayable data (%d bytes)", spv.getLength());
```
srfarley: I think we should i18n this message.

joeferner: If we want that to happen we need to come up with a better way to handle i18n on the backend I really don't want to pass locale or user all over the place

On April 27, 2016 5:10:09 PM EDT, Steven Farley <notifications@github.com> wrote:
>>          result.propertyKey = hpv.getPropertyKey();
>>          result.propertyName = hpv.getPropertyName();
>>          result.propertyVisibility =
>hpv.getPropertyVisibility().getVisibilityString();
>>          return result;
>>      }
>>  
>> +    private static String
>readStreamingPropertyValueForHistory(StreamingPropertyValue spv) {
>> +        if (spv.getValueType() == String.class) {
>> +            return readStreamingPropertyValueStringForHistory(spv);
>> +        } else {
>> +            return String.format("Non-displayable data (%d bytes)",
>spv.getLength());
>
>I think we should i18n this message.
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/v5analytics/visallo/pull/297/files/da9e874080b7b07e4df39afae2d21a178e3a1bea#r61336884

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

srfarley: In this case, whatever additional parameter we add would be limited to this method and the two `toClientApi` methods above it. Those are called directly by routes.

joeferner: done

PR #303, opened 2016-04-29
Add option for popovers to designate whether they stay in view

joeybrk372:
Set vertex preview popover to stay with node by default.

- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley


web/war/src/main/webapp/js/util/popovers/withPopover.js
```diff
@@ -158,6 +158,7 @@ define([], function() {
                     windowHeight = $(window).height(),
                     maxLeft = windowWidth - width,
                     maxTop = windowHeight - height,
+                    keepInView = this.attr.keepInView === undefined ? true : this.attr.keepInView,
```
EvanOxfeld: Would it be cleaner to initialize `this.attr.keepInView` to true in `this.defaultAttrs`?

PR #306, opened 2016-05-01
Fix MessagingFilter NPE

EvanOxfeld:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

Fixes null pointer exception preventing the session expiration message from reaching the client.

v5analytics/visallo#53

### DIFF
--- a/web/web-base/src/main/java/org/visallo/web/MessagingFilter.java
+++ b/web/web-base/src/main/java/org/visallo/web/MessagingFilter.java
@@ -13,12 +13,13 @@ import org.visallo.core.util.VisalloLogger;
 import org.visallo.core.util.VisalloLoggerFactory;
 
 import javax.servlet.http.HttpSession;
-import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
 public class MessagingFilter implements PerRequestBroadcastFilter {
     private static final VisalloLogger LOGGER = VisalloLoggerFactory.getLogger(MessagingFilter.class);
+    public static final String TYPE_SESSION_EXPIRATION = "sessionExpiration";
+    public static final String TYPE_SET_ACTIVE_WORKSPACE = "setActiveWorkspace";
     private UserRepository userRepository;
 
     @Override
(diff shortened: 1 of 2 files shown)


EvanOxfeld:
Should `MessagingFilter` be unit tested?

srfarley:
Yes, please unit test if possible.

web/web-base/src/main/java/org/visallo/web/MessagingFilter.java
```diff
@@ -36,48 +37,102 @@ public BroadcastAction filter(String broadcasterId, AtmosphereResource r, Object
...
+        if (type != null) {
+            switch (type) {
+                case TYPE_SET_ACTIVE_WORKSPACE:
+                    return false;
+                case TYPE_SESSION_EXPIRATION:
+                    return true;
+            }
+        }
+
+        if (session == null) {
```
srfarley: `return session != null && !shouldRejectMessageByPermissions(json, session);`

EvanOxfeld: @srfarley Done. Changed to `return session != null && shouldSendMessageByPermissions(json, session);`

joeferner: I don't particularly like this change. These two tests are for different reasons. Combining them just makes it harder to read in my opinion.

EvanOxfeld: @srfarley @joeferner: How about a compromise?

```
if (session == null) {
  return false;
}

return shouldSendMessageByPermissions(json, session);
```

srfarley: It's a matter of style. I usually like to combine conditions with two terms, and don't find it harder to read.

srfarley: It doesn't matter to me. I'm not going to hold up this PR over this.

joeferner: No need to hold up the PR I just think these two statements have different purposes and just as I would split out two classes to handle responsibilities I would split out these statements as well if on the other hand these two statements were working together then I would join them it feels like a run on sentence to me.

EvanOxfeld: Merging with the compromise code snippet above. That said, I defer to you all on Java code style if you would like to change anything.

PR #310, opened 2016-05-02
Web: Initial public JS API

EvanOxfeld:
- [x] @srfarley @joeferner
- [ ] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley

Exposes an API module with synchronous functions for registering Visallo extensions and defining flight components as well as a `connect` function that returns a promise with formatter functions and public built-in flight components. `connect` is purposefully asynchronous so as not to hide from the developer that asynchronous actions such as loading the ontology are performed. If this looks along the right lines I'll update the developer docs.

Here's are a couple requirejs examples:

```js
define(['public/v1/api'], function(api) {
  var registry = api.registry;
  registry.registerExtension([extension point name], [extension point object]);
});
```

```js
define(['public/v1/api'], function(api) {
  var defineComponent = api.defineComponent;
  return api
    .connect()
    .then(function(visallo) {
      var withDataRequest = visallo.withDataRequest;
      return defineComponent(MyDataRequestingComponent, withDataRequest);
    });

  function MyDataRequestingComponent() {
  }
});
```

### DIFF
--- a/web/war/src/main/webapp/js/public/v1/api.js
+++ b/web/war/src/main/webapp/js/public/v1/api.js
@@ -1,0 +1,52 @@
+define([
+    'flight/lib/component',
+    'configuration/plugins/registry'
+], function(
+    defineComponent,
+    registry) {
+    'use strict';
+
+    return {
+        connect: connect,
+        defineComponent: defineComponent,
+        registry: registry
+    };
+
+    function connect() {
+        return Promise.all([
+            'util/element/list',
+            'util/ontology/conceptSelect',
+            'util/ontology/propertySelect',
+            'util/ontology/relationshipSelect',
+            'util/vertex/formatters',
+            'util/vertex/justification/viewer',
+            'util/visibility/edit',
+            'util/visibility/view',
+            'util/withDataRequest'
+        ].map(function(module) {
+            return Promise.require(module);
+        })).spread(function(
+            List,
+            ConceptSelector,
+            PropertySelector,
+            RelationshipSelector,
+            F,
+            JustificationViewer,
+            VisibilityEditor,
+            VisibilityViewer,
+            withDataRequest) {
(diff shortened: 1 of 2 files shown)


srfarley:
How did you determine the components exposed by the `api` object?

srfarley:
Should we have `visallo` in the component path, e.g. `visallo/public/v1/api`?

EvanOxfeld:
@srfarley: @jharwig suggested the components. Let me know if there are any other components you think we should expose on the API.

I'm fine either way with including `visallo` in the module path, though the more that I think about it, the API module should be a separate UMD distribution (supports just about every module type including requirejs, ES6, and commonjs/node) so as to not force requirejs onto web plugin authors.

srfarley:
I'd like to change `withDataRequest` to `dataRequest`, and have that point directly to the function. There's no need to expose `withDataRequest` as a FlightJS mixin.

EvanOxfeld:
This is also a good opportunity to tweak the `withDataRequest` / `dataRequest` API to reject promises with an `Error` instance, which is the type of object expected (hence the bluebird warning). I'm leaning towards attaching the fields on the existing custom object, `status` and `statusText`, to the `Error`, even though the two fields are hardly used.

jharwig:
Yup, don't need to give a flight mixin, I'm okay just passing the withDataRequest.dataRequest function.

Probably need a web worker api too. `public/v1/worker-api/` or something with the `util/ajax` and maybe the store?

Otherwise it's a good start for support v1 api.

EvanOxfeld:
Removed the flight mixin and added an initial `workerApi.js` with `util/ajax`. I'm hesitant to provide direct access to the store because of the potential to break the world. Let's have that discussion outside of the scope of this pull request.

PR #312, opened 2016-05-05
Make it much easier to create a default WAR file

kunklejr:
This task is intended to make it much easier to create a standalone WAR file that mimics what is started by the Maven command in the docs describing running. 

Please note that the dependency exclusion in the modified POM does not work, despite all documentation I'm able to find suggesting that it should. It's failure has no bearing on the operation of the web app, but it does end up including jetty JAR files in the WAR.

- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

### DIFF
--- a/config/visallo.properties
+++ b/config/visallo.properties
@@ -66,3 +66,10 @@ security.visibilityTranslator=org.visallo.core.security.DirectVisibilityTranslat
 # org.visallo.core.formula.FormulaEvaluator
 #
 #org.visallo.core.formula.FormulaEvaluator.max.threads=1
+
+
+# Uncomment to allow plain HTTP. HTTPS is forced otherwise.
+#http.transportGuarantee=NONE
+
+# Uncomment when deploying to Tomcat
+#http.gzipEnabled=false
--- a/config/x-visallo.properties
+++ b/config/x-visallo.properties
@@ -30,10 +30,3 @@ graph.sql.jdbc.password=${sql.password}
 graph.sql.jdbc.driverClass=${sql.driverClass}
 graph.sql.jdbc.url=${sql.connectionString}
 graph.tableNamePrefix=${sql.tablePrefixName}
-
-
-# Uncomment to allow plain http
-http.transportGuarantee=NONE
-
-# Uncomment when deploying to Tomcat
-http.gzipEnabled=false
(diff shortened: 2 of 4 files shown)


config/visallo.properties
```diff
@@ -66,3 +66,10 @@ security.visibilityTranslator=org.visallo.core.security.DirectVisibilityTranslat
 # org.visallo.core.formula.FormulaEvaluator
 #
 #org.visallo.core.formula.FormulaEvaluator.max.threads=1
+
+
+# Uncomment to allow plain HTTP. HTTPS is forced otherwise.
+#http.transportGuarantee=NONE
```
srfarley: Shouldn't NONE be the default since setting up SSL, for both Jetty and Tomcat, is an impediment to getting started quickly?

config/visallo.properties
```diff
@@ -66,3 +66,10 @@ security.visibilityTranslator=org.visallo.core.security.DirectVisibilityTranslat
...
+
+# Uncomment when deploying to Tomcat
+#http.gzipEnabled=false
```
srfarley: `false` should be the default - it's one less step to get up and running with Tomcat. For Jetty, it's relatively easy for users to enable gzip compression if performance is an issue:
- http://www.eclipse.org/jetty/documentation/9.1.2.v20140210/gzip-filter.html
- http://www.eclipse.org/jetty/documentation/9.1.2.v20140210/override-web-xml.html

config/visallo.properties
dsingley: Both jetty and tomcat provide documentation for configuring HTTPS, we should be secure by default.

dsingley: Particularly because this was only a problem for Jeff when deploying to a container. We have the run from maven solution for first time users.

config/visallo.properties
EvanOxfeld: Is there any downside to turning on gzip compression? Who's the consumer of this WAR file -- us internally or a customer that will need to be knowledgable about Visallo's configuration?

srfarley: The customer is the user.  The downside is that Tomcat doesn't work if we enable our own compression

config/visallo.properties
kunklejr: We can always change defaults later, but can we get this PR pushed through?

PR #313, opened 2016-05-05
fix: add metadata (modified date, modified by, and visibility json) to justification to fix publish

joeferner:
- [x] @srfarley
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley


core/core/src/main/java/org/visallo/core/model/graph/GraphRepository.java
```diff
@@ -201,26 +201,29 @@ public Vertex addVertex(
             vertexBuilder = graph.prepareVertex(visalloVisibility.getVisibility());
         }
         VisalloProperties.VISIBILITY_JSON.setProperty(vertexBuilder, visibilityJson, visalloVisibility.getVisibility());
-        Metadata propertyMetadata = new Metadata();
+        Metadata conceptTypeMetadata = new Metadata();
```
srfarley: I kind of like the prior variable name better, since this metadata is for more that just the concept type property.

joeferner: It's actually not. From what I can tell it's only for concept type right now

srfarley: Sorry, I misread. You're right.

PR #317, opened 2016-05-05
Detect Tomcat container and do not enable gzip filter

kunklejr:
This simple pull request attempts to detect when the web app is running under tomcat and prevent the gzip filter from being applied (it breaks Tomcat). This is important since the default config is to enable the filter. **The safer alternative to this PR is to make the default `false` and only enable it when you know it will work, like when running Jetty.** However, that is a change in the default setting that could impact current users.

- [x] @srfarley @joeferner
- [ ] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

### DIFF
--- a/config/visallo.properties
+++ b/config/visallo.properties
@@ -70,3 +70,7 @@ security.visibilityTranslator=org.visallo.core.security.DirectVisibilityTranslat
 
 # Uncomment to allow plain HTTP. HTTPS is forced otherwise.
 #http.transportGuarantee=NONE
+
+# Enable/Disable Visallo Gzip filter (on Jetty the default is true, other containers default to false)
+#  The Visallo Gzip filter is known to have issues on Tomcat containers.
+#http.gzipEnabled=false
--- a/config/visallo.properties
+++ b/config/visallo.properties
@@ -70,6 +70,3 @@ security.visibilityTranslator=org.visallo.core.security.DirectVisibilityTranslat
 
 # Uncomment to allow plain HTTP. HTTPS is forced otherwise.
 #http.transportGuarantee=NONE
-
-# Uncomment when deploying to Tomcat
-#http.gzipEnabled=false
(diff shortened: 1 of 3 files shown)


web/web-base/src/main/java/org/visallo/web/ApplicationBootstrap.java
```diff
@@ -234,6 +234,13 @@ private void addCacheFilter(ServletContext context) {
         }
     }
 
+    private boolean shouldAddGzipFilter(Configuration config) {
+        if (System.getProperty("catalina.home") != null || System.getProperty("catalina.base") != null) {
```
srfarley: Would it be better to use `ServletContent#getServerInfo()`. It should return a string like "Apache Tomcat/7".

web/web-base/src/main/java/org/visallo/web/ApplicationBootstrap.java
```diff
@@ -234,6 +234,15 @@ private void addCacheFilter(ServletContext context) {
         }
     }
 
+    private boolean shouldAddGzipFilter(ServletContext context, Configuration config) {
+        if (System.getProperty("catalina.base") != null ||
+                System.getProperty("catalina.home") != null ||
+                (context.getServerInfo() != null && context.getServerInfo().toLowerCase().contains("tomcat"))) {
```
dsingley: can we not just only use `getServerInfo()`?

srfarley: We were unsure if that was set when starting Tomcat programmatically, using https://tomcat.apache.org/tomcat-7.0-doc/api/org/apache/catalina/startup/Tomcat.html

kunklejr: Perhaps `getServerInfo()` would be sufficient, but it's hard to say for sure. Doesn't hurt to include two extra property checks.

config/visallo.properties
```diff
@@ -70,6 +70,3 @@ security.visibilityTranslator=org.visallo.core.security.DirectVisibilityTranslat
 
 # Uncomment to allow plain HTTP. HTTPS is forced otherwise.
 #http.transportGuarantee=NONE
-
-# Uncomment when deploying to Tomcat
```
joeferner: Put back in change comment to `disable gzip`

joeferner:
I changed how the initial implementation works. Instead of forcing it on or off. I only changed the default behavior and you can always override. On Jetty the default is on. Other servers the default is off.

kunklejr:
I like this better.

PR #318, opened 2016-05-05
Detail: Fix updating compound property visibility or justification

EvanOxfeld:
- [x] @srfarley @joeferner
- [x] @joeybrk372
- [x] @mwizeman @dsingley


web/war/src/main/webapp/js/detail/dropdowns/propertyForm/propForm.js
```diff
@@ -545,7 +546,6 @@ define([
                             metadata: this.currentMetadata
                         }, justification)
                 });
-                self.teardown();
```
srfarley: Is removing `teardown()` part of the fix?

srfarley:
The JS test `webapp/test/unit/spec/detail/itemTest.js` is failing on Travis. Does it pass on your machine?

web/war/src/main/webapp/js/detail/dropdowns/propertyForm/propForm.js
EvanOxfeld: Yes. This component is torn down elsewhere. Tearing down immediately after `addProperty` is triggered caused a silent failure.

EvanOxfeld:
@srfarley The test passes locally. I'll kick off another travis build just in case.

EvanOxfeld:
The travis build failure doesn't seem specific to make change, and I can't reproduce this locally. I'm chalking the failure up to karma/phantom. Fixing `itemTest.js`, if it breaks again, is out of scope of this pull request.

PR #320, opened 2016-05-05
Allow Visallo to run at non-root context paths

kunklejr:
There were only a few fixes that needed to be made to allow Visallo to run at alternate context paths. I did not, however, test every plugin to make sure it doesn't do something that's not compatible with an alternate context path. Fortunately, these changes should have no affect when leaving things at the root (i.e. /) context.

- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley @kunklejr

### DIFF
--- a/web/war/src/main/webapp/index.hbs
+++ b/web/war/src/main/webapp/index.hbs
@@ -15,7 +15,7 @@
 
     <link href='css/visallo.css' rel='stylesheet' media='screen'>
     {{#each pluginCssResources}}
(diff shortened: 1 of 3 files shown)


joeferner:
Weird I'm pretty sure I had visallo running in a non-root context just the other day without these changes.

kunklejr:
I'm not sure how it could have worked before. The app will start but none of the plugins will load, including the authentication plugin.

PR #322, opened 2016-05-06
Refactor search to allow searches to originate from the backend

joeferner:
- [x] @srfarley
- [x] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley

This is a pretty large PR. Most of it consists of moving code from the search web routes to the back-end.

The biggest change is with the introduction of `SearchOptions` which is in place to decouple the code from `HttpRequest`.

@sfeng88 @rygim @jharwig Please look at https://github.com/v5analytics/visallo/compare/refactor-search-to-backend?expand=1#diff-b9711c6c54a8d174d6af7d034a70e6a2R43 I would like to move this code as well to the back-end but I know your team is relying on it so maybe you could help out with that move.

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/search/EdgeSearchRunner.java
+++ b/core/core/src/main/java/org/visallo/core/model/search/EdgeSearchRunner.java
@@ -1,0 +1,34 @@
+package org.visallo.core.model.search;
+
+import com.google.inject.Inject;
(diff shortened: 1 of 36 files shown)


sfeng88:
@joeferner go ahead and move the code you mentioned above to the backend. our team is no longer depending on that.

PR #323, opened 2016-05-09
Handle edge case where array key is already formatted correctly

joeybrk372:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

### DIFF
--- a/web/war/src/main/webapp/js/data/web-worker/util/ajax.js
+++ b/web/war/src/main/webapp/js/data/web-worker/util/ajax.js
@@ -40,7 +40,7 @@ define(['util/promise'], function(Promise) {
 
                 // TODO: support fixing nested arrays
                 if (_.isArray(params[key])) {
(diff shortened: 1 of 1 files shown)


web/war/src/main/webapp/js/data/web-worker/util/ajax.js
```diff
@@ -40,7 +40,7 @@ define(['util/promise'], function(Promise) {
 
                 // TODO: support fixing nested arrays
                 if (_.isArray(params[key])) {
-                    str += _.map(params[key], _.partial(paramPair, key + '[]')).join('&') + '&';
+                    str += _.map(params[key], _.partial(paramPair, !(/\[\]$/).test(key) ? key + '[]' : key)).join('&') + '&';
```
EvanOxfeld: While I'm ok with this change as is, should unfriendly to JavaScript key names such as `sort[]` continue to be used?

PR #329, opened 2016-05-12
Fix select connected (ctrl-e) to use more reliable way to get edges.

jharwig:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

Also introduce a slight delay in the case the user clicks and uses key combo immediately

Adds some info popups that show what is happening

### DIFF
--- a/core/messages/src/main/resources/MessageBundle.properties
+++ b/core/messages/src/main/resources/MessageBundle.properties
@@ -276,6 +276,11 @@ graph.help.scope=Graph
 graph.path.edge.label.one=1 Entity
 graph.path.edge.label.some={0} Entities
 
+graph.select.all.starting=Selecting All Objects...
+graph.select.connected.starting=Selecting Connected...
+graph.select.connected.finished=Selected {0} Entity
+graph.select.connected.finished.plural=Selected {0} Entities
+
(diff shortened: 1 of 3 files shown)


EvanOxfeld:
Looks good. `this.onSelectConnected` is starting to grow -- consider raising the level of abstraction and reducing the promise nesting, which is always easier to read.

Also, what's the source of the delay -- is it due to passing messages between workers and the main thread or is it due to fetching vertices and edges? If it's the latter, is the timeout enough in the case of  a high latency / low bandwidth connection?

jharwig:
@EvanOxfeld The delay is to compensate between the time the graph notices the click and triggers an event that propagates to the document. There is delay built into the graph because of the way we have to coalesce selection changes (cytoscape sends them one at a time – select 10 things, get 10 events) Into one selection change event. It's not based on any network latency, only the built in graph delay (`SELECTION_THROTTLE` in graph.js) and the bubbling delay (which might be synchronous anyways)

Now I could make that a constant to ensure they stay in sync

EvanOxfeld:
A constant sounds good. And thanks for the helpful explanation -- makes more sense now.

jharwig:
@EvanOxfeld done

PR #330, opened 2016-05-12
Diff: Fix incorrect badge count after animation

EvanOxfeld:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

Always resets the badge to the latest diff count after the CSS transition has completed.

### DIFF
--- a/web/war/src/main/webapp/js/workspaces/overlay.js
+++ b/web/war/src/main/webapp/js/workspaces/overlay.js
@@ -266,8 +266,8 @@ define([
                         }
                         count++;
                         return true;
-                    }),
-                    formattedCount = F.number.pretty(count);
+                    });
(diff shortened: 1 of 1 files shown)


web/war/src/main/webapp/js/workspaces/overlay.js
```diff
@@ -212,6 +212,7 @@ define([
             );
         };
 
+        var formattedCount;
```
jharwig: Should we just use `this.formattedCount` instead of a closed var?

EvanOxfeld: Fixed. Was following prior art in this component.

PR #334, opened 2016-05-16
Switch transcript entries + fixes

joeybrk372:
- Clicking on an audio/video link will switch current transcript to that one
- Clicking on audio/video link will scroll the media in view and does not autoplay anymore
- Fix transcript forcing overflowing width of fullscreen detail in Firefox
- Fix transcript entries not being properly removed when switching to a split detail layout.

- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley


web/war/src/main/webapp/js/detail/text/text.js
```diff
@@ -1048,12 +1053,30 @@ define([
             }
         };
 
+        this.scrollToMediaPreview = function($detailBody) {
+            var $preview = $detailBody.find(PREVIEW_SELECTORS.audio).parent().length ?
+                    $detailBody.find(PREVIEW_SELECTORS.audio).parent() :
```
EvanOxfeld: `$.find` can be expensive depending on the DOM structure. To be safe I'd store the result of `$detailBody.find(PREVIEW_SELECTORS.audio).parent()` into a variable rather than executing the same code twice.

web/war/src/main/webapp/less/detail/detail.less
```diff
@@ -18,7 +18,6 @@
   padding: 0;
 
   .org-visallo-layout-root {
-    flex: 1;
```
jharwig: this change worries me that it breaks all kind of layout and scrolling

jharwig: Specifically test resizing of the detail pane, and the multiple selection detail pane with the element list at the bottom. Scrolling should match the mouse and not seem to have weird limits

joeybrk372: This change preserves current layout and scrolling. `org-visallo-layout-root` is filling its parent and is the only flex item at this level. `flex: 1` was not winning to declare any styles in this case.

jharwig: okay, didn't know it had inline styles, thanks

PR #335, opened 2016-05-16
Allow alert.ejs to display html content. Also fix scoping of alert.ejs variables

joeferner:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley

So it turns out that `alert.ejs` was leaking the `info` variable. I wrapped this template in a `function`.

### DIFF
--- a/web/war/src/main/webapp/js/util/alert.ejs
+++ b/web/war/src/main/webapp/js/util/alert.ejs
@@ -1,13 +1,23 @@
+<% (function() { var info, infoHtml; %>
 <% if (typeof error !== "undefined") { %>
   <div class="alert alert-error">
     <% info = error; %>
+<% } else if (typeof errorHtml !== "undefined") { %>
+  <div class="alert alert-error 2222">
+    <% infoHtml = errorHtml; %>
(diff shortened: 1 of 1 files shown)


jharwig:
I'd be worried about displaying user generated content in the alert causing xss vulnerabilities...

jharwig:
and this is starting to have too much logic, it should be a component

joeferner:
@jharwig That's why I made it very explicit. You have to pass `errorHtml`, `infoHtml`, or `messageHtml` instead of `error`, `info`, or `message`

joeferner:
@jharwig also agreed about the component aspect but I didn't really want to change it in the 13 other places it's used.

jharwig:
re: okay okay, I see

You wouldn't need to change all of them initially, leave the ejs file (since the component would use hbs) and add a deprecation console warning inside the template.

It would also allow behavior in the alert panel. Like expandable more info

joeferner:
@jharwig For now can we just live with the changes. I think a component is the right move but it's probably not a Flight component but a React component. Can we make that change after the current release.

PR #340, opened 2016-05-16
Fix fullscreen detail rendering improperly when switching to and from…

joeybrk372:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley

… split layout

### DIFF
--- a/web/war/src/main/webapp/js/util/domElement.js
+++ b/web/war/src/main/webapp/js/util/domElement.js
@@ -7,7 +7,12 @@ define([], function() {
     };
 
     function removeAllStyles(el) {
-        removeStyles.apply(null, [el].concat(_.keys(el.style)));
+        var styles = [];
+        //loop through style object for IE11
+        for (var i in el.style) {
+            styles.push(i);
+        }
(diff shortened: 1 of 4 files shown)


web/war/src/main/webapp/js/util/domElement.js
```diff
@@ -7,7 +7,11 @@ define([], function() {
     };
 
     function removeAllStyles(el) {
-        removeStyles.apply(null, [el].concat(_.keys(el.style)));
+        var styles = [];
+        for (var i in el.style) {
+            styles.push(i);
+        }
```
EvanOxfeld: Converting to an array can be simplified with `_.values(el.style)`

joeybrk372: Here I'm gathering the keys. Switched to the for loop because microsoft's style object doesn't like `Object.keys`

EvanOxfeld: Sounds like a good workaround. Would you please add a comment so we remember? I don't think code will be self-documenting enough in this case.

PR #344, opened 2016-05-17
Graph: Fix hamburger menu position in IE

EvanOxfeld:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley


jharwig:
Would like an addition css definition that makes all the graph options extension be on it's own line.

something like `.tools > div { display: block; }`

EvanOxfeld:
@jharwig Thanks for the feedback. Done.

PR #350, opened 2016-05-18
Fix multiselections in detail history

joeybrk372:
Also allow graph vertices to be deselected with ctrl/command click.

- [x] @srfarley @joeferner
- [ ] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley

### DIFF
--- a/web/war/src/main/webapp/js/data/withObjectSelection.js
+++ b/web/war/src/main/webapp/js/data/withObjectSelection.js
@@ -418,9 +418,9 @@ define([], function() {
                 currentEdgeIds = currentVertexIds.length === 0 && _.keys(currentObjects.edgeIds);
 
             if (!_.isEmpty(currentVertexIds) || !_.isEmpty(currentEdgeIds)) {
-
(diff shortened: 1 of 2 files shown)


web/war/src/main/webapp/js/data/withObjectSelection.js
```diff
@@ -418,9 +418,9 @@ define([], function() {
                 currentEdgeIds = currentVertexIds.length === 0 && _.keys(currentObjects.edgeIds);
 
             if (!_.isEmpty(currentVertexIds) || !_.isEmpty(currentEdgeIds)) {
-
                 var listedIndex = -1,
-                    newSelectionIndex = -1;
+                    newSelectionIndex = -1,
+                    multiple = (currentVertexIds.length || 0 + currentEdgeIds.length || 0) > 1;
```
EvanOxfeld: What's going on here? Is it possible for either `currentVertexIds` or `currentEdgeIds` to not be an array?

PR #351, opened 2016-05-19
fix publishing error for video frames

sfeng88:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley


core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
```diff
@@ -534,7 +534,7 @@ private void publishVertex(Vertex vertex, ClientApiPublishItem.Action action, Au
         // we need to alter the visibility of the json property, otherwise we'll have two json properties, one with the old visibility and one with the new.
         VisalloProperties.VISIBILITY_JSON.alterVisibility(vertexElementMutation, visalloVisibility.getVisibility());
         VisalloProperties.VISIBILITY_JSON.setProperty(vertexElementMutation, visibilityJson, metadata, visalloVisibility.getVisibility());
-        vertexElementMutation.save(authorizations);
+        vertexElementMutation.save(authWithVideoFrame);
```
srfarley: I think it's really strange to make special exceptions in this method for video frames. I understand that `authWithVideoFrame` was already being used, at line 508, and this line fixes the bug appropriately. Are we using a special authorization to hide frame properties from the user?

joeferner: @srfarley Yes we are using a special authorization to hide video frames. By using authorizations we can query and not even pull them across the network from the tablet servers.

PR #361, opened 2016-05-23
Disable some non-map vertex context menu options

joeybrk372:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley


web/war/src/main/webapp/js/util/vertex/menu.js
```diff
@@ -79,7 +82,7 @@ define([
                         event: 'addRelatedItems',
                         shortcut: 'alt+r',
                         shouldDisable: function(selection, vertexId, target) {
-                            return !visalloData.currentWorkspaceEditable;
+                            return !visalloData.currentWorkspaceEditable || $(target).closest('.graph-pane').length === 0;
```
EvanOxfeld: `$.closest` can be expensive because jquery has to walk the DOM subtree. Does this change add noticeable delay, particularly in IE?

joeybrk372: Does not add noticeable delay. I think this paradigm for map/graph context detection is fine for now, as it will most likely get refactored anyways when workspace is supporting multiple graphs and maps.

jharwig:
why add related? That could be used on the map and if the related items have geolocations they will show up

joeybrk372:
Having it on map is misleading, because you will not see all the changes you are causing and could lead to a tedious cleanup of your graph if you are not aware of what is happening. We could alter the behavior to only add related things with geo properties, but since even then you can not view the relationships directly I am in favor of limiting to graph.

PR #375, opened 2016-05-26
Redirect to graph after adding a vertex via URL parameter

EvanOxfeld:
- [ ] @srfarley @joeferner
- [ ] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley


jharwig:
Should probably check to see if graph isn't visible before toggling. There's a configure to make anything the first view after login and this would break that.

EvanOxfeld:
@jharwig Thanks -- I added the graph visible check.

PR #376, opened 2016-05-27
User notification refactor, add getActiveNotificationsOlderThan, add notified field

joeferner:
- [x] @srfarley
- [ ] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley

### DIFF
--- a/core/core-test/src/test/java/org/visallo/core/model/notification/UserNotificationRepositoryTest.java
+++ b/core/core-test/src/test/java/org/visallo/core/model/notification/UserNotificationRepositoryTest.java
@@ -1,0 +1,133 @@
+package org.visallo.core.model.notification;
+
+import com.v5analytics.simpleorm.InMemorySimpleOrmSession;
+import com.v5analytics.simpleorm.SimpleOrmContext;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.visallo.core.model.user.UserRepository;
+import org.visallo.core.model.workQueue.WorkQueueRepository;
+import org.visallo.core.user.User;
+
+import java.util.Date;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Matchers.eq;
+import static org.powermock.api.mockito.PowerMockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class UserNotificationRepositoryTest {
+    private InMemorySimpleOrmSession simpleOrmSession;
+    private UserNotificationRepository userNotificationRepository;
+
+    @Mock
+    private WorkQueueRepository workQueueRepository;
+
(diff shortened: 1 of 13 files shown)


core/core/src/main/java/org/visallo/core/util/PeriodicBackgroundService.java
```diff
@@ -0,0 +1,95 @@
...
+        t.setDaemon(true);
+        t.setName(getThreadName());
+        t.start();
+    }
+
+    private void runPeriodically() {
+        try {
+            while (enabled) {
+                LOGGER.debug("running periodically");
+                Date now = new Date();
```
srfarley: Using the new `java.time` classes might be better.

core/core/src/main/java/org/visallo/core/util/PeriodicBackgroundService.java
```diff
@@ -0,0 +1,95 @@
+package org.visallo.core.util;
+
+import org.apache.commons.lang.time.DateUtils;
+import org.visallo.core.exception.VisalloException;
+import org.visallo.core.model.lock.LeaderListener;
+import org.visallo.core.model.lock.LockRepository;
+
+import java.util.Date;
+
+public abstract class PeriodicBackgroundService {
```
srfarley: Why not use `java.util.concurrent.ScheduledExecutorService`?

core/core-test/src/test/java/org/visallo/core/model/notification/UserNotificationRepositoryTest.java
```diff
@@ -0,0 +1,133 @@
...
+        assertEquals("Current", activeNotifications.get(0).getTitle());
+    }
+
+    @Test
+    public void testGetActiveNotificationsInDateRange() {
+        List<UserNotification> activeNotifications;
+        long currentTime = new Date().getTime();
+
+        simpleOrmSession.save(
+                new UserNotification("user1", "t-120", "Message 1", null, null, new Date(currentTime - 120000), null),
```
srfarley: I would prefer using the new `java.time` classes for the `UserNotification` API.

core/core/src/main/java/org/visallo/core/util/PeriodicBackgroundService.java
joeferner: This does more than just periodic schedules event. It ensures that only one web server is doing this at a given time. This is also just copied from existing code and any further refactors deserves a new PR

core/core/src/main/java/org/visallo/core/util/PeriodicBackgroundService.java
joeferner: again copied from existing code. I think any additional changes should be a new PR

core/core-test/src/test/java/org/visallo/core/model/notification/UserNotificationRepositoryTest.java
joeferner: Are we going to change the API? Again sounds like a new PR

srfarley:
@joeferner I'm OK with a separate PR to upgrade to the new Java time API.

PR #387, opened 2016-06-02
Add AuthorizationMapper to allow applications to override user's authorization and privileges

joeferner:
- [x] @srfarley
- [ ] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

This PR introduces two new classes AuthorizationMapper and AuthorizationContext. AuthorizationMapper maps AuthorizationContext into authorizations and privileges. AuthorizationContext contains what is to be mapped (existing user, user name, ldap info, etc).

### DIFF
--- a/core/core-test/src/main/java/org/visallo/core/model/user/DefaultAuthorizationMapperTest.java
+++ b/core/core-test/src/main/java/org/visallo/core/model/user/DefaultAuthorizationMapperTest.java
@@ -1,0 +1,75 @@
+package org.visallo.core.model.user;
+
+import com.google.common.collect.Sets;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.vertexium.Authorizations;
+import org.vertexium.inmemory.InMemoryAuthorizations;
+import org.visallo.core.user.User;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class DefaultAuthorizationMapperTest {
+    private DefaultAuthorizationMapper defaultAuthorizationMapper;
+
+    @Mock
+    private User user;
+
+    @Mock
+    private UserRepository userRepository;
+
+    @Before
+    public void before() {
+        defaultAuthorizationMapper = new DefaultAuthorizationMapper(userRepository);
+    }
+
+    @Test
+    public void testGetPrivilegesForNewUser() {
+        HashSet<String> expected = Sets.newHashSet("userRepositoryPrivilege1", "userRepositoryPrivilege2");
+        when(userRepository.getDefaultPrivileges()).thenReturn(expected);
+
+        AuthorizationContext authorizationContext = new AuthorizationContext(null);
+        Set<String> privileges = defaultAuthorizationMapper.getPrivileges(authorizationContext);
+        assertEquals(expected, privileges);
+    }
+
+    @Test
+    public void testGetPrivilegesForExisting() {
+        HashSet<String> expected = Sets.newHashSet("userPrivilege1", "userPrivilege2");
+        when(user.getPrivileges()).thenReturn(expected);
+
+        AuthorizationContext authorizationContext = new AuthorizationContext(user);
+        Set<String> privileges = defaultAuthorizationMapper.getPrivileges(authorizationContext);
+        assertEquals(expected, privileges);
+    }
+
+    @Test
+    public void testGetAuthorizationsForNewUser() {
+        HashSet<String> expected = Sets.newHashSet("userRepositoryAuthorization1", "userRepositoryAuthorization2");
+        when(userRepository.getDefaultAuthorizations()).thenReturn(expected);
+
+        AuthorizationContext authorizationContext = new AuthorizationContext(null);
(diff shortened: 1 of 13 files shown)


core/core-test/src/main/java/org/visallo/core/model/user/DefaultAuthorizationMapperTest.java
```diff
@@ -0,0 +1,75 @@
...
+    public void before() {
+        defaultAuthorizationMapper = new DefaultAuthorizationMapper(userRepository);
+    }
+
+    @Test
+    public void testGetPrivilegesForNewUser() {
+        HashSet<String> expected = Sets.newHashSet("userRepositoryPrivilege1", "userRepositoryPrivilege2");
+        when(userRepository.getDefaultPrivileges()).thenReturn(expected);
+
+        AuthorizationContext authorizationContext = new AuthorizationContext(null);
```
srfarley: Use a no-arg constructor instead. Even better, use a more descriptive factory method like `AuthorizationContext.newUserContext()` which can return a singleton.

core/core/src/main/java/org/visallo/core/model/user/DefaultAuthorizationMapper.java
```diff
@@ -0,0 +1,37 @@
...
+        }
+        return authorizationContext.getExistingUser().getPrivileges();
+    }
+
+    @Override
+    public Set<String> getAuthorizations(AuthorizationContext authorizationContext) {
+        if (authorizationContext.isNewUser()) {
+            return this.userRepository.getDefaultAuthorizations();
+        }
+        return Sets.newHashSet(
```
srfarley: Can this return an immutable set?

core/core/src/main/java/org/visallo/core/model/user/UserRepository.java
```diff
@@ -347,9 +364,28 @@ public final void delete(User user) {
     protected abstract void internalDelete(User user);
 
     public final void setPrivileges(User user, Set<String> privileges, User authUser) {
```
srfarley: While you're in here, do we still want this method to be `final`?

core/core/src/main/java/org/visallo/core/model/user/UserRepository.java
```diff
@@ -347,9 +364,28 @@ public final void delete(User user) {
...
-        fireUserPrivilegesUpdatedEvent(user, privileges);
+        if (!privileges.equals(user.getPrivileges())) {
+            internalSetPrivileges(user, privileges, authUser);
+            sendNotificationToUserAboutPrivilegeChange(user, privileges, authUser);
+            fireUserPrivilegesUpdatedEvent(user, privileges);
+        }
+    }
+
+    public void setAuthorizations(User user, Set<String> authorizations, User authUser) {
+        Set<String> currentAuthorizations = Sets.newHashSet(getAuthorizations(user).getAuthorizations());
```
srfarley: This set can be immutable.

core/core-test/src/main/java/org/visallo/core/model/user/DefaultAuthorizationMapperTest.java
joeferner: I would rather not pollute the public API with a method that should never be called. This is here only for testing.

srfarley:
Travis is failing due to failed JS tests. Does this branch need rebasing?

core/core/src/main/java/org/visallo/core/model/user/DefaultAuthorizationMapper.java
joeferner: done, as a new commit since it also touches User to do it correctly.

core/core/src/main/java/org/visallo/core/model/user/UserRepository.java
joeferner: Yes subclasses should extend `internalSetPrivileges` not this method.

core/core/src/main/java/org/visallo/core/model/user/UserRepository.java
joeferner: done

core/core/src/main/java/org/visallo/core/user/User.java
```diff
@@ -41,7 +42,7 @@
 
     JSONObject getUiPreferences();
 
-    Set<String> getPrivileges();
+    ImmutableSet<String> getPrivileges();
```
srfarley: I really don't like public interfaces depending on Guava classes. There's little advantage to it since the interface is still a `java.util.Set` with mutation methods. I always hide the immutable implementation. If this were C#, on the other hand...

joeferner: I like exposing it because it self documents the API and the expectations.

srfarley: But we also don't do this anywhere else. I'd rather lean toward consistency right now, and overhaul the public API separately.

joeferner: I think it's even more important in cases like this to force behavior of subclasses. For example if `InMemoryUser` didn't return an immutable set and you wrote code against that expectation in other parts of the code base then later switched to `VertexiumUser` which did return an immutable set you would be broken.

joeferner: In that case, if we are going for consistency then I think we should return a mutable Set until we decide to refactor the API and return a `ImmutableSet`

joeferner: I reverted the `ImmutableSet` changes and we can create a new PR to address public API mutable/immutable issues

PR #388, opened 2016-06-03
Support comments on edges

joeferner:
- [x] @srfarley
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

Introduces `topObjectProperty` which is the standard owl top level that all relationships inherit from. This allows us to attach the comment property to topObjectProperty and therefore the ACLProvider can verify.

### DIFF
--- a/core/core/src/main/java/org/visallo/core/security/ACLProvider.java
+++ b/core/core/src/main/java/org/visallo/core/security/ACLProvider.java
@@ -85,8 +85,12 @@ public abstract class ACLProvider {
                 iri = concept.getParentConceptIRI();
             }
         } else if (element instanceof Edge) {
-            Relationship relationship = ontologyRepository.getRelationshipByIRI(((Edge) element).getLabel());
-            populatePropertyAcls(relationship, element, user, propertyAcls);
+            String iri = ((Edge) element).getLabel();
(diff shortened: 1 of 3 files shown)


core/core/src/main/java/org/visallo/core/security/ACLProvider.java
```diff
@@ -172,9 +178,14 @@ private void populatePropertyAcls(HasOntologyProperties hasOntologyProperties, E
 
         // for properties that don't exist on the element, use the ontology property definition and omit the key.
         propertyAcls.addAll(ontologyProperties.stream()
-                .filter(ontologyProperty -> !addedPropertyNames.contains(ontologyProperty.getTitle()))
-                .map(ontologyProperty -> newClientApiPropertyAcl(element, null, ontologyProperty.getTitle(), user))
-                .collect(Collectors.toList()));
+                                    .filter(ontologyProperty -> !addedPropertyNames.contains(ontologyProperty.getTitle()))
```
srfarley: Nitpick: Excessive indentation.

joeferner: fixed

PR #389, opened 2016-06-03
Allow fullscreen detail to add/remove items

joeybrk372:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

### DIFF
--- a/web/war/src/main/webapp/js/appFullscreenDetails.js
+++ b/web/war/src/main/webapp/js/appFullscreenDetails.js
@@ -39,13 +39,15 @@ define([
                 e.stopPropagation();
                 e.preventDefault();
 
-                var vertexIds = d.vertexIds || (d.objects ? _.pluck(d.objects, 'id') : null);
-                if (vertexIds) {
-                    vertexIds = _.isArray(vertexIds) ? vertexIds : [vertexIds];
-                    self.updateVertices({
-                        add: vertexIds
-                    })
-                }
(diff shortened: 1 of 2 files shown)


web/war/src/main/webapp/js/appFullscreenDetails.js
```diff
@@ -39,13 +39,15 @@ define([
...
 
-                var vertexIds = d.vertexIds || (d.objects ? _.pluck(d.objects, 'id') : null);
-                if (vertexIds) {
-                    vertexIds = _.isArray(vertexIds) ? vertexIds : [vertexIds];
-                    self.updateVertices({
-                        add: vertexIds
-                    })
-                }
+                var vertexIds = _.compact(_.isArray(d.vertexIds) ? d.vertexIds : [d.vertexIds]);
+                var edgeIds = _.compact(_.isArray(d.edgeIds) ? d.edgeIds : [d.edgeIds]);
```
EvanOxfeld: Is it valid for `d.vertexIds` and `d.edgeIds` be sparse arrays? Not only does it seem semantically invalid for a `selectObjects` event, but also this check would be required in every component that listens to `selectObjects`.

joeybrk372: `selectObjects` will not come with sparse arrays, I am compacting to prevent a sparse array from my transformation of the event data here to be useable by the next functions.

PR #393, opened 2016-06-06
Fix propForm not tearing down when resolving a property from text

joeybrk372:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [ ] @mwizeman @dsingley

### DIFF
--- a/web/war/src/main/webapp/js/detail/dropdowns/propertyForm/propForm.js
+++ b/web/war/src/main/webapp/js/detail/dropdowns/propertyForm/propForm.js
@@ -563,7 +563,8 @@ define([
             _.defer(this.buttonLoading.bind(this, this.attr.deleteButtonSelector));
             this.trigger('deleteProperty', {
                 vertexId: this.attr.data.id,
-                property: _.pick(this.currentProperty, 'key', 'name')
+                property: _.pick(this.currentProperty, 'key', 'name'),
+                node: this.node
             });
         };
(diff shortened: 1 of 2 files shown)


web/war/src/main/webapp/js/detail/dropdowns/propertyForm/propForm.js
```diff
@@ -607,7 +608,8 @@ define([
                             visibilitySource: this.visibilitySource.value,
                             oldVisibilitySource: oldVisibilitySource,
                             metadata: this.currentMetadata
-                        }, justification)
+                        }, justification),
+                    node: this.node
```
jharwig: this isn't necessary `this.trigger` will fire the event on this.node, and will bubble up the DOM. Use `event.target` to get `this.node` in the listener of this event.

joeybrk372: It is necessary to pass `this.node` here because the detail pane is retriggering the event on the properties component itself.

jharwig: Okay, It's bad practice to pass non-serializable content in event data, like functions or Elements. I think it's okay to let this through for now though. To truly fix this would be more work and these components (properties, comments, dropdown) are prime cases for react.

PR #395, opened 2016-06-07
Persist saved search card titles

joeybrk372:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

Don't override the card's title when switching to a different search if it is a custom title.


EvanOxfeld:
Change as-is look good. That said, would it be less confusing if the saved search name was the title's placeholder value rather than the actual title?

joeybrk372:
No, because I think the user will be more likely to want to edit from their own custom title.

jharwig:
Can someone explain what behavior we are implementing here?

joeybrk372:
Upon changing which search is displayed in the card:
- The card title will change to match the search name unless the user has edited the card's title.
- The title input placeholder will always match the card's current title.

jharwig:
Okay, I could argue that it should always change when the search does. But I can see preferring a users edit too. Good as is.

PR #398, opened 2016-06-09
Fix edge detail having add property option with no edge properties

joeybrk372:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley


srfarley:
Was this issue also occurring with vertices?

joeybrk372:
Yeah, it would have also happened for vertices.

PR #404, opened 2016-06-13
Ontology code generation

mwizeman:
- [x] @srfarley @joeferner
- [ ] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley


tools/ontology-ingest/codegen/src/main/java/org/visallo/tools/ontology/ingest/codegen/CodegenCLI.java
```diff
@@ -0,0 +1,62 @@
...
+import org.visallo.web.clientapi.UserNameAndPasswordVisalloApi;
+import org.visallo.web.clientapi.VisalloApi;
+import org.visallo.web.clientapi.model.ClientApiOntology;
+
+import java.io.File;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+
+@Parameters(commandDescription = "Generate model classes based on a Visallo ontology.")
+public class CodegenCLI extends CommandLineTool {
```
joeferner: Can you rename this to `ModelCodeGen`. When using the cli tool you can run the tools with `cli codegencli` which would be nicer if it were `cli modelcodegen`

tools/ontology-ingest/codegen/src/main/java/org/visallo/tools/ontology/ingest/codegen/CodegenCLI.java
```diff
@@ -0,0 +1,62 @@
...
+            ontologyJsonString = new String(Files.readAllBytes(inputJsonFile.toPath()), Charset.forName("UTF-8"));
+        } else if (!Strings.isNullOrEmpty(visalloUrl) && !Strings.isNullOrEmpty(visalloUsername) && !Strings.isNullOrEmpty(visalloPassword)) {
+            VisalloApi visalloApi = new UserNameAndPasswordVisalloApi(visalloUrl, visalloUsername, visalloPassword, true);
+            ontologyJsonString = visalloApi.invokeAPI("/ontology", "GET", null, null, null, null, "application/json");
+            visalloApi.logout();
+        } else {
+            throw new VisalloException("inputJsonFile or visalloUrl, visalloUsername, and visalloPassword parameters are required");
+        }
+
+        ClientApiOntology ontology = JsonUtil.getJsonMapper().readValue(ontologyJsonString, ClientApiOntology.class);
```
joeferner: Is there a reason we are using the client api here instead of `OntologyRepository`?

tools/ontology-ingest/codegen/src/main/java/org/visallo/tools/ontology/ingest/codegen/ConceptWriter.java
```diff
@@ -0,0 +1,54 @@
...
+        super(outputDirectory, ontology);
+    }
+
+    protected void writeClass(ClientApiOntology.Concept concept) {
+        String conceptPackage = packageNameFromIri(concept.getId());
+        if (conceptPackage != null) {
+            String conceptClassName = classNameFromIri(concept.getId());
+
+            // Don't expose the visallo internal concepts to the generated code
+            if (conceptPackage.startsWith("org.visallo") && !conceptClassName.equals("Root")) {
```
joeferner: Can we have a toggle for this? I'm thinking it might be nice to code gen many of the concepts and properties from things like `comment.owl`, `longRunningProcess.owl`, etc

tools/ontology-ingest/common/src/main/java/org/visallo/tools/ontology/ingest/common/IngestRepository.java
```diff
@@ -0,0 +1,342 @@
...
+                        propertyAddition.getValue(),
+                        buildMetadata(propertyAddition.getMetadata(), propertyAddition.getVisibility()),
+                        getTimestamp(propertyAddition.getTimestamp()),
+                        getVisibility(propertyAddition.getVisibility())
+                );
+            }
+        }
+    }
+
+    public ValidationResult validateConceptBuilder(ConceptBuilder builder) {
```
joeferner: Might be nice if we returned all `ValidationResult`s

tools/ontology-ingest/common/src/main/java/org/visallo/tools/ontology/ingest/common/IngestRepository.java
```diff
@@ -0,0 +1,342 @@
...
+        }
+
+        return metadata;
+    }
+
+    private Authorizations getAuthorizations() {
+        return userRepository.getAuthorizations(getIngestUser());
+    }
+
+    private User getIngestUser() {
```
joeferner: Typically in most if not all of the Visallo APIs we pass in a user should we do the same instead of using this method?

tools/ontology-ingest/common/src/main/java/org/visallo/tools/ontology/ingest/common/PropertyAddition.java
```diff
@@ -0,0 +1,79 @@
...
+public class PropertyAddition<T> {
+    private String iri;
+    private T value;
+    private String key;
+    private Map<String, Object> metadata;
+    private Long timestamp;
+    private String visibility;
+
+    public PropertyAddition(String iri, String key, T value) {
+        assert !Strings.isNullOrEmpty(iri);
```
joeferner: Can you change this to google's guava check* API to be consistant

tools/pom.xml
```diff
@@ -22,5 +22,6 @@
         <module>rdf-import</module>
         <module>reindex</module>
         <module>owl</module>
+        <module>ontology-ingest</module>
```
joeferner: @dsingley says make this alphabetical ordering

tools/ontology-ingest/codegen/src/main/java/org/visallo/tools/ontology/ingest/codegen/CodegenCLI.java
```diff
@@ -0,0 +1,62 @@
...
+        CommandLineTool.main(new CodegenCLI(), args, false);
+    }
+
+    @Override
+    protected int run() throws Exception {
+        String ontologyJsonString;
+        if (inputJsonFile != null) {
+            ontologyJsonString = new String(Files.readAllBytes(inputJsonFile.toPath()), Charset.forName("UTF-8"));
+        } else if (!Strings.isNullOrEmpty(visalloUrl) && !Strings.isNullOrEmpty(visalloUsername) && !Strings.isNullOrEmpty(visalloPassword)) {
+            VisalloApi visalloApi = new UserNameAndPasswordVisalloApi(visalloUrl, visalloUsername, visalloPassword, true);
```
joeferner: Should we abstract this logic out somehow to ClientAPI. How would I use this tool if I'm not UserName and password?

tools/ontology-ingest/common/src/main/java/org/visallo/tools/ontology/ingest/common/IngestRepository.java
```diff
@@ -0,0 +1,342 @@
...
+import org.visallo.core.user.User;
+import org.visallo.core.util.VisalloLogger;
+import org.visallo.core.util.VisalloLoggerFactory;
+import org.visallo.web.clientapi.model.PropertyType;
+import org.visallo.web.clientapi.model.VisibilityJson;
+
+import java.math.BigDecimal;
+import java.util.*;
+
+public class IngestRepository {
```
joeferner: Typically all of our repositories are thread safe and singletons should we instead pass around Ingest options or something instead of storing state?

tools/ontology-ingest/codegen/src/main/java/org/visallo/tools/ontology/ingest/codegen/CodegenCLI.java
mwizeman: Fixed with 8deb7ebd7591cdd799738715edadc94f3e965718

tools/ontology-ingest/codegen/src/main/java/org/visallo/tools/ontology/ingest/codegen/CodegenCLI.java
mwizeman: This was done to allow the flexibility for a development team to generate ingest code for a production environment when they don't have backend access.

tools/ontology-ingest/codegen/src/main/java/org/visallo/tools/ontology/ingest/codegen/ConceptWriter.java
mwizeman: Fixed with 8deb7ebd7591cdd799738715edadc94f3e965718. I added a new CLI flag `--includeVisalloClasses` that will cause the tool to generate classes for all concepts and relationships.

tools/ontology-ingest/common/src/main/java/org/visallo/tools/ontology/ingest/common/IngestRepository.java
mwizeman: I added a set method to allow the caller to specify the `ingestUser`. This fix addresses this particular comment, but will need to be changed if we update the code to be thread safe per your other comment.

tools/ontology-ingest/common/src/main/java/org/visallo/tools/ontology/ingest/common/IngestRepository.java
```diff
@@ -0,0 +1,342 @@
...
+        public static final ValidationResult VALID_RESULT = new ValidationResult(true, null);
+
+        private boolean valid = false;
+        private String validationError;
+
+        public ValidationResult(String validationError) {
+            this.validationError = validationError;
+        }
+
+        public ValidationResult(boolean valid, String validationError) {
```
mwizeman: It isn't. I left it in there as public, but it is only ever used to create the VALID_RESULT.

tools/ontology-ingest/common/src/main/java/org/visallo/tools/ontology/ingest/common/PropertyAddition.java
mwizeman: Fixed with 8deb7ebd7591cdd799738715edadc94f3e965718

tools/pom.xml
mwizeman: Fixed with 8deb7ebd7591cdd799738715edadc94f3e965718

tools/ontology-ingest/common/src/main/java/org/visallo/tools/ontology/ingest/common/IngestRepository.java
mwizeman: Moved to IngestOptions with 46ed03ae58039c5127d7ae240c409b250e74c441

tools/ontology-ingest/common/src/main/java/org/visallo/tools/ontology/ingest/common/IngestRepository.java
mwizeman: Fixed with 46ed03ae58039c5127d7ae240c409b250e74c441 by adding an IngestOptions class

tools/ontology-ingest/codegen/src/main/java/org/visallo/tools/ontology/ingest/codegen/CodegenCLI.java
mwizeman: The username/password support is there as a convenience. The other way to provide the ontology to the tool is by passing a file reference. The idea is that the user will visit `/ontology` of their Visallo instance in a browser and save the output of that route to a local file.

tools/ontology-ingest/README.md
```diff
@@ -0,0 +1,115 @@
...
+        <dependency>
+            <groupId>org.visallo</groupId>
+            <artifactId>visallo-tools-ontology-ingest-codegen</artifactId>
+            <version>${visallo.version}</version>
+        </dependency>
+```
+
+##### generated
+
+The `generated` module will only include the generated source files and will be a dependancey of your `ingest` module. Specify the follwing Maven dependency:
```
srfarley: Change "dependancey" to "dependency".

tools/ontology-ingest/README.md
```diff
@@ -0,0 +1,115 @@
...
+        --visalloPassword password
+        --outputDirectory /path/to/project/generated/src/main/java
+```
+
+If using a saved `.json` file provide the following command line arguments:
+
+```bash
+        --inputJsonFile /path/to/saved/ontology.json
+        --outputDirectory /path/to/project/generated/src/main/java
+```
```
srfarley: Should we encourage more structure around saving the ontology file to a project subdirectory, and committing it to source control? We could go even further and have `maven compile` automatically regenerate and compile the code, assuming the current ontology file is always in that known location.

Should we define these modules in our Maven Archetype? It already provides an ontology file in a known location, and expects it to be committed along with the rest of the project. Then the source code can be generated before the plugin modules are compiled. The plugin modules would have dependencies on the generated module.

dsingley: check in .json - i don't think so, it's only needed once
add modules to archetype - sure

srfarley: I didn't realize the ontology is a JSON file. Why aren't you using the standard OWL file format that you're referring to in the documentation link (http://docs.visallo.org/getting-started/ontology.html)? Where is the JSON format documented?

dsingley: The .json file from /ontology (or the same accessed via the API) is the
union of all the .owl files loaded in a Visallo instance. It's not
explicitly documented but it is the existing data structure consumed by the
Javascript front end.

PR #409, opened 2016-06-15
Allow sort to be passed as string array or json array

jharwig:
- [x] @srfarley @joeferner
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

Convenience for the front-end from having to JSON.encode sort parameters. Allow both legacy `sort[]` with string array, and `sort` that accepts `JsonArray`. Fixes some issues with plugins that call search services.


srfarley:
Why do we need to support both formats?

jharwig:
Allows plugins to send search sort parameters to the server without doing the exact transformation that the search service does.

PR #416, opened 2016-06-20
Add WorkspaceRepository findall method (and reformat code)

joeferner:
- [ ] @srfarley
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

Adds to https://github.com/v5analytics/visallo/pull/413 review that PR first

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/user/DefaultUserListener.java
+++ b/core/core/src/main/java/org/visallo/core/model/user/DefaultUserListener.java
@@ -1,6 +1,7 @@
 package org.visallo.core.model.user;
 
 import org.visallo.core.user.User;
+import org.visallo.web.clientapi.model.UserStatus;
 
 import java.util.Set;
 
@@ -27,6 +28,16 @@ public class DefaultUserListener implements UserListener {
 
     @Override
     public void userAddAuthorization(User user, String auth) {
+
+    }
+
+    @Override
+    public void userLogin(User user, String remoteAddr) {
+
+    }
+
+    @Override
+    public void userStatusChange(User user, UserStatus status) {
 
     }
 }
--- a/core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
+++ b/core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
@@ -117,6 +117,13 @@ public abstract class WorkspaceRepository {
      */
     public abstract Iterable<Workspace> findAllForUser(User user);
 
+    /**
+     * Finds all workspaces irregardless of access.
+     *
+     * @param user a user with access to all workspaces such as system user.
+     */
+    public abstract Iterable<Workspace> findAll(User user);
+
     public abstract void setTitle(Workspace workspace, String title, User user);
 
     public abstract List<WorkspaceUser> findUsersWithAccess(String workspaceId, User user);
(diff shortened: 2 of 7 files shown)


core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
```diff
@@ -376,20 +444,29 @@ private void publishVertices(ClientApiPublishItem[] publishData, ClientApiPublis
...
+                        FetchHint.ALL_INCLUDING_HIDDEN,
+                        authorizations
+                );
                 Vertex outVertex = edge.getVertex(Direction.OUT, authorizations);
                 Vertex inVertex = edge.getVertex(Direction.IN, authorizations);
-                if (SandboxStatusUtil.getSandboxStatus(edge, workspaceId) == SandboxStatus.PUBLIC && !WorkspaceDiffHelper.isPublicDelete(edge, authorizations)) {
+                if (SandboxStatusUtil.getSandboxStatus(
+                        edge,
+                        workspaceId
+                ) == SandboxStatus.PUBLIC && !WorkspaceDiffHelper.isPublicDelete(edge, authorizations)) {
```
srfarley: This didn't reformat very nicely. It's strange to see `) ==` at the beginning of the line.

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
```diff
@@ -224,11 +272,37 @@ public void addWorkspaceToUser(Vertex workspaceVertex, Vertex userVertex, Author
     }
 
     @Override
+    public Iterable<Workspace> findAll(User user) {
+        if (!user.equals(getUserRepository().getSystemUser())) {
```
srfarley: What's the point of passing `user` if you always have to pass the system user? Same with getting the authorizations - just use the system "visallo" authorization.

core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
joeferner: missed that one... fixed

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
joeferner: Couple of reasons...
* Following convention. All of our methods should be accepting a `User` or `Authorizations`.
* `VertexiumWorkspaceRepository` is only one implementation. This call could be going to an external system which might require user authorizations. So to keep the public API consistent I pass in user. 
* Currently I only allow system user because this circumvents security and I want the caller to be explicit in what they are doing.

core/core/src/main/java/org/visallo/core/model/user/DefaultUserListener.java
```diff
@@ -29,4 +30,14 @@ public void userRemoveAuthorization(User user, String auth) {
     public void userAddAuthorization(User user, String auth) {
 
     }
+
+    @Override
+    public void userLogin(User user, String remoteAddr) {
+
```
EvanOxfeld: Is it typical to have default no-op function implementations like this? At minimum should we log something?

srfarley: There's no need to log. This is just a very uninterested listener.

core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
```diff
@@ -117,6 +117,13 @@ public Workspace add(String title, User user) {
      */
     public abstract Iterable<Workspace> findAllForUser(User user);
 
+    /**
+     * Finds all workspaces irregardless of access.
```
EvanOxfeld: Not a word

srfarley: I was going to mention, but technically it is a word, it's just frowned upon.

joeferner: Google thinks it is a word. It even has it's own wikipedia page https://www.wikiwand.com/en/Irregardless :)

PR #417, opened 2016-06-21
upgrade guice to 4.1.0

joeferner:
- [x] @srfarley
- [x] @EvanOxfeld @joeybrk372
- [x] @mwizeman @dsingley

Upgrading guava appears to fix the issue of getting `ArrayIndexOutOfBoundsException`s if injecting fails.

I excluded guava because guice wants v19 for some reason even though it doesn't appear to be used in their code anywhere https://github.com/google/guice/search?utf8=%E2%9C%93&q=guava

### DIFF
--- a/root/pom.xml
+++ b/root/pom.xml
@@ -62,7 +62,7 @@
         <curator.version>2.8.0</curator.version>
         <elasticsearch.version>1.4.4</elasticsearch.version>
         <evo.inflector.version>1.0.1</evo.inflector.version>
-        <guice.version>3.0</guice.version>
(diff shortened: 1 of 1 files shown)


srfarley:
It looks like they only use Guava indirectly for testing OSGI: https://github.com/google/guice/blob/01e417a6e11047ac1d86c9907184c2ad8a1f198f/core/test/com/googlecode/guice/OSGiContainerTest.java

joeferner:
@srfarley  That's what I saw too but it's not marked as a test dependency which I think is odd.

PR #418, opened 2016-06-21
Fix concept type visibility when element visibility is changed by the client

srfarley:
- [x] @sfeng88 @rygim @jharwig @joeybrk372
- [x] @joeferner
- [x] @EvanOxfeld @mwizeman @dsingley

Without this fix, users that can view an element due to a visibility change see the element as a "Thing" and not as its actual concept type.


core/core/src/main/java/org/visallo/core/model/graph/GraphRepository.java
```diff
@@ -84,16 +84,26 @@ public void writeVersion(int version) {
...
-        ExistingElementMutation<T> m = element.<T>prepareMutation().alterElementVisibility(visalloVisibility.getVisibility());
-        if (VisalloProperties.VISIBILITY_JSON.getPropertyValue(element) != null) {
+        if (VisalloProperties.VISIBILITY_JSON.hasProperty(element)) {
             Property visibilityJsonProperty = VisalloProperties.VISIBILITY_JSON.getProperty(element);
-            m.alterPropertyVisibility(visibilityJsonProperty.getKey(), VisalloProperties.VISIBILITY_JSON.getPropertyName(), defaultVisibility);
+            m.alterPropertyVisibility(
+                    visibilityJsonProperty.getKey(), VisalloProperties.VISIBILITY_JSON.getPropertyName(),
+                    defaultVisibility);
+        }
+        VisalloProperties.VISIBILITY_JSON.setProperty(m, visibilityJson, defaultVisibility);
```
srfarley: Note that I'm now storing the VISIBILITY_JSON property with empty metadata. It's unclear to me whether or not metadata is necessary for this property.

PR #422, opened 2016-06-24
Detail previous selection history: support removal of items from history

joeybrk372:
- [x] @srfarley @joeferner


joeferner:
I know it's existing code but that is one heck of a long function any chance we can break it down a bit?

PR #427, opened 2016-06-27
add a react Alert component

joeferner:
- [x] @srfarley

I pulled this file out of PR https://github.com/v5analytics/visallo/pull/405 because I need it sooner for another project

This is a partial copy of /web/war/src/main/webapp/js/util/alert.ejs but changed to react.


srfarley:
Would it be difficult to remove the old alert component and replace its usages with this one?

joeferner:
No, it's all over the place and it's usage is many times very flight specific unfortunately.

PR #434, opened 2016-06-29
Handle bad comment property keys in javascript

joeferner:
- [x] @srfarley

If a comment is created via RDF or entered by an external process the key may not always be constructed correctly (Should be timestamp). This fix allows the UI to still function under these circumstances. Without it the details pane becomes completely unresponsive in Firefox and IE.

### DIFF
--- a/web/war/src/main/webapp/js/comments/comments.js
+++ b/web/war/src/main/webapp/js/comments/comments.js
@@ -88,7 +88,11 @@ define([
         var isLegacyKeyInServerTimezone = !(/Z$/).test(p.key),
             date = null;
         if (isLegacyKeyInServerTimezone) {
-            date = F.date.utc(new Date(p.key + 'Z').getTime());
+            var time = new Date(p.key + 'Z').getTime();
+            if (isNaN(time)) {
+                return undefined;
+            }
+            date = F.date.utc(time);
         } else {
             date = F.date.local(p.key);
         }
@@ -96,6 +100,7 @@ define([
         if (millis && !isNaN(millis)) {
             return millis;
         }
+        return undefined;
     }
 
     function isEdited(created, modified) {


joeferner:
Should we log an error here so this doesn't go unnoticed? One concern is that bad data could continue to be ingested incorrectly and never be fixed. This PR will hide that fact.

srfarley:
Where will it log? This would be discovered on the client side only. RDF import doesn't validate anything.

EvanOxfeld:
Who's responsible for bad comment data if it's likely to happen? Would it make more sense to handle bad data in whatever service fetches comment data or in the backend endpoint? I tend to think the comment component should only be concerned with presentation.

joeferner:
@srfarley fixed a bug with Edge comments (https://github.com/v5analytics/visallo/pull/443) which were putting in incorrect data.  But really this bad data could come from all sorts of places (RDF Import, Spark, custom ingest logic, etc.) we would never be able to catch them all.

srfarley:
@joeferner is correct, there's currently no way to validate and sanitize all data ingested from the back-end. The UI needs to be resilient. I don't like it either - I would prefer that the UI stick to presentation, but it's our only choice right now.

EvanOxfeld:
@srfarley Got it. What I'm asking is which is the more appropriate place for sanitizing bad vertex metadata, the workspace web worker service responsible for fetching vertices/comments or the component responsible for presenting comments?

joeferner:
Unfortunately I don't think the router can sanitize the key because the UI uses it for additional actions such as editing, replying, viewing metadata, etc. So we'll need to keep it "incorrect" at least up until that point.

srfarley:
I don't think the web worker services are the right place. Maybe @jharwig has a preference.

EvanOxfeld:
This LGTM assuming my comments didn't sway @jharwig.

PR #435, opened 2016-06-29
Remove sessionTimeout command line option from WebServer.java. This option is not used by the Jetty instance

joeferner:
- [x] @srfarley

The option isn't used and is moved closer to a web server which does use it.


dsingley:
can we instead improve the Jetty implementation to respect the parameter so we can test in both containers?

joeferner:
Closed see https://github.com/v5analytics/visallo/pull/444

PR #438, opened 2016-06-29
Precompile all .jsx -> .js

jharwig:
- [x] @srfarley @joeferner
- [x] @joeybrk372 @sfeng88 @rygim

Remove the jsx requirejs plugin as the performance was almost 1/2 second for transformation.

This comes in 2 parts:
1. For Visallo JS: grunt babel plugin precompiles all jsx to js before requirejs copies over at compile time.
2. For Plugin JS: New `registerJavaScriptComponent` Webapp method that compiles jsx using embedded babel at runtime.

Source maps are always created. But in `pluginDevMode=true` the maps are inlined in the files. When `false` source maps are external and linked with a header. React dev tools shows the component displayNames properly.

# Incompatibility

```js
require('jsx!...')
```
is no longer available as it has major performance issues, especially as the component number increases.

## To Fix

1. If it's a plugin file, register with the new `registerJavaScriptComponent` method.
2. Change any `require` or `define` to remove the `jsx!` bit.


jharwig:
This needs testing on other platforms. Just `make clean all` in webapp

web/war/src/main/webapp/js/components/Alert.jsx
```diff
@@ -1,6 +1,7 @@
 define([
     'react'
-], function (React) {
+], function(React) {
+    'use strict';
```
EvanOxfeld: Babel should automatically add `use strict` when transpiling. That said, it's not like this is harmful and isn't a big deal if it makes running eslint easier.

jharwig: yeah, it was lint errors after i enabled eslint to look at jsx

jharwig:
Working on the npm failure, seems to be a bug in our npm version. https://github.com/npm/npm/issues/10727

Upgrading both node and npm fixes for me. Will see about travis...

web/web-base/src/main/java/org/visallo/web/WebApp.java
```diff
@@ -191,6 +192,32 @@ public void registerJavaScript(String scriptResourceName) {
...
+     * Register a JSX react component.
+     *
+     * Converts .jsx files to .js files using babel.
+     *
+     * Source maps are always created, but placed inline in
+     * pluginDevMode and externally linked when not.
+     *
+     * @param scriptResourceName
+     */
+    public void registerJavaScriptComponent(String scriptResourceName) {
```
srfarley: Should this method name be more explicit about what type of component it accepts, i.e. `registerJavaScriptJsxComponent()`, or is it intentionally generic so we can handle other types in the future?

jharwig: Yeah, generic was the plan – Other methods are pretty generic too (registerJavaScriptTemplate).

PR #440, opened 2016-06-30
ACL improvements and new privileges for comments

srfarley:
- [x] @joeferner

This PR includes the following changes:
* The ACL properties (addable, updatable, deletable) are now derived by considering the user privileges in addition to the `ACLProvider` implementation on the back-end. Client-side code now uses only those properties to configure the UI without having to replicate the same logic as the back-end.
* Implementing an `ACLProvider` now requires fewer abstract methods to be implemented.
* The `/vertex/acl` and `/edge/acl` routes have been removed. The ACL object is now attached to the vertex and edge JSON objects, and are cached along with those objects in the UI.
* Two new privileges have been added: `COMMENT_EDIT_ANY` and `COMMENT_DELETE_ANY`. These allow a user to edit or delete any other users' comments. Normally, with the `COMMENT` privilege, a user can only edit/delete his own comments.


core/core/src/main/java/org/visallo/core/security/ACLProvider.java
```diff
@@ -1,24 +1,39 @@
...
+import java.util.*;
 import java.util.stream.Collectors;
 
 public abstract class ACLProvider {
+    private Graph graph;
+    private UserRepository userRepository;
+    private OntologyRepository ontologyRepository;
+
+    @Inject
+    public final void setGraph(Graph graph) {
```
joeferner: Why did you choose setter methods over constructor?

web/client-api/src/main/java/org/visallo/web/clientapi/model/Privilege.java
```diff
@@ -7,7 +7,9 @@
 
 public final class Privilege {
     public static final String READ = "READ";
-    public static final String COMMENT = "COMMENT";
+    public static final String COMMENT = "COMMENT"; // add comments and edit/delete own comments
+    public static final String COMMENT_EDIT_ANY = "COMMENT_EDIT_ANY"; // edit other users' comments
+    public static final String COMMENT_DELETE_ANY = "COMMENT_DELETE_ANY"; // delete other users' comments
```
joeferner: Do these changes belong in this PR?

core/core/src/main/java/org/visallo/core/security/ACLProvider.java
srfarley: I did it that way to make it easier for the subclasses, but I don't mind using the constructor.

srfarley: I switched to constructor injection.

joeferner: I've been thinking about this for a lot of our classes. If we expect someone to extend our classes we should keep the interface relatively stable which includes the constructor. Should we be using setters everywhere? I think this might be a topic for cross team.

PR #442, opened 2016-06-30
Add a shutdown service to provide more consistent shutdown between web and cli

joeferner:
- [x] @srfarley 

Features
- Any plugin can now register to be shutdown via `ShutdownService#register` method. I didn't want to use the service discovery pattern here because that could prematurely initialize plugins/classes during the shutdown phase.
- In `VisalloBootstrap` I register both `Graph` and `SimpleOrmSession` for shutdown so not more special handling of those classes.
- Shutdown order is in reverse order from startup order. So this should create a cleaner shutdown.

### DIFF
--- a/core/core-test/src/main/java/org/visallo/core/model/lock/LockRepositoryTestBase.java
+++ b/core/core-test/src/main/java/org/visallo/core/model/lock/LockRepositoryTestBase.java
@@ -2,6 +2,8 @@ package org.visallo.core.model.lock;
 
 import org.junit.After;
 import org.junit.Before;
+import org.visallo.core.util.ShutdownListener;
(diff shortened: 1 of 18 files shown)


srfarley:
There seems to be some duplication of the concept of "shutdown" between this PR and https://github.com/v5analytics/visallo/pull/186/. We shouldn't have two different ways to handle it.

joeferner:
@srfarley This PR actually works at a higher level than that PR. PR #186 manages individual threads within the worker and ensures they are properly shutdown. This PR ensures the repository/helper/service/etc, that might start threads under the covers, is shutdown. I think they are targeting two different scenarios.

PR #443, opened 2016-06-30
Fix bug with rendering edge comments, and factor out common code between vertex and edge set property routes into a helper class

srfarley:
- [x] @joeferner

The comment key is now consistent (a timestamp) between vertices and edges.

### DIFF
--- a/web/web-base/src/main/java/org/visallo/web/SetPropertyRouteHelper.java
+++ b/web/web-base/src/main/java/org/visallo/web/SetPropertyRouteHelper.java
@@ -1,0 +1,61 @@
+package org.visallo.web;
+
+import com.google.inject.Inject;
+import org.vertexium.Authorizations;
+import org.vertexium.Graph;
+import org.visallo.core.exception.VisalloException;
+import org.visallo.core.model.properties.VisalloProperties;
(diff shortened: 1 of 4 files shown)


joeferner:
Is there a big enough difference between Vertex and Edge that they couldn't just be subclasses of a `SetPropertyRouteBase` class?

srfarley:
There are enough differences where the base class would be the same as `SetPropertyRouteHelper` with protected instead of public methods. It doesn't matter to me either way.

joeferner:
I was just guessing that they should be a lot more similar and if they aren't why not. Why is an edge different from a vertex?

srfarley:
`VertexSetProperty` accepts multiple property values, has special logic for dependent properties, and saves multiple results which are returned as JSON. `EdgeSetProperty` broadcasts an workspace updates for its vertices, while `VertexSetProperty` broadcasts only itself. The `handle()` methods are quite different between the two.

jharwig:
Shouldn't edges handle dependent properties too?

srfarley:
I suppose they should. I guess we've never had the need.

srfarley:
I switched to using a base class for the routes so we can migrate towards common functionality.

PR #448, opened 2016-07-04
Clean up OntologyRepository cache functionality in Logout class

j782070772:
- [ ] @srfarley @joeferner

Add clean up OntologyRepository cache functionality in Logout class

if not ,while use OwlImport imported *.owl File,we Have to restart visallo to see concepts in the page


joeferner:
The Ontology cache is global for all users. Clearing the cache when any user logs out is probably not what you intended. Why do you need to clear the ontology cache? Could a button on the admin panel suffice?

j782070772:
because if it is not clear,when I import a owl file, in the  concept filter of the search page or in the graph page create entity, it don't display i imported concepts. i must have to restart the visallo. but you are right, Ontology cache is global, it shouldn't clear with any user. thanks ! my english is bad  :) @joeferner

j782070772:
and there has  AdminUploadOntology handler, But I don't know where on the page can trigger its url "/admin/upload-ontology"

joeferner:
We have features in our Enterprise offering that use that route. It will probably be removed from OS in a future release. You can create an admin plugin yourself using the [client API](http://docs.visallo.org/extension-points/front-end/admin/index.html). For an example you may want to look at [Admin RDF Import plugin](https://github.com/v5analytics/visallo/tree/master/web/plugins/admin-import-rdf).

j782070772:
Thank you very much.@joeferner

PR #452, opened 2016-07-05
Set text property visibility

jharwig:
- [x] @srfarley @joeferner

Since text properties aren't editable, there was no way to know or edit the visibility. This adds a button to view and edit visibility. Along with showing any other metadata.

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/graph/GraphRepository.java
+++ b/core/core/src/main/java/org/visallo/core/model/graph/GraphRepository.java
@@ -196,7 +196,8 @@ public class GraphRepository {
     ) {
         Visibility defaultVisibility = visibilityTranslator.getDefaultVisibility();
 
-        Property oldProperty = getProperty(element, propertyKey, propertyName, oldVisibilitySource, workspaceId);
+        Visibility oldPropertyVisibility = getVisibilityWithWorkspace(oldVisibilitySource, workspaceId);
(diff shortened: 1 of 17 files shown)


jharwig:
Waiting on 404 fix...

core/core/src/main/java/org/visallo/core/model/graph/GraphRepository.java
```diff
@@ -124,14 +196,7 @@ public void writeVersion(int version) {
...
 
-        Visibility oldPropertyVisibility = null;
-        if (oldVisibilitySource != null) {
-            VisibilityJson oldVisibilityJson = new VisibilityJson();
-            oldVisibilityJson.setSource(oldVisibilitySource);
-            oldVisibilityJson.addWorkspace(workspaceId);
-            oldPropertyVisibility = visibilityTranslator.toVisibility(oldVisibilityJson).getVisibility();
-        }
-
+        Visibility oldPropertyVisibility = getVisibilityWithWorkspace(oldVisibilitySource, workspaceId);
```
joeferner: Is this correct? The current logic only finds properties that are marked with the workspaceId in the visibility (ie private). With my new method `updatePropertyVisibilitySource` I need to first try and find the property with the workspaceId then try it without the workspaceId to really find it. I'm guessing this logic stays the same since we only want to work on sandboxed properties and if it's public we want to create a new one.

PR #456, opened 2016-07-06
change toolbar text to say Delete Edge instead of Delete Entity for e…

sfeng88:
- [x] @srfarley @joeferner

…dges


EvanOxfeld:
Should the toolbar text be relationship instead of entity?

sfeng88:
@EvanOxfeld fixed

PR #465, opened 2016-07-07
Move message bundle into core module and remove the visallo-messages module

joeferner:
- [x] @srfarley @joeferner

Fixes a dist issue where sometimes messages jar is not included. This moves messages into the core module/jar.


EvanOxfeld:
@joeferner Why was messages a separate module? Would it be less risky for the 2.0 release to have the  plugin that depends on messages add it as a dependency?

sfeng88:
@EvanOxfeld we couldn't figure out why we made it into a separate module yesterday.

joeferner:
@EvanOxfeld I would say it's a bit more risky without it in core because we might be missing other dist locations where it should be bundled as well.

PR #472, opened 2016-07-07
Keep track of open vertex previews

joeybrk372:
- [x] @srfarley @joeferner


web/war/src/main/webapp/js/util/popovers/detail/detail.js
```diff
@@ -22,6 +22,7 @@ define([
         });
 
         this.after('teardown', function() {
+            this.trigger('closePreviewVertex', { vertexId: this.attr.vertexId })
```
jharwig: Should move this to before('teardown') as it relies on bubbling the event and that should be down before teardown.

PR #484, opened 2016-07-12
Fix z-index calculation for multiple stacked modals

srfarley:
- [x] @joeferner


web/war/src/main/webapp/js/app.js
```diff
@@ -920,7 +920,11 @@ define([
 
     function fixMultipleBootstrapModals() {
         $(document).on('show.bs.modal', '.modal', function() {
-            var zIndex = 1040 + (10 * $('.modal:visible').length);
+            var zIndex = 10 + _.max(
+                _.map($('.modal'), function(modal) {
+                    return $(modal).zIndex();
```
jharwig: Can you change to  `parseInt($(modal).css('z-index'), 10)` so we aren't adding more jquery ui dependencies than we already have.

srfarley:
@jharwig Thanks. Done.

PR #485, opened 2016-07-12
Babel object rest support

jharwig:
- [x] @srfarley @joeferner

Allow spread object syntax in `jsx` files
```jsx
var component = <Component {...props} foo={'override'} />;
```

The previous client-side jsx support included [object rest/spread support](https://facebook.github.io/react/docs/jsx-spread.html). When we switched to Babel transpiling, the configuration was to convert jsx+es2015 syntax. But the object spread is not in es2015, so this PR adds a Babel plugin.

Because of the syntax change: 
* Requires eslint upgrade to support experimental react jsx object spread. `{ ...params }`, etc
* Eslint upgrade required grunt-eslint and some other grunt updates.
* Fix some files that were failing the new eslint config (I couldn't get a perfect mapping from old to new config)


web/war/src/main/webapp/js/graph/graph.js
```diff
@@ -1936,11 +1936,10 @@ define([
...
             this.cytoscapeReady()
                 .then(function(cy) {
-                    var offset = self.$node.offset(),
-                        r = cy.renderer(),
+                    var r = cy.renderer(),
                         pos = r.projectIntoViewport(
-                            data.pageX,// - offset.left,
-                            data.pageY// - offset.top
+                            data.pageX,
+                            data.pageY
```
srfarley: Should these changes be part of this PR? Are they related somehow?

jharwig: they were failing eslint. (I couldn't get the configuration to match exactly as before)

jharwig: My choice was either to loosen the rules, or fix the problem. But this (and duration.js) was the only failing

web/war/src/main/webapp/.babelrc
```diff
@@ -1,4 +1,4 @@
 {
   "presets": ["es2015"],
-  "plugins": ["transform-react-jsx", "transform-react-display-name"]
+  "plugins": ["transform-react-jsx", "transform-react-display-name", "transform-object-rest-spread"]
```
EvanOxfeld: Is there a reason we are using the `transform-react-jsx` plugin instead of the [react preset](http://babeljs.io/docs/plugins/preset-react/)? At minimum we also want `syntax-jsx` and `transform-jsx-display-name`.

EvanOxfeld:
Great -- Object rest spread is incredibly handy! 👍

srfarley:
Please check Travis. There are many errors.

jharwig:
@srfarley Working on it, forgot to update npm shrink-wrap so it's not upgrading the npm packages...

web/war/src/main/webapp/.babelrc
jharwig: I didn't want the flow stuff in the preset. And these plugins make jsx work with displayNames just fine.

PR #486, opened 2016-07-12
Fix vertex visibility updates on shared workspaces

joeybrk372:
- [ ] @srfarley @joeferner


jharwig:
Can we make it so that when the user is dragging items around that browser doesn't have to call `exists` in the store? Having 2 posts per drag stutters the interface.

I noticed that removing from workspace is broken (not sure if it's just this branch)

PR #487, opened 2016-07-12
add visibility source to TM vertex id if it exists

sfeng88:
- [x] @srfarley @joeferner

This also fixes the issue when the same text is resolved to multiple entities with different visibilities.

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/termMention/TermMentionBuilder.java
+++ b/core/core/src/main/java/org/visallo/core/model/termMention/TermMentionBuilder.java
@@ -1,5 +1,7 @@
 package org.visallo.core.model.termMention;
 
+import com.google.common.base.Charsets;
+import com.google.common.hash.Hashing;
 import org.vertexium.*;
 import org.vertexium.mutation.EdgeMutation;
 import org.visallo.core.model.properties.VisalloProperties;
@@ -247,6 +249,12 @@ public class TermMentionBuilder {
 
     private String createVertexId() {
         String id = TERM_MENTION_VERTEX_ID_PREFIX + this.outVertex.getId();
+        if (this.visibilityJson == null) {
+            LOGGER.warn ("Visibility Json should not be null");
+        } else if (this.visibilityJson.getSource() != null && this.visibilityJson.getSource().length() > 0) {
+            String hash = Hashing.sha1().hashString(this.visibilityJson.getSource(), Charsets.UTF_8).toString();
+            id += "-" + hash;
+        }
         if (this.propertyName != null) {
             id += "-" + this.propertyName;
         }


core/core/src/main/java/org/visallo/core/model/termMention/TermMentionBuilder.java
```diff
@@ -247,6 +247,9 @@ public Vertex save(Graph graph, VisibilityTranslator visibilityTranslator, User
 
     private String createVertexId() {
         String id = TERM_MENTION_VERTEX_ID_PREFIX + this.outVertex.getId();
+        if (this.visibilityJson != null && this.visibilityJson.getSource() != null && this.visibilityJson.getSource().length() > 0) {
+            id += "-" + this.visibilityJson.getSource();
```
joeferner: This isn't going to work very well when the source is very long such as a very big XML document generated by another system

joeferner: When would the visibilityJson not be filled in. Should this be required to save a TM?

srfarley: Why do we need a specially-crafted ID? Is it being parsed somewhere else?

sfeng88: @srfarley no it's not being parsed anywhere. 

@joeferner It should always be filled in, but I'm sure we have places where don't fill it in so I wanted to protect against that. Also wouldn't the visibility be longer?

srfarley: Why not use a randomly generated ID then?

sfeng88: @srfarley It can't be a randomly generated ID because we could potentially end up generating duplicate term mentions vertices. Let's say we re-queued all of the data it would generate new term mentions instead of using the old vertex if nothing has changed. It might be ok if we do generate duplicate TM vertices, but it would need QAing/regression testing.

https://github.com/v5analytics/visallo/blob/51cca978b7dd3a3f96c91fe28e8e873e4dffdfd0/core/core/src/main/java/org/visallo/core/model/termMention/TermMentionBuilder.java#L186

srfarley: @joeferner is right that we can't append the visibility source as-is onto the ID. Can you use a hash value instead?

`String hash = Hashing.sha1().hashString(visibilityJson.getSource(), Charsets.UTF_8).toString();`

sfeng88: @srfarley done

joeferner: Can we log a warn message stating visibility source should be filled in.

core/core/src/main/java/org/visallo/core/model/termMention/TermMentionBuilder.java
```diff
@@ -247,6 +249,10 @@ public Vertex save(Graph graph, VisibilityTranslator visibilityTranslator, User
 
     private String createVertexId() {
         String id = TERM_MENTION_VERTEX_ID_PREFIX + this.outVertex.getId();
+        if (this.visibilityJson != null && this.visibilityJson.getSource() != null && this.visibilityJson.getSource().length() > 0) {
+            String hash = Hashing.sha1().hashString(this.visibilityJson.getSource(), Charsets.UTF_8).toString();
+            id += "-" + hash;
+        }
```
joeferner: Can you add a warn log message if not filled in. Not having visibility source is not good.

sfeng88: I added a warning if visibility json is not filled in

PR #488, opened 2016-07-13
Add ConfigDirectoryStaticResourceHandler helper class to load files and serve them from the Visallo config directories

joeferner:
- [x] @srfarley


joeferner:
@srfarley added a static method to construct using inject helper.

PR #495, opened 2016-07-14
Fix NPE when calling toString on ClientApiWorkspace when users or vertices are null

joeferner:
- [x] @srfarley @joeferner

`users` and `vertices` are allowed to be null this just fixes a NPE when `toString` is called. Probably don't need to print out all users and vertices anyway


web/client-api/src/main/java/org/visallo/web/clientapi/model/ClientApiWorkspace.java
```diff
@@ -93,8 +93,6 @@ public String toString() {
                 ", isSharedToUser=" + isSharedToUser +
                 ", isEditable=" + isEditable +
                 ", active=" + active +
-                ", users=" + StringUtils.join(users) +
```
dsingley: prefer "users=<none>"

joeferner: I think `toString` should not be bloated. I'm more in favor of only providing the basic details.

PR #496, opened 2016-07-14
Fix UI issue if no creator of the workspace is found

joeferner:
- [x] @srfarley

The UI currently does not like (doesn't display the workspace in the list, infinite loading spinners) if a workspaces is found but the creator cannot be found. This PR allows viewing that workspace.

@jharwig not sure why you were requiring the creator, do you remember?

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
+++ b/core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
@@ -271,7 +271,9 @@ public abstract class WorkspaceRepository {
             workspaceClientApi.setTitle(workspace.getDisplayTitle());
 
             String creatorUserId = getCreatorUserId(workspace.getWorkspaceId(), user);
-            if (creatorUserId != null) {
+            if (creatorUserId == null) {
+                workspaceClientApi.setSharedToUser(true);
+            } else {
                 workspaceClientApi.setCreatedBy(creatorUserId);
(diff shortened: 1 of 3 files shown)


core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
```diff
@@ -271,7 +271,9 @@ public ClientApiWorkspace toClientApi(
             workspaceClientApi.setTitle(workspace.getDisplayTitle());
 
             String creatorUserId = getCreatorUserId(workspace.getWorkspaceId(), user);
-            if (creatorUserId != null) {
+            if (creatorUserId == null) {
+                workspaceClientApi.setSharedToUser(true);
```
srfarley: Why is `true` the default if you don't know the creator?

joeferner: true mean's this workspace was shared to you. If I don't know the creator I assume it wasn't created by you (ie you can't see the other user, the user was deleted, etc)

srfarley: Oh, I get it. If it wasn't shared to you, it wouldn't even be here.

jharwig:
I don't remember but if a user was deleted (that shared a workspace with you) all kinds of other issues happened, so I filtered the list. That might not be the case anymore.

PR #500, opened 2016-07-15
Fix published edge diffs being incorrectly filtered out

joeybrk372:
- [x] @srfarley @joeferner


core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
```diff
@@ -964,12 +964,6 @@ protected String convert(WorkspaceEntity workspaceEntity) {
                 .filter(vertex -> vertex != null)
                 .collect(Collectors.toList());
         Iterable<String> edgeIds = getGraph().findRelatedEdgeIdsForVertices(vertices, authorizations);
-        edgeIds = getGraph().filterEdgeIdsByAuthorization(
```
joeferner: Anyone remember why this was there.  We had to have a reason right?

PR #501, opened 2016-07-19
Allow workspaces without a created by to be used

joeferner:
- [x] @srfarley

If the system creates a workspace for the user. The workspace is created by system but system does not have a user vertex to point the created by to. This allows these types of workspaces


EvanOxfeld:
@joeferner Is this a duplicate of #496?

joeferner:
Yes, this was a PR that had #496 and other commits that have since been merged.

PR #508, opened 2016-07-25
Use babel to transpile all JavaScript, not just JSX.

jharwig:
- [x] @srfarley @joeferner

Requires removing the babel `preset-2015` and using all plugins except for commonjs modules, which breaks some files in our codebase until we move off requireJS. So, we can't use es2015 modules like `export default MyClass`

Includes source maps.


EvanOxfeld:
This is great -- thanks @jharwig!

PR #513, opened 2016-07-27
Reconnect admin user delete

joeybrk372:
- [x] @srfarley @joeferner

Also some styling fixes from post authorization refactor.


web/plugins/admin-user-tools/src/main/resources/org/visallo/web/adminUserTools/UserAdminPlugin.jsx
```diff
@@ -45,7 +45,8 @@ define([
         handleUserLoaded(user) {
             this.setState({
                 user: user,
-                reloadUser: false
+                reloadUser: false,
+                canDelete: user.id !== visalloData.currentUser.id
```
jharwig: rename var since you use it in a disabled attr later. The name doesn't match the expression here

PR #517, opened 2016-08-01
UserAdminCLI: Support setting authorizations and privileges from the user admin CLI

joeferner:
- [x] @srfarley


srfarley:
Is there an alternative to using the `CliPrivilegeRepository` and `CliAuthorizationRepository` interfaces? Why not just subclass `UserPropertyPrivilegeRepository ` and `UserPropertyAuthorizationRepository` to create CLI-aware implementations?

web/client-api/src/main/java/org/visallo/web/clientapi/model/Privilege.java
```diff
@@ -62,6 +62,14 @@ public static String toString(Collection<String> privileges) {
         return StringUtils.join(sortPrivileges(privileges));
     }
 
+    public static String toStringPrivileges(Collection<Privilege> privileges) {
```
srfarley: `Collection<Privilege>` can be `Iterable<Privilege>` since you don't need to modify the collection or know the size.

web/client-api/src/main/java/org/visallo/web/clientapi/model/Privilege.java
```diff
@@ -71,6 +79,11 @@ public static boolean hasAll(Set<String> userPrivileges, Set<String> requiredPri
         return true;
     }
 
+    @Override
+    public String toString() {
+        return name;
```
srfarley: Could call `getName()` here as an alternative.

web/client-api/src/main/java/org/visallo/web/clientapi/model/Privilege.java
joeferner: fixed

web/client-api/src/main/java/org/visallo/web/clientapi/model/Privilege.java
joeferner: fixed

joeferner:
@srfarley `RolesPrivilegeRepository` could also implement `CliPrivilegeRepository` to set roles from the command line. I guess the big question would be, should we just stuff all the CLI methods into `PrivilegeRepository` and you just have to implement them.

srfarley:
I think I'd rather see `CliPrivilegeRepository` and `CliAuthorizationRepository` as wrapper classes that delegate to their actual `PrivilegeRepository`/`AuthorizationRepository` classes and add somehow add the additional CLI functionality.

joeferner:
That would be a little weird from a configuration perspective. If you put `repository.privilege=NonCliUserPropPrivilegeRepository` you would get different behavior than `repository.privilege=CliUserPropPrivilegeRepository`. `CliPrivilegeRepository` is going to be a different implementation for each implementing `PrivilegeRepository`, you wouldn't be able to reuse it. For example the user prop priv repo needs to accept a list of privileges and validate those against the list of privs. The roles priv repo needs to accept a list of roles and validate against those. The project-xxxx repo may have built in privs and would not need to implement cli priv repo.

joeferner:
@srfarley refactored out cli stuff into new class. It feels much better now.

srfarley:
Yes, I do feel better now.   q(❂‿❂)p 👍

PR #519, opened 2016-08-01
added the server's host name to the ping vertex id

dsingley:
- [x] @srfarley @joeferner
- [x] @mwizeman @EvanOxfeld 

### DIFF
--- a/core/core/src/main/java/org/visallo/core/ping/PingOntology.java
+++ b/core/core/src/main/java/org/visallo/core/ping/PingOntology.java
@@ -1,12 +1,18 @@
 package org.visallo.core.ping;
 
-import org.visallo.core.model.properties.types.*;
+import com.google.common.base.Strings;
+import org.visallo.core.model.properties.types.DateSingleValueVisalloProperty;
+import org.visallo.core.model.properties.types.LongSingleValueVisalloProperty;
+import org.visallo.core.model.properties.types.StringSingleValueVisalloProperty;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 
 public class PingOntology {
     private static final String DATE_TIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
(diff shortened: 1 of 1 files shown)


core/core/src/main/java/org/visallo/core/ping/PingOntology.java
```diff
@@ -26,6 +32,14 @@
     public static final LongSingleValueVisalloProperty LONG_RUNNING_PROCESS_WAIT_TIME_MS = new LongSingleValueVisalloProperty(BASE_IRI + "#lrpWaitTimeMs");
 
     public static String getVertexId(Date date) {
-        return "PING_" + new SimpleDateFormat(DATE_TIME_FORMAT).format(date);
+        return "PING_" + new SimpleDateFormat(DATE_TIME_FORMAT).format(date) + "_" + HOST_NAME;
+    }
+
+    private static String getHostName() {
```
srfarley: This is going to throw an exception at class load time (when the static `HOST_NAME ` is initialized). That's a particularly insidious way to fail. How about instead of throwing an exception, we use the IP address. Actually, lets use both the IP address and hostname if both are available. If neither is available, use "Unknown address" as a placeholder rather than throwing the exception.

dsingley: i think it's very unlikely and if two hosts end up with "unknown" then we have the same runtime problem we see now

joeferner: Never say never... http://stackoverflow.com/questions/1881546/inetaddress-getlocalhost-throws-unknownhostexception

joeferner: I like this answer... http://stackoverflow.com/a/31408548/39431 or this one http://stackoverflow.com/a/20793241/39431

srfarley: Pick one. Just don't throw an exception.

joeferner: Made the update for @dsingley

PR #521, opened 2016-08-01
Add option requiredPrivilege to registerExtension to suppress extension based on privileges

joeferner:
- [x] @srfarley

This also shows the admin menu bar item if any admin items exist


jharwig:
@joeferner, I'd rather have this check in the admin list plugin rather than at the registry level. The consumer of the extensions should apply this logic, not the registry.

I just can't imagine this being needed and applied for every extension in the system and it expands the responsibilities of the registry into permissions, which I don't think belongs in the registry.

jharwig:
Also update the admin extension docs to include new parameter `requirePrivilege`

joeferner:
@jharwig can you have another look. I moved `requiredPrivilege` to the admin extension and documented.

PR #523, opened 2016-08-02
add methods to graph repository to support cases where you know the visibility json

sfeng88:
- [x] @srfarley @joeferner

### DIFF
--- a/core/core/src/main/java/org/visallo/core/ingest/graphProperty/TermMentionFilter.java
+++ b/core/core/src/main/java/org/visallo/core/ingest/graphProperty/TermMentionFilter.java
@@ -1,16 +1,18 @@
 package org.visallo.core.ingest.graphProperty;
 
 import com.google.inject.Inject;
-import org.visallo.core.config.Configuration;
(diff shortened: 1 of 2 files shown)


core/core/src/main/java/org/visallo/core/ingest/graphProperty/TermMentionFilter.java
```diff
@@ -31,6 +33,15 @@ protected final Graph getGraph() {
     }
 
     @Inject
+    public final void setGraphRepository(GraphRepository graphRepository) {
+        this.graphRepository = graphRepository;
```
srfarley: Are we now preferring setters over constructor injection?

sfeng88: I did it to be consistent with the rest of the class. Will convert all of the setters to constructor injection.

sfeng88: Decided to leave it as is for now.

PR #524, opened 2016-08-02
add rate to RDF triple import helper to help gauge time remaining

rygim:
- [ ] @srfarley @joeferner


joeferner:
Do we want to use https://github.com/v5analytics/visallo/blob/e9f0a0cf9c310171a09b83c1c22fb4b5104ac898/core/core/src/main/java/org/visallo/core/status/MetricsManager.java the timer metric would keep track of rate and count.

rygim:
maybe down the road but that is way too overkill/engineered for letting me watch progress on the command line.  Can we merge it now and convert it to the metrics manager later?

rygim:
I  am going to merge this because I need it for a deployment.  We can change it if needed later.

PR #526, opened 2016-08-03
Fix exceptions in iterator finalize

jharwig:
- [x] @srfarley @joeferner

Also only call filterAuths once per update

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
+++ b/core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
@@ -29,7 +29,9 @@ import org.visallo.web.clientapi.model.ClientApiWorkspace;
 import org.visallo.web.clientapi.model.UserStatus;
 
 import java.util.*;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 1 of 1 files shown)


joeferner:
I don't love the previous implementation because it calls `filterVertexIdsByAuthorization` for every vertex. But isn't your new logic essentially doing the same checks? How are they different.

core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
```diff
@@ -735,14 +737,15 @@ protected void broadcastWorkspace(
...
-                                ElementFilter.ALL,
-                                authorizations);
-
-                        return filtered.iterator().hasNext();
-                    }).collect(Collectors.toList());
+            List<String> vertexIds = workspace.getVertices().stream()
+                    .map(vertex -> vertex.getVertexId())
+                    .collect(Collectors.toList());
+            Iterable<String> visibleVertexIds = graph.filterVertexIdsByAuthorization(vertexIds, workspace.getWorkspaceId(), ElementFilter.ALL, authorizations);
+            Map<String, ClientApiWorkspace.Vertex> workspaceVerticesByVertexId = workspace.getVertices().stream()
```
joeferner: Would it make the next couple lines a little easier to understand by filtering `workspace.getVertices()` using `visibleVertexIds.contains(...)` instead of going to a `Map` then a `spliterator`?

jharwig:
I was getting finalize exceptions, then SQL connection timeouts before

core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
jharwig: yes, I could do that

jharwig: Something like this?
```
            List<ClientApiWorkspace.Vertex> filteredVertices = workspace.getVertices()
                    .stream()
                    .filter(vertex -> Iterables.contains(visibleVertexIds, vertex.getVertexId()))
                    .collect(Collectors.toList());
```

jharwig: Actually, that is causing the same issues I was trying to fix...


Caused by: java.sql.SQLTransientConnectionException: vertexium - Connection is not available, request timed out after 30003ms.
	at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:196) ~[HikariCP-2.4.6.jar:na]
	at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:147) ~[HikariCP-2.4.6.jar:na]
	at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:83) ~[HikariCP-2.4.6.jar:na]
	at org.skife.jdbi.v2.DataSourceConnectionFactory.openConnection(DataSourceConnectionFactory.java:36) ~[jdbi-2.63.1.jar:2.63.1]
	at org.skife.jdbi.v2.DBI.open(DBI.java:212) ~[jdbi-2.63.1.jar:2.63.1]

jharwig: Do I have to check if it's a CloseableIterator and call close?

jharwig: okay, I'll try toList

jharwig:
@joeferner okay, made the change. `Iterables.toList` works

PR #542, opened 2016-08-08
check acl of ontology property if acl doesn't exist on property

sfeng88:
- [x] @srfarley @joeferner

We weren't accounting for situations where the acl doesn't exist on `config.properties` (e.g compound properties).

### DIFF
--- a/web/war/src/main/webapp/js/util/popovers/propertyInfo/propertyInfo.js
+++ b/web/war/src/main/webapp/js/util/popovers/propertyInfo/propertyInfo.js
@@ -39,13 +39,13 @@ define([
                 config.canAdd = config.canEdit = config.canDelete = false;
 
                 if (config.isComment && visalloData.currentWorkspaceCommentable) {
-                    config.canAdd = config.property.addable !== false;
-                    config.canEdit = config.property.updateable !== false;
-                    config.canDelete = config.property.deleteable !== false;
+                    config.canAdd = config.property.addable ? config.property.addable !== false : config.ontologyProperty.addable !== false;
+                    config.canEdit = config.property.updateable ? config.property.updateable !== false : config.ontologyProperty.updateable !== false;
+                    config.canDelete = config.property.deleteable ? config.property.deleteable !== false : config.ontologyProperty.deleteable !== false;
                 } else if (!config.isComment && visalloData.currentWorkspaceEditable) {
-                    config.canAdd = config.property.addable !== false;
-                    config.canEdit = config.property.updateable !== false;
-                    config.canDelete = config.property.deleteable !== false &&
+                    config.canAdd = config.property.addable ? config.property.addable !== false : config.ontologyProperty.addable !== false;
+                    config.canEdit = config.property.updateable ? config.property.updateable !== false : config.ontologyProperty.updateable !== false;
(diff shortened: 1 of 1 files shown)


web/war/src/main/webapp/js/util/popovers/propertyInfo/propertyInfo.js
```diff
@@ -39,13 +39,13 @@ define([
                 config.canAdd = config.canEdit = config.canDelete = false;
 
                 if (config.isComment && visalloData.currentWorkspaceCommentable) {
-                    config.canAdd = config.property.addable !== false;
-                    config.canEdit = config.property.updateable !== false;
-                    config.canDelete = config.property.deleteable !== false;
+                    config.canAdd = config.property.addable ? config.property.addable !== false : config.ontologyProperty.addable !== false;
```
joeferner: If `config.property.addable` is false should it fall back to `config.ontologyProperty.addable`?

srfarley:
I question the use of the ontology for ACL decisions. (That only applies when `OntologyACLProvider` is being used, and it should be handled exclusively on the back end.) It's an error to have missing ACL properties. This should be fixed on the back end.

sfeng88:
**Situation:**
Ontology has a compound property `name` that contains `first name` and `last name`. The `name` property is not updateable. For some reason, Vertex A has a `first name` property set. On the detail pane, it shows Vertex A has a `name` property with the `first name` property value. The user can "edit" the name property but when the user clicks save, we throw an error.

The name property doesn't actually exist on Vertex A so we can't put the ACL on the property object, therefore we default to the ontology property.
https://github.com/v5analytics/visallo/blob/master/core/core/src/main/java/org/visallo/core/security/ACLProvider.java#L199

What should we do in this case?

sfeng88:
@srfarley I fixed it to look at the master property acl list.

web/war/src/main/webapp/js/util/popovers/propertyInfo/propertyInfo.js
```diff
@@ -31,21 +31,33 @@ define([
...
+                return prop[0];
+            };
+
             config.template = 'propertyInfo/template';
             config.isFullscreen = visalloData.isFullscreen;
             if (config.property) {
 
                 config.isComment = config.property.name === 'http://visallo.org/comment#entry';
                 config.canAdd = config.canEdit = config.canDelete = false;
+                var allAclProperties = config.data.acl.propertyAcls;
```
srfarley: Should be `allPropertyAcls`.

web/war/src/main/webapp/js/util/popovers/propertyInfo/propertyInfo.js
```diff
@@ -31,21 +31,33 @@ define([
         });
 
         this.before('initialize', function(node, config) {
+             var findPropertyAcl = function(propertiesAcl, propName) {
+                var prop = _.filter(propertiesAcl, function (p){
```
srfarley: Nitpick: space before `{`.

srfarley:
@sfeng88 See comments. Just minor issues. Otherwise 👍

sfeng88:
@srfarley done :)

PR #546, opened 2016-08-09
Fix flashing of right aligned numbers in multiple selection histogram (master)

jharwig:
- [x] @srfarley @joeferner

… detail pane.

The svg is set to be 100% of the detail pane width, IE doesn't update those elements sizes until after mouseup. So the numbers were being pushed offscreen for a little.

Use a media query to target ie 10+ to fix IE11, and force the scrollbar to stick around so the width isn't changing based on scrollbar visibility.

### DIFF
--- a/web/war/src/main/webapp/less/detail/detail.less
+++ b/web/war/src/main/webapp/less/detail/detail.less
@@ -371,6 +371,13 @@
     }
   }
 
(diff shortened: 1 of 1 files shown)


EvanOxfeld:
@jharwig Can you please resubmit against the `release-2.1` branch? Otherwise LGTM.

PR #547, opened 2016-08-09
Limit available property sorts and filters in search

joeybrk372:
- [x] @srfarley @joeferner

Can only sort/filter based on properties that are on vertices/edges depending on which match type you have.

Will now limit available relationship properties if there is a relationship type selected(vertex search filters already did this previously)

### DIFF
--- a/web/war/src/main/webapp/js/data/web-worker/services/ontology.js
+++ b/web/war/src/main/webapp/js/data/web-worker/services/ontology.js
@@ -68,6 +68,25 @@ define([
                     });
             }),
 
+            propertiesByDomainType: memoize(function(type) {
+                return getOntology()
+                    .then(function(ontology) {
+                        var items = (type === 'concept' || type === 'vertex') ? ontology.concepts : ontology.relationships;
+
+                        return _.chain(items)
+                            .pluck('properties')
+                            .compact()
+                            .flatten()
+                            .uniq()
+                            .map(function(propertyName) {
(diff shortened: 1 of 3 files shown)


web/war/src/main/webapp/js/data/web-worker/services/ontology.js
```diff
@@ -68,6 +68,29 @@ define([
                     });
             }),
 
+            propertiesByDomainType: memoize(function(type) {
+                return getOntology()
+                    .then(function(ontology) {
+                        var items = (type === 'concept' || type === 'vertex') ? ontology.concepts : ontology.relationships;
+                        var properties = [];
```
jharwig: use map instead of foreach/push
```
var properties = items.map(function(i) {
    return i.properties;
})
```

web/war/src/main/webapp/js/data/web-worker/services/ontology.js
```diff
@@ -68,6 +68,29 @@ define([
...
+
+                        items.forEach(function(item) {
+                            properties.push(item.properties);
+                        });
+
+                        return _.chain(properties)
```
jharwig: actually instead of previous comment, just combine it into here
```
_.chain(items)
.pluck('properties')
...
```

EvanOxfeld:
@joeybrk372 Would you please merge this into `release-2.1` as well?

PR #552, opened 2016-08-10
Fix saved search issues (release-2.1)

srfarley:
- [ ] @joeferner

===
* Fix issues so saved searches can switch between global and private with the appropriate user privilege
* Add more tests for global saved searches
* Return all privileges using Privilege.ALL
* Add SEARCH_SAVE_GLOBAL to UserPropertyPrivilegeRepository.defaultPrivileges config property

### DIFF
--- a/web/client-api/src/main/java/org/visallo/web/clientapi/model/Privilege.java
+++ b/web/client-api/src/main/java/org/visallo/web/clientapi/model/Privilege.java
@@ -15,7 +15,19 @@ public final class Privilege {
     public static final String PUBLISH = "PUBLISH";
     public static final String ADMIN = "ADMIN";
 
-    public static final Set<String> NONE = Collections.emptySet();
+    static {
+        // NOTE: Keep allNames in sync with the above public static strings.
+        final String[] allNames = new String[] {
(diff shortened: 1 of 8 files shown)


web/client-api/src/main/java/org/visallo/web/clientapi/model/Privilege.java
```diff
@@ -15,7 +15,19 @@
...
+                READ, COMMENT, COMMENT_EDIT_ANY, COMMENT_DELETE_ANY, SEARCH_SAVE_GLOBAL, EDIT, PUBLISH, ADMIN
+        };
+        Set<Privilege> allPrivileges = new HashSet<Privilege>(allNames.length);
+        for (String name : allNames) {
+            allPrivileges.add(new Privilege(name));
+        }
+        ALL = Collections.unmodifiableSet(allPrivileges);
+    }
+
+    public static final Set<Privilege> ALL;
```
joeferner: I don't love calling it ALL here since it is only all built in

srfarley: Good point. How about `ALL_BUILT_IN` or `ALL_CORE`?

joeferner: Works for me

srfarley:
@joeferner Renamed to `ALL_BUILT_IN`.

PR #556, opened 2016-08-10
fix accumulo security exception when access is deleted from another user

rygim:
- [ ] @srfarley @joeferner

Fix for https://github.com/v5analytics/visallo-lts/issues/160

When a workspace is deleted, accumulo throws an exception if trying to use an invalid authorization.  That invalid authorization is used while trying to generate the authorizations for the /workspace GET call.  The UI would then fail and the workspace would not disappear

Fixing that problem caused a problem where /workspace/all would also run into troubles getting the workspaces, so that method is slightly modified to eat the exception and just use the normal authorizations if possible.


web/web-base/src/main/java/org/visallo/web/routes/workspace/WorkspaceList.java
```diff
@@ -34,10 +35,17 @@ public ClientApiWorkspaces handle(
...
-            authorizations = authorizationRepository.getGraphAuthorizations(user, workspaceId);
-        } else {
+        Authorizations authorizations = null;
+        try {
+            if (workspaceId != null && workspaceRepository.hasReadPermissions(workspaceId, user)) {
+                authorizations = authorizationRepository.getGraphAuthorizations(user, workspaceId);
+            }
+        }
+        catch(SecurityVertexiumException e){
+            //do nothing
```
joeferner: I don't like this. I'm guessing this is from the call to hasReadPermissions. Can you restructure the code to only wrap that call. Maybe a new method 'hasAccess' and return false in the catch.

web/web-base/src/main/java/org/visallo/web/parameterProviders/AuthorizationsParameterProviderFactory.java
```diff
@@ -45,20 +47,25 @@ public static Authorizations getAuthorizations(
...
+                    );
+                }
+                return authorizationRepository.getGraphAuthorizations(user, workspaceId);
             }
-            return authorizationRepository.getGraphAuthorizations(user, workspaceId);
-        }
 
-        return authorizationRepository.getGraphAuthorizations(user);
+            return authorizationRepository.getGraphAuthorizations(user);
+        }
```
joeferner: Code formatting

web/web-base/src/main/java/org/visallo/web/parameterProviders/AuthorizationsParameterProviderFactory.java
```diff
@@ -45,20 +47,25 @@ public static Authorizations getAuthorizations(
...
-        String workspaceId = VisalloBaseParameterProvider.getActiveWorkspaceIdOrDefault(request);
-        User user = VisalloBaseParameterProvider.getUser(request, userRepository);
-        if (workspaceId != null) {
-            if (!workspaceRepository.hasReadPermissions(workspaceId, user)) {
-                throw new VisalloAccessDeniedException(
-                        "You do not have access to workspace: " + workspaceId,
-                        user,
-                        workspaceId
-                );
+        try {
```
joeferner: Is it possible to narrow the scope of this try/catch? What call is actually throwing?

web/web-base/src/main/java/org/visallo/web/parameterProviders/AuthorizationsParameterProviderFactory.java
```diff
@@ -45,20 +47,25 @@ public static Authorizations getAuthorizations(
...
+                return authorizationRepository.getGraphAuthorizations(user, workspaceId);
             }
-            return authorizationRepository.getGraphAuthorizations(user, workspaceId);
-        }
 
-        return authorizationRepository.getGraphAuthorizations(user);
+            return authorizationRepository.getGraphAuthorizations(user);
+        }
+        catch(SecurityVertexiumException e){
+            throw new VisalloAccessDeniedException(ExceptionUtils.getStackTrace(e), null, null);
```
joeferner: Can you not call get stack trace here. I think that will make it all the way to the front end.

web/web-base/src/main/java/org/visallo/web/routes/workspace/WorkspaceList.java
rygim: Good idea, Changed it in the diff

web/web-base/src/main/java/org/visallo/web/parameterProviders/AuthorizationsParameterProviderFactory.java
rygim: fixed

web/web-base/src/main/java/org/visallo/web/parameterProviders/AuthorizationsParameterProviderFactory.java
rygim: Done

web/web-base/src/main/java/org/visallo/web/parameterProviders/AuthorizationsParameterProviderFactory.java
rygim: Sounds good, I made it return a similar message to the exception just above it

web/web-base/src/main/java/org/visallo/web/parameterProviders/AuthorizationsParameterProviderFactory.java
```diff
@@ -45,16 +47,26 @@ public static Authorizations getAuthorizations(
...
-            if (!workspaceRepository.hasReadPermissions(workspaceId, user)) {
+            try {
+                if (!workspaceRepository.hasReadPermissions(workspaceId, user)) {
+                    throw new VisalloAccessDeniedException(
+                            "You do not have access to workspace: " + workspaceId,
+                            user,
+                            workspaceId
+                    );
+                }
+            } catch (RuntimeException e) {
```
jharwig: Does this need to be RuntimeException (what's wrong with using VertexiumSecurityException?)

Some other runtime exception like can't connect to accumulo would cause the workspace to be removed from users page.

rygim: good catch, when I was looking at it, the stack trace told me it was a RuntimeException  so I caught that, but failed to notice that it was actually a It actually throws a VertexiumSecurityException which is a subclass of that.  I ran an experiment and it turns out I can catch the VertexiumSecurityException  and it should work just fine.  Making sure it works and submitting a fix in a sec

sfeng88:
@rygim I cherry picked these commits to master

PR #558, opened 2016-08-11
Add filters to aggregation property selects and check for mapzen(release-2.1)

joeybrk372:
- [ ] @srfarley @joeferner

- check if visallo has access to mapzen before adding geo-coordinate cluster aggregation option
- allow aggregations to specify their own filter for selectable properties
- add filters to geo-coordinate cluster(only location properties) and histogram(only date properties)
- change geo-coordinate cluster and histogram property placeholders to ‘Location Property’ and ‘Date Property’ respectively


web/war/src/main/webapp/js/search/dashboard/aggregation.js
```diff
@@ -87,11 +87,30 @@ define([
             this.on('propertyselected', this.onPropertySelected);
             this.on('filterProperties', this.onFilterProperties);
 
-            this.$node.html(template({
-                aggregations: AGGREGATIONS,
-                precisions: PRECISIONS,
-                intervalUnits: INTERVAL_UNITS
-            }));
+            d3.json('mapzen/osm/all/0/0/0.json', function(error, json) {
```
jharwig: this will make a request on every configuration open. Extract to function `this.mapzenSupported()`

```
this.mapzenSupported = function() {
   if (!_mapzenPromise) { // some file level variable that persists across configuration instances
      return new Promise(function(f) {
          ... // make d3 json request and return true/false
      });
   }

   return _mapzenPromise;
}
```

PR #561, opened 2016-08-11
export rdf fixes to export data on graph and then be able to reimport…

rygim:
- [ ] @joeferner
- [ ] @srfarley @kunklejr

CHANGELOG
Added: 
Changed: 
Documentation: 
Deprecated:
Removed:
Fixed:
Security:


web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexRaw.java
```diff
@@ -158,7 +158,7 @@ private void copy(InputStream in, OutputStream out, Long length) throws IOExcept
     }
 
     private String getMimeType(Vertex artifactVertex) {
-        String mimeType = VisalloProperties.MIME_TYPE.getOnlyPropertyValue(artifactVertex);
+        String mimeType = VisalloProperties.MIME_TYPE.getFirstPropertyValue(artifactVertex);
```
rygim: This was changed because there ended up being more than one mimetype when rdf files are imported.  its more of a bandaid fix to just take the first property value than checking all of them (in case it gets changed somehow) but it ends up working

common/rdf/src/main/java/org/visallo/common/rdf/SetMetadataVisalloRdfTriple.java
```diff
@@ -89,17 +89,8 @@ public ImportContext updateImportContext(
...
             Authorizations authorizations
     ) {
-        // Currently Vertexium only supports updating metadata on ExistingElementMutation if that ever
-        //  changes this logic can be removed, createImportContext can be renamed createImportContext
-        //  and ImportContext can be removed from the parameter list
-        if (!(ctx.getElementMutation() instanceof ExistingElementMutation)) {
-            ctx.save(authorizations);
-
-            Element element = getExistingElement(rdfTripleImportHelper.getGraph(), this, authorizations);
+            Element element = ctx.save(authorizations);
```
rygim: @joeferner I am 100% sure that this is a horrible, slow way to fix this, but the element that I was getting at the end of the import didn't have all of the properties on it so I couldnt add metadata to them.  Do you have any suggestions other than saving it after every line?

common/rdf/src/main/java/org/visallo/common/rdf/PropertyVisalloRdfTriple.java
```diff
@@ -104,7 +108,19 @@ protected String getValueRdfString() {
...
-        LOGGER.warn ("\"Unhandled value type " + getValue().getClass().getName() + " to convert to RDF string\"");
+        if (getValue() instanceof StreamingPropertyValue) {
+            StreamingPropertyValue value = (StreamingPropertyValue) getValue();
+            //do we want to limit this in some way?  Could possibly run the server out of memory
+            byte[] b = new byte[(int) value.getLength()];
+            try {
+                IOUtils.readFully(value.getInputStream(), b);
+                return getValueRdfStringWithType(new String(Base64.encodeBase64(b)), VisalloRdfTriple.PROPERTY_TYPE_STREAMING_PROPERTY_VALUE_INLINE_BASE64);
+            } catch (IOException e) {
+                e.printStackTrace();
```
joeferner: You should throw a `new VisalloException("Could not export, blah blah", e)`

common/rdf/src/main/java/org/visallo/common/rdf/PropertyVisalloRdfTriple.java
```diff
@@ -104,7 +108,19 @@ protected String getValueRdfString() {
...
+            }
+        }
+
+        LOGGER.warn("\"Unhandled value type " + getValue().getClass().getName() + " to convert to RDF string\"");
```
joeferner: Not part of this PR but while you're here, can you change this to use `%s` string format built into the LOGGER.

common/rdf/src/main/java/org/visallo/common/rdf/SetMetadataVisalloRdfTriple.java
joeferner: Can you move this change out of this PR. Can you replicate this problem in a unit test?

web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexRaw.java
joeferner: This change works for me. I think getting the only property value would have bitten us in other ways, like two GPWs identifying a mime type for example.

sfeng88:
@rygim  please add to changelog

rygim:
closing because this is not a priority any more and the changes would take a little for me to make

rygim:
Reopening after talking to @EvanOxfeld

common/rdf/src/main/java/org/visallo/common/rdf/PropertyVisalloRdfTriple.java
```diff
@@ -104,7 +108,19 @@ protected String getValueRdfString() {
         if (getValue() instanceof DateOnly) {
             return getValueRdfStringWithType(getValue(), VisalloRdfTriple.PROPERTY_TYPE_DATE);
         }
-        LOGGER.warn ("\"Unhandled value type " + getValue().getClass().getName() + " to convert to RDF string\"");
+        if (getValue() instanceof StreamingPropertyValue) {
```
kunklejr: Consider calling getValue() once at the top of the method instead of calling it for every if block.

common/rdf/src/main/java/org/visallo/common/rdf/PropertyVisalloRdfTriple.java
```diff
@@ -104,7 +108,19 @@ protected String getValueRdfString() {
...
+            StreamingPropertyValue value = (StreamingPropertyValue) getValue();
+            //do we want to limit this in some way?  Could possibly run the server out of memory
+            byte[] b = new byte[(int) value.getLength()];
```
kunklejr: Are we just hoping the length will never be greater than available memory?

common/rdf/src/main/java/org/visallo/common/rdf/PropertyVisalloRdfTriple.java
rygim: its sort of a preference thing, I dont have a problem with the method being called so I just followed the pattern that was already established.  It will get pushed into something that looks like this when the JVM is optimized so I think it is really just a style thing

common/rdf/src/main/java/org/visallo/common/rdf/PropertyVisalloRdfTriple.java
rygim: For now, yes.  Requests that pull values back that are longer will throw an exception and so whatever calls this just needs to be prepared to handle that exception.

EvanOxfeld:
Closing temporarily while @rygim waits for feedback.

PR #564, opened 2016-08-12
Allow ACLProvider to continue even if a concept or relationship cannot be found.

joeferner:
- [x] @srfarley @kunklejr 

Currently if a vertex or edge is create with a given concept or relationship and
that concept or relationship ontology item is later removed ACLProvider throws
an exception causing the UI to become unusable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ==================
 
   * Add methods which take timestamps to SingleValueVisalloProperty and VisalloProperty
+  * Allow ACLProvider to continue even if a concept or relationship cannot be found
 
 #v2.1.1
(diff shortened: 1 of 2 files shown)


sfeng88:
@joeferner If the concept or relationship doesn't exist, won't it cause issues in the UI (e.g detail pane)? We check for the ACL on the UI in a couple of places (e.g property info on the detail pane).

joeferner:
Right now everything breaks back and front this PR fixes the backend. I created an issue for the front-end

sfeng88:
@joeferner please add to the changelog and then merge

joeferner:
@sfeng88 done and merging

PR #565, opened 2016-08-12
Use a more specific selector in the create connection popover

joeferner:
- [x] @srfarley @kunklejr

This allows the use of the `btn-link` class within the popover contents.


kunklejr:
This seems reasonable to me. Why problem is it solving?

joeferner:
@kunklejr I have a plugin which has a link in the authorization plugin which was triggering the invert

PR #568, opened 2016-08-12
Fix aggregations select list not having all valid options (release-2.1)

jharwig:
- [ ] @joeferner
- [x] @srfarley @kunklejr

After checking for mapzen, we mutated the static list of aggregations. Create 2 static lists and switch between them.

### DIFF
--- a/web/war/src/main/webapp/js/search/dashboard/aggregation.js
+++ b/web/war/src/main/webapp/js/search/dashboard/aggregation.js
@@ -26,6 +26,9 @@ define([
             }},
             { value: 'statistics', name: 'Statistics' }
         ],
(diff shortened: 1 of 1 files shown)


web/war/src/main/webapp/js/search/dashboard/aggregation.js
```diff
@@ -90,13 +93,6 @@ define([
 
             this.mapzenSupported()
                 .then(function(mapzen) {
-                    if (!mapzen) {
-                        var index = AGGREGATIONS.findIndex(function(a) {
-                            return a.value === 'geohash';
-                        });
-                        AGGREGATIONS.splice(index, 1);
```
jharwig: this would be `-1` after the second aggregation rendered, so would remove the last item

PR #572, opened 2016-08-12
throw visallo access denied exception when there a security exception(master)

rygim:
- [ ] @joeferner
- [ ] @srfarley @kunklejr

Fixes another variant of the bad authorizations error when the diff panel tries to diff a workspace the user no longer has access for.


joeferner:
Couldn't Vertexium security exceptions happen all over should we move this logic to the same place we catch Visallo security exceptions?

PR #586, opened 2016-08-19
Admin: Allow extensions to request a sort within a section

EvanOxfeld:
- [x] @joeferner
- [x] @srfarley @kunklejr

Enables admin items to request an exact position within a section of the admin pane.

Depends on #585 (polyfills `Number.isInteger` in IE)

### DIFF
--- a/web/war/src/main/webapp/js/admin/admin.js
+++ b/web/war/src/main/webapp/js/admin/admin.js
@@ -162,12 +162,13 @@ define([
                             return d[0];
                         })
                         .each(function(d) {
-                            const [sorted, toSort] = _.partition(d[1], ({ options }) => options && Number.isInteger(options.sortHint));
-                            d[1] = _.sortBy(toSort, 'name');
-                            sorted.forEach(ext => {
-                                const { sortHint } = ext.options;
-                                d[1].splice(sortHint, 0, ext);
-                            });
+                            d[1] = _.chain(d[1])
+                                .sortBy('name')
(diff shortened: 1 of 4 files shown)


web/war/src/main/webapp/js/admin/admin.js
```diff
@@ -162,7 +162,12 @@ define([
                             return d[0];
                         })
                         .each(function(d) {
-                            d[1] = _.sortBy(d[1], 'name');
+                            const [sorted, toSort] = _.partition(d[1], ({ options }) => options && Number.isInteger(options.sortHint));
+                            d[1] = _.sortBy(toSort, 'name');
```
jharwig: Would prefer multiple sort by rather than this splicing. Underscores sortBy is stable.

```
_.chain(d[1])
.sortBy('name')
.sortBy(function sortHint({options}) {
    return options && Number.isInteger(options.sortHint) ?
        options.sortHint :
        Number.MAX_VALUE;
})
.value()

jharwig: Also verify Number.isInteger is polyfilled because it's nots supported in IE.

Or use _.isNumber

EvanOxfeld: Thanks -- reimplemented as multiple sort.

`Number.isInteger` is polyfilled by babel-polyfill. See #585 for details. I've verified that this change works in IE.

PR #593, opened 2016-08-22
Authorization React component did not update auths when switching users (2.1)

rygim:
- [x] @joeferner
- [x] @srfarley @kunklejr


srfarley:
Can we please have a more descriptive title?

rygim:
Not done making it Steve, one sec

rygim:
k all done @srfarley :)

srfarley:
Thanks @rygim. Sorry I jumped the gun :-)

web/plugins/admin-user-property-auth/src/main/resources/org/visallo/web/plugin/adminUserTools/userPropertyAuth/UserAdminAuthorizationPlugin.jsx
```diff
@@ -33,6 +33,12 @@ define([
                 })
         },
 
+        componentWillReceiveProps(nextProps) {
+            this.setState({
+                authorizations: nextProps.user.authorizations
+            });
```
EvanOxfeld: When a `userName` or `authorizations` prop changes, should the component reset back to its initial state e.g. reset `state.error`, `state.saveInProgress`, and `state. addAuthorizationValue`?

rygim: I think so.  The problem that is being solved here is that the state wasn't being updated from the new user that was selected.  From a functional standpoint, when a new user is selected, I consider that to be a whole new view of the authorizations of that user so I want the state set to only what the current user sees.
That being said, I am open to discussing what that means and different implications of how it works.

rygim:
@EvanOxfeld talked about resetting the state fully last night so I pushed a change to both branches that does just that

PR #599, opened 2016-08-23
Fix propertySelect not disabling when no properties were available(2.1 cherry-pick)

joeybrk372:
- [x] @joeferner
- [x] @srfarley @kunklejr

Also fix i property search

CHANGELOG
Fixed: Property select field not disabling properly when there are no available properties.

### DIFF
--- a/web/war/src/main/webapp/js/util/ontology/propertySelect.js
+++ b/web/war/src/main/webapp/js/util/ontology/propertySelect.js
@@ -28,7 +28,6 @@ define([
         this.after('initialize', function() {
             var self = this;
 
-            this.updatePropertiesSource();
             this.on('filterProperties', this.onFilterProperties);
 
             if (this.attr.selectedProperty) {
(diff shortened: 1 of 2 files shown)


web/war/src/main/webapp/js/util/ontology/propertySelect.js
```diff
@@ -265,7 +232,57 @@ define([
...
+            this.select('findPropertySelection')
+                .attr('placeholder', i18n(placeholderMessage))
+                .attr('disabled', !hasProperties);
+        };
+
+        this.filterSourceProperties = function(properties) {
+            var self = this;
+
+            return _.chain(properties)
+                .filter(function(p) {
```
EvanOxfeld: As `filter` is the only step in the chain, assuming `properties` is an array, I'd recommend using `Array.prototype.filter` over `_.chain` and/or `_.filter`.

web/war/src/main/webapp/js/util/ontology/propertySelect.js
```diff
@@ -76,66 +75,32 @@ define([
...
+                                .sortBy(function(itemJson) {
+                                    var item = JSON.parse(itemJson),
+                                        lower = item.displayName.toLowerCase();
+
+                                    if (item.propertyGroup) {
+                                        return '1' + item.propertyGroup + lower;
+                                    }
+                                    return '0' + lower;
+                                })
+                                .value();
```
EvanOxfeld: I like how `filterSourceProperties` is broken out into its own function above. I'd recommend refactoring one step further using a more declarative approach by splitting out the filter, map, and sortBy functions.

For example:

```
return _.chain(self.propertiesForSource)
  .filter(filterBySourceProperties)
  .map(mapProperty)
  .sortBy(sortByPropertyGroupAndDisplayName)
  .value();
```

web/war/src/main/webapp/js/util/ontology/propertySelect.js
```diff
@@ -265,7 +236,57 @@ define([
...
+            this.select('findPropertySelection')
+                .attr('placeholder', i18n(placeholderMessage))
+                .attr('disabled', !hasProperties);
+        };
+
+        this.filterSourceProperties = function(properties) {
+            var self = this;
+
+            return _.chain(properties)
+                .filter(function(p) {
```
EvanOxfeld: This `filter` is the only item in the chain, which is what led to my suggestion to refactor `filterSourceProperties` back into the `typeahead`'s `source` function.

Alternatively, I recommend using `Array.prototype.filter` over a single item `_.chain`. `_.chain` adds overhead https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba#.2qttwkyck

jharwig: 👍 Prefer `Array.prototype.map` and `filter` if only one step.

Eventually we could move to transducers that alleviate the multiple passes through the array using `Array.prototype.reduce` and move all the way to FP.

https://medium.com/@roman01la/understanding-transducers-in-javascript-3500d3bd9624#.ziildwllq

PR #602, opened 2016-08-24
Filter out string properties from histogram aggregation(2.1 cherry-pick)

joeybrk372:
- [x] @joeferner
- [ ] @srfarley @kunklejr

### DIFF
--- a/core/core/src/main/resources/MessageBundle.properties
+++ b/core/core/src/main/resources/MessageBundle.properties
@@ -526,7 +526,6 @@ dashboard.report.pie.more=and {0} more...
 dashboard.text-overview.limits.other=Other
 dashboard.text-overview.limits.none=No Limit
 dashboard.search.aggregation.geohash.property.placeholder=Location Property
-dashboard.search.aggregation.histogram.property.placeholder=Date Property
 dashboard.search.results.open.button=Open in Search...
 dashboard.search.results.none=No Results Found
 dashboard.search.results.error=Server Error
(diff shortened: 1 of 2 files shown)


joeferner:
This PR looks to allow any field type. We can't histogram on any type. We can only histogram on dates and numbers.

joeybrk372:
At @mwizeman 's suggestion, changed this pr to block strings. There are still some more types that don't fit into a histogram, but this will remove the majority of them. I think a more thorough solution for filtering aggregation properties would be better suited in a later release(possibly 2.2).

jharwig:
Don't block all strings, zip codes are strings can be histogramed.

jharwig:
@mwizeman  Can zip codes (which are strings) be histogramed? with the zip code boundary plugin

dsingley:
zipcode boundaries are only used for choropleths,
https://en.wikipedia.org/wiki/Choropleth_map (not be confused
with geo-coordinate clusters)

joeferner:
Based on https://www.elastic.co/guide/en/elasticsearch/reference/2.2/search-aggregations-bucket-histogram-aggregation.html and https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html histograms can only be applied to numeric or dates. Which makes sense. I think we handle zipcodes using a count aggregation.

mwizeman:
@joeferner is correct on my previous experience with zip codes. We've always just used the count aggregation with those.

jharwig:
@joeferner 
re: zip codes,
yeah counting is only what we do, thanks.

re: histogram versus couting,
I think you are right, we could combine the two into some better category. Not sure how best to implement that, whether it's client or server. Currently when count is selected and the buckets are too many (~250 maybe) we just punt. Is a count aggregation just a histogram with interval = 0?

joeferner:
@jharwig I don't think so, they are two different ES calls. We could probably handle that in the back-end it's up to you.

The only problem I could I see with combining is what if you want to run counts on numbers. I guess setting the interval to 0 would work.

PR #605, opened 2016-08-25
Refactor: Move logic from UserPropertyAuthorizationRepository to a base class

joeferner:
- [x] @joeferner
- [x] @srfarley @kunklejr

This refactor will ease the implementation of non-updateable auth repos


srfarley:
How about a couple of unit tests?

joeferner:
@srfarley We already have a unit test for `UserPropertyAuthorizationRepositoryTest` which still passes. The primary change here is to help other project proprietary authorization repositories by providing a base class which does not allow updating through a public API.

srfarley:
@joeferner :+1:

sfeng88:
@joeferner won't this change break people who are running on a released version?

PR #610, opened 2016-08-26
Fix: Vertexium user property map not being updated through Proxy User (2.1 branch)

sfeng88:
- [x] @joeferner
- [ ] @srfarley @kunklejr

You will run into this issue when you have a web plugin that sets properties for new users on login.

CHANGELOG
Fixed: Vertexium user property map not being updated through Proxy User

### DIFF
--- a/core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/user/VertexiumUserRepository.java
+++ b/core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/user/VertexiumUserRepository.java
@@ -22,6 +22,7 @@ import org.visallo.core.model.user.*;
 import org.visallo.core.model.workQueue.WorkQueueRepository;
 import org.visallo.core.security.VisalloVisibility;
 import org.visallo.core.trace.Traced;
+import org.visallo.core.user.ProxyUser;
 import org.visallo.core.user.SystemUser;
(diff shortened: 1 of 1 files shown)


core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/user/VertexiumUserRepository.java
```diff
@@ -385,6 +386,12 @@ public void setPropertyOnUser(User user, String propertyName, Object value) {
         if (!value.equals(user.getCustomProperties().get(propertyName))) {
             Vertex userVertex = findByIdUserVertex(user.getUserId());
             userVertex.setProperty(propertyName, value, VISIBILITY.getVisibility(), authorizations);
+            if (user instanceof ProxyUser && ((ProxyUser)user).getProxiedUser() instanceof VertexiumUser){
+                User proxiedUser = ((ProxyUser) user).getProxiedUser();
+                if (proxiedUser instanceof VertexiumUser) {
```
joeferner: You don't need this check you already checked it on line 389

sfeng88: good catch. fixed

PR #615, opened 2016-08-28
Added: WorkerBase: add support to not terminate if nextTuple fails

joeferner:
- [x] @joeferner
- [x] @srfarley @kunklejr

In some scenarios such as running the workers from your web servers
you may not want a worker to stop completely on exception while
retrieving the next tuple from the spout. This commit adds support to
continue on failure.

CHANGELOG
Added: WorkerBase: add support to not terminate if nextTuple fails

### DIFF
--- a/config/visallo.properties
+++ b/config/visallo.properties
@@ -90,3 +90,7 @@ security.visibilityTranslator=org.visallo.core.security.DirectVisibilityTranslat
 #systemProperty.javax.net.ssl.keyStorePassword=password
 #systemProperty.javax.net.ssl.trustStore=path_to_truststore_file
 #systemProperty.javax.net.ssl.trustStorePassword=password
+
+# If set to false allows runners to continue if fetching the next tuple fails
+# org.visallo.core.ingest.graphProperty.GraphPropertyRunner.exitOnNextTupleFailure=true
+# org.visallo.core.model.longRunningProcess.LongRunningProcessRunner.exitOnNextTupleFailure=true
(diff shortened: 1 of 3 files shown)


config/visallo.properties
```diff
@@ -90,3 +90,7 @@ security.visibilityTranslator=org.visallo.core.security.DirectVisibilityTranslat
 #systemProperty.javax.net.ssl.keyStorePassword=password
 #systemProperty.javax.net.ssl.trustStore=path_to_truststore_file
 #systemProperty.javax.net.ssl.trustStorePassword=password
+
+# If set to false allows runners to continue if fetching the next tuple fails
+# org.visallo.core.ingest.graphProperty.GraphPropertyRunner.exitOnNextTupleFailure=true
+# org.visallo.core.model.longRunningProcess.LongRunningProcessRunner.exitOnNextTupleFailure=true
```
joeferner: Wasn't sure where else this should be documented.

core/core/src/main/java/org/visallo/core/model/WorkerBase.java
```diff
@@ -58,6 +66,15 @@ public void run() throws Exception {
         }
     }
 
+    protected void handleNextTupleException(VisalloLogger logger, Exception ex) throws InterruptedException {
+        if (exitOnNextTupleFailure) {
+            throw new VisalloException("Failed to get next tuple", ex);
+        } else {
+            logger.error("Failed to get next tuple", ex);
+            Thread.sleep(10 * 1000);
```
srfarley: What's the purpose of sleeping and why 10 seconds?

joeferner: In my case, I was getting a network issue and if I didn't sleep it would go into a fast spin loop.

config/visallo.properties
sfeng88: can you add this to the dev docs?

joeferner: done... https://github.com/v5analytics/visallo-lts/commit/3176376cfcd21cbf78ff5b1b4fbe3114480c935b

PR #617, opened 2016-08-30
Search filters accidentally overwriting the ontology property title

mwizeman:
2.1 pull request: https://github.com/v5analytics/visallo/pull/618

CHANGELOG
Fixed: Search filters had a bug due to using a single equals rather than comparing with triple equals.


jharwig:
At least the web worker gives some protection here to not overwrite the property title for all of visallo. I guess it's just search that will show weird display values?

PR #620, opened 2016-08-30
Fix search property filters in IE(2.1)

joeybrk372:
- [x] @joeferner
- [ ] @srfarley @kunklejr

Change search property filters to use `_.find` instead of `Array.prototype.find` which is not supported in IE

CHANGELOG
Fixed: Issue where search property filters did not show up in IE.


sfeng88:
@joeybrk372 can you add the changelog section to the PR please?

joeybrk372:
added changelog

jharwig:
Looks good. Also good example of changelog. @dsingley 

"Issue where search property filters did not show up in IE." versus the git commit: "Change search property filters to use _.find instead of Array.prototype.find"

PR #622, opened 2016-08-30
Upgrade React

joeybrk372:
- [x] @joeferner
- [x] @srfarley @kunklejr

CHANGELOG
Changed: Upgraded React from 0.14.8 to 15.3.0

### DIFF
--- a/web/war/src/main/webapp/npm-shrinkwrap.json
+++ b/web/war/src/main/webapp/npm-shrinkwrap.json
@@ -1,6 +1,5 @@
 {
   "name": "visallo",
-  "version": "0.1.0",
   "dependencies": {
     "@visallo/bootstrap": {
       "version": "2.3.2",
@@ -60,9 +59,9 @@
       "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
       "dependencies": {
         "acorn": {
-          "version": "3.2.0",
-          "from": "acorn@3.2.0",
-          "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.2.0.tgz"
+          "version": "3.3.0",
+          "from": "acorn@>=3.0.4 <4.0.0",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.3.0.tgz"
         }
       }
     },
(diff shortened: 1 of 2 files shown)


web/war/src/main/webapp/js/util/component/ConnectVisallo.jsx
```diff
@@ -0,0 +1,57 @@
+define([
```
jharwig: Not sure we should put this in `util/component` is this for plugins or bundled components? If it's for plugins written by others it should itself be apart of the public api

jharwig:
These probably shouldn't be bundled together

web/war/src/main/webapp/js/util/component/ConnectVisallo.jsx
```diff
@@ -0,0 +1,57 @@
...
+        },
+
+        getInitialState() {
+            return {
+                connected: false,
+                api: null
+            };
+        },
+
+        displayName: 'ConnectVisallo',
```
EvanOxfeld: Thinking more about your concern of sprinkling `ConnectVisallo` components throughout the component tree, I like how the redux library implements `displayName` for its higher order component. If the wrapped component has a display name `Foo`, the higher order component's display name is `Connect(Foo)`. If the wrapped component is anonymous the higher order component's display name is `Connect(Component)`.

See https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L56 and https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L18-L20

EvanOxfeld:
Agree that it would be better to split into two PRs. That said, both changes LGTM.

web/war/src/main/webapp/js/util/component/ConnectVisallo.jsx
```diff
@@ -0,0 +1,57 @@
+define([
+    'react'
+], function(React) {
+    'use strict';
+
+    const API_VERSIONS = ['v1'];
+    const loadApiPromise = loadApiVersions();
+
+    const ConnectVisallo = (Component, mapApi) => React.createClass({
```
EvanOxfeld: From an API design perspective, I recommend a couple small tweaks to the API -- rather than `ConnectVisallo` taking both `mapApi` and the component as arguments, I'd have a function that takes `mapApi` that returns a function that takes the component. This gives us room in the future to add additional optional parameters without breaking existing callers. Use a lowercase first letter for the function name as PascalCase generally denotes a component.

More concretely, instead of connecting via `ConnectVisallo(myComponent, mapApiToProps)`, connect via something like `connectVisallo(mapApiToProps)(myComponent)`

joeybrk372:
Took out HOC, will submit separate pr. Updated this pr changelog

jharwig:
@joeybrk372 Please merge master, then merge

PR #630, opened 2016-09-02
Fixed: User search when display name and user name are different

joeferner:
- [x] @joeferner
- [x] @srfarley @kunklejr

If a user's display name is different from the user name it can
be nice to search for them using either.

This commit also fixes the user admin panel to retrieve the user
by the user name which is expected by the get user route instead
of display name which can not be found.

CHANGELOG
Changed: UserRepository#find(query) now searches by both user name and display name
Fixed: User search when display name and user name are different


web/plugins/admin-user-tools/src/main/resources/org/visallo/web/adminUserTools/UserTypeaheadInput.jsx
```diff
@@ -13,7 +13,10 @@ define([
             onSelected: React.PropTypes.func.isRequired,
 
             // The current value of the typeahead input
-            username: React.PropTypes.string
+            user: React.PropTypes.oneOfType([
+                React.PropTypes.object,
+                React.PropTypes.string
```
EvanOxfeld: Can we tweak the design so that `user` only has one possible type? It would make the component much easier to reason about.

joeferner: How does it look now?

PR #631, opened 2016-09-02
Fixed: Dashboard pie chart infinite loop

joeferner:
- [x] @joeferner
- [x] @srfarley @kunklejr

While resizing the browser windows on Firefox if the size of the
browser is just right an infinite loop occurs because the length
of the string cannot be less than 0 yet the width is greater
than the available legend width.

CHANGELOG
Fixed: Dashboard pie chart infinite loop


EvanOxfeld:
Would you please also submit this change against 2.2?

joeferner:
@EvanOxfeld done: https://github.com/v5analytics/visallo/pull/642

PR #637, opened 2016-09-03
Openlayers 3 upgrade

jharwig:
- [x] @joeferner
- [x] @srfarley @kunklejr

* Remove google support as it's integration is not production ready
* Add more configure options for setting up different sources

web.ui.map.provider=[Provider]

Provider can be: osm, or anything in [ol.source](http://openlayers.org/en/master/apidoc/ol.source.html) package. Configuration to source can be provided with config properties of the form:

web.ui.map.provider.[Provider].optionName=optionValue

CHANGELOG
Changed: Upgrade OpenLayers to 3.18.2. Much more configuration is available for custom tile servers.
Removed: Google Maps integration has been removed because of limited OpenLayers support, but will automatically use OpenStreetMap instead.


web/war/src/main/webapp/js/map/map.js
```diff
@@ -198,47 +195,62 @@ define([
...
+
+                function zoomByDelta(delta) {
+                    return () => {
+                        var currentResolution = view.getResolution();
+                        if (currentResolution) {
+                            map.beforeRender(ol.animation.zoom({
+                                resolution: currentResolution,
+                                duration: ANIMATION_DURATION
+                            }));
+                            var newResolution = view.constrainResolution(currentResolution, delta);
```
EvanOxfeld: As this function already contains some ES6, I recommend using `const` here so that `newResolution` is correctly scoped to the if block.

web/war/src/main/webapp/js/map/multipointCluster.js
```diff
@@ -0,0 +1,98 @@
...
+            return;
+        }
+        this.features_.length = 0;
+        //var extent = ol.extent.createEmpty();
+        var mapDistance = this.distance_ * this.resolution_;
+        var source = this.source_;
+        var features = source.getFeatures();
+        var clustered = {};
+
+        for (var i = 0, ii = features.length; i < ii; i++) {
```
EvanOxfeld: Why did you opt for two different for loop implementations here? (line 48 uses `forEach`)

jharwig: This file is a fork of the built-in cluster source with added support for multi-point geometries.

PR #640, opened 2016-09-07
Changed: Move visibility in the details pane into the info box

joeferner:
- [x] @joeferner
- [x] @srfarley @kunklejr

This commit moves the visibility in the details pane for both
properties and comments from below the value to the popup info
box.

When using long visibility strings the details pane can become
cluttered quite fast. This addresses this by moving what most
users will not not typically use to get the job down into the
info box which is just one click away.

CHANGELOG
Changed: Move visibility in the details pane into the info box

### DIFF
--- a/core/core/src/main/resources/MessageBundle.properties
+++ b/core/core/src/main/resources/MessageBundle.properties
@@ -603,6 +603,7 @@ properties.metadata.label.modified_date=Modified
 properties.metadata.label.modified_by=By
 properties.metadata.label.status=Status
 properties.metadata.label.confidence=Confidence
+properties.metadata.label.visibility=Visibility
 properties.button.add_property=Add Property
 properties.button.show_more=Show {0} More ({1})
 properties.button.hide_more=Hide {0} ({1})
--- a/web/war/src/main/webapp/js/comments/comments.js
+++ b/web/war/src/main/webapp/js/comments/comments.js
@@ -155,7 +155,7 @@ define([
             }
         };
 
-        this.renderCommentLevel = function(maxDepth, level, selection) {
+        this.renderCommentLevel = function(maxDepth, level, selection, config) {
             var self = this;
 
             if (level > maxDepth) {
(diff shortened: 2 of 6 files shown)


jharwig:
The visibility is to the right when there's enough space. For customers with shorter strings that would be preferred. Would rather have this as a config option if your project needs it

dsingley:
I agree that it should be configurable

jharwig:
Since visibility display is a extension point you could also show a short string and a hover div? Entirely in a plugin

joeferner:
I just don't see how seeing visibility helps the analyst/investigator do their job, it only adds visual clutter. Modified date in my opinion is more useful so should it be visible as well?

jharwig:
I think the cell level visibility is a core tenant of the product, it should be displayed by default. 

But, I could see use cases where customers wouldn't want it at all or a different piece of data there. So we should make it configurable to remove it and maybe even allow some other metadata there.

kunklejr:
While I agree with @jharwig that cell visibility is a core tenant of the product, I do not think it should be shown on every field. It's a ton of noise on a display with lots of properties with long visibility strings. @jharwig and @dsingley, I think you'd change your opinion if you saw what Joe sees on his project. It's pretty ugly. I also agree with @joeferner that it doesn't really help the analyst. In fact, I'd say it probably hinders the analyst most of the time by distracting them from the data they really care about.

I suggest we move it under the "i" button with no configuration. Configuration implies multiple implementations, which I’d like to avoid as much as possible. It's just another permutation that will rarely, if ever, be tested.

jharwig:
GW used the visibility and they were short. And that's the issue, we have to make things configurable. The maxim that you don't need options doesn't work for Visallo.

jharwig:
the UI is already extremely customizable. But it has areas that aren't. Those should be made more configurable, not less.

dsingley:
It may be ugly but I believe that it is a legal requirement. If the DO
wants to ignore it or get a waiver then Joe can hide it.

jharwig:
This needs the new **visibility shown in properties list** configuration variable

joeferner:
@jharwig done can you re-review

PR #650, opened 2016-09-09
Changed: Add context menu to select connected entities

joeferner:
- [x] @joeferner
- [x] @srfarley @kunklejr

For a user, discovering that you can select connected entites
is not obvious and requires remembering keyboard shortcuts,
this commit adds the feature to the right click context menu.

CHANGELOG
Changed: Add context menu to select connected entities


web/war/src/main/webapp/js/data/withObjectSelection.js
```diff
@@ -128,7 +128,19 @@ define([], function() {
 
             this.displayInfo('graph.select.connected.starting');
 
-            Promise.all([getSelectedVertices(), getWorkspaceEdges()])
```
joeybrk372: this is more readable, can we update `getSelectedVertices` instead of putting the vertex-getting promises on top of the chain?

joeferner: done

PR #654, opened 2016-09-15
Added: Support for markdown syntax in properties with display type "link"

joeferner:
- [x] @joeferner
- [ ] @srfarley @kunklejr

Currently links only display their href as the value. This commit
allows you to use markdown syntax `[title](href)` to display an
alternate title.

CHANGELOG
Added: Support for markdown syntax in properties with display type "link"

### DIFF
--- a/web/war/src/main/webapp/js/util/vertex/formatters.js
+++ b/web/war/src/main/webapp/js/util/vertex/formatters.js
@@ -168,7 +168,16 @@ define([
                 link: function(el, property, vertex) {
                     var anchor = document.createElement('a'),
                         value = V.prop(vertex, property.name),
-                        href = $.trim(value);
+                        href = $.trim(value),
+                        text;
+
+                    if (href) {
+                        var m = href.match(/^\[(.*?)\]\((.*)\)$/);
+                        if (m) {
+                            text = m[1];
+                            href = m[2];
+                        }
+                    }
 
                     if (!(/^http/).test(href)) {
                         href = 'http://' + href;
@@ -176,7 +185,7 @@ define([
 
                     anchor.setAttribute('href', href);
                     anchor.setAttribute('target', '_blank');
-                    anchor.textContent = href;
+                    anchor.textContent = text || href;
 
                     el.appendChild(anchor);
                 },


joeferner:
@EvanOxfeld I'll let @jharwig comment on that. This was the fastest way to getting what I needed done

jharwig:
I don't think markdown for display formulas really provides that much value. It's not really a tempting language.

That said I think this change is okay, but I'd prefer a new enhancement that adds a custom field that asks for display/href as 2 fields, and stores the display in a metadata description or something.

EvanOxfeld:
Why make this change at all if we have no plans for markdown display formulas? Even if we leave this feature undocumented, it's one more thing to support.

jharwig:
@EvanOxfeld Yeah, thinking about it I would rather @joeferner just override `F.vertex.properties.link` formatter in a plugin with this for his use case.

dsingley:
what about using a compound property composed of a "title" and "href" with a display formula?

joeferner:
@dsingley display formulas cannot produce HTML as far as I know

@jharwig is `F.vertex.properties.link` part of the public API and can I guarantee the API will not change?

jharwig:
Yes it is part of the public API, and it's meant to be overridden as part of the docs.

It's a todo item to use an extension point, but this will always work

joeferner:
@jharwig How would I go about overriding this and fall back to the default behavior if I don't have markdown?

I'm also trying to think of the original use case of display type link what was it originally built for. Does the user ever want to see the original URL?

jharwig:
Not sure about the original use case. Think it was the project with @sfeng88 and @dsingley that wanted it.

```
F.vertex.properties.link = _.wrap(F.vertex.properties.link, function() {
    var args = _.toArray(arguments);
    var originalLink = args.shift();
    if (markdown) {
        ...
    } else {
        originalLink.apply(null, args);
    }
});
```

joeferner:
@jharwig thanks, that's what I was thinking. I'm guessing they would want this title as well for their project. Would users want a title more often than not. If we built a UI around entering links in this format would that be helpful for other?

jharwig:
Yes, I think a real 2 input UI would be best. And if the value is markdown that works. I think it would be nicer if title went in a metadata field, but the field components aren't set up to set metadata.

joeferner:
I'm going to close this PR and rework to add it to metadata instead of markdown, I like that much better. Then I'll take a crack at creating a UI

PR #655, opened 2016-09-16
Allow all component attacher methods to chain(2.2)

joeybrk372:
- [x] @joeferner
- [ ] @kunklejr

`attacher.prototype.teardown` and `attacher.prototype.verifyState` didn't return themselves so couldn't chain. 

For example: https://github.com/v5analytics/visallo/blob/master/web/war/src/main/webapp/js/dashboard/dashboard.js#L221


web/war/src/main/webapp/js/util/component/attacher.js
```diff
@@ -42,11 +42,13 @@ define([
                 throw new Error('Params must be an object')
             }
         }
+        return this;
```
jharwig: `verifyState` isn't a public function so I wouldn't add chaining here. Probably should be underscored

PR #664, opened 2016-09-22
Fix map radius search 

rygim:
- [x] @joeferner
- [ ] @kunklejr

CHANGELOG
Fixed: Corrected the calculation for detecting the clicked center of the search radius on map

The way that the center radius was being calculated for both chrome + ff
used object properties that either were there before and were removed
with new open street maps or were possibly how google calculated it.
Now, the shape of the object is detected and the appropriate calculation
for detecting the center of the map search is used.

Fixes issue https://github.com/v5analytics/visallo-lts/issues/1724


web/war/src/main/webapp/js/map/map.js
```diff
@@ -667,8 +667,19 @@ define([
...
-                    var offset = self.$node.offset(),
-                        centerCoordinate = map.getCoordinateFromPixel([evt.pageX - offset.left, evt.pageY - offset.top]),
+                    var pixelX, pixelY;
+
+                    if (evt.pageX) {
+                        var offset = self.$node.offset();
+                        pixelX = evt.pageX - offset.left;
+                        pixelY = evt.pageY - offset.top;
+                    }
+                    else if (evt.pixel){
```
jharwig: This handler is a regular jquery event, so pageX/pageY should be used. What is `pixel`?

jharwig: My mistake it is from open layers. But it should probably be

`var pixel = map.getEventPixel(evt)`

I would check the pixel result and make sure you don't also need to adjust by `offset`

jharwig: it looks like it is relative to the map, but just verify that the x isn't +~30 (menubar width)

rygim: changed as per our discussion and verified

joeferner:
I ran across this issue today and fixed it slightly different, before I realized you had a PR.

`evt` has `evt.coordinate` which replaces the need for `pageX`, `pageY`, etc. Much simpler.

jharwig:
@rygim Please update with just evt.coordinate like joe suggested. Don't need any offsets, or pageX references.

rygim:
changed

joeferner:
LGTM... squash the commits before merge

PR #665, opened 2016-09-25
GPW/LRP test base class changes

joeferner:
- [x] @joeferner
- [x] @kunklejr

see commit messages for more descriptions

CHANGELOG
Added: LongRunningProcessWorkerTestBase
Changed: Moved GraphPropertyWorkerTestBase and GraphPropertyWorkerTestSetupBase to core-test
Added: Call GPW process with null property to simulate actual behavior


core/core-test/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyWorkerTestSetupBase.java
```diff
@@ -30,6 +29,7 @@
  * TODO: It might make sense to combine this with {@link GraphPropertyWorkerTestBase}.
  * TODO: There are a number of GPW unit tests with copied/pasted setup that should extend this base class.
  */
+@Deprecated
```
mwizeman: Since this is deprecated, maybe add a note in the javadoc as to what someone using this should do instead?

joeferner: done

PR #666, opened 2016-09-25
Changed: Moved threading logic for GPW/LRP from initializers to runners

joeferner:
- [x] @joeferner
- [ ] @kunklejr

By moving the logic of starting threads from the initializer to the runner
we are able to make standalone applications which start GPW and LRP processing
threads.

CHANGELOG
Changed: Moved threading logic for GPW/LRP from initializers to runners. Allows for making standalone application which start GPW and LRP processing threads.

### DIFF
--- a/core/core/src/main/java/org/visallo/core/config/Configuration.java
+++ b/core/core/src/main/java/org/visallo/core/config/Configuration.java
@@ -178,12 +178,12 @@ public class Configuration {
         return subset;
     }
 
-    public void setConfigurables(Object o, String keyPrefix) {
+    public <T> T setConfigurables(T o, String keyPrefix) {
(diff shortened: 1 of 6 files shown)


sfeng88:
@joeferner in the changelog entry, can we add a little bit more detail please? I would be ok with using the commit message.

joeferner:
@sfeng88 How is that?

mwizeman:
Should ExternalResourceRunner be refactored to work like this too? Conceptually it seems like a pretty similar task to these refactored versions.

joeferner:
@mwizeman `ExternalResourceRunner` was already in core and didn't have dependencies on web. It also works a little differently.

PR #680, opened 2016-10-07
Support aliases in resource bundles

joeybrk372:
- [x] @joeferner
- [ ] @kunklejr @mwizeman @sfeng88
- [x] @dsingley @EvanOxfeld 

Example Usage:

```
//messsages.properties
alias.case=Case
com.visallo.workspace.id=@{alias.case} ID
```
```js
//results in 'Case ID'
i18n('com.visallo.workspace')
```

CHANGELOG
Added: Support for aliasing in resource bundles.

### DIFF
--- a/core/core/src/main/java/org/visallo/core/config/VisalloResourceBundle.java
+++ b/core/core/src/main/java/org/visallo/core/config/VisalloResourceBundle.java
@@ -1,19 +1,33 @@
 package org.visallo.core.config;
 
+import org.visallo.core.util.VisalloLogger;
+import org.visallo.core.util.VisalloLoggerFactory;
+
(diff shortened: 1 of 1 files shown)


kunklejr:
How is this different than simply changing the `com.visallo.workspace.id` property name to say Case?

joeybrk372:
The example was only one replacement but it does all instances.

```
//messsages.properties
alias.case=Case
com.visallo.workspace.id=@{alias.case} ID
com.visallo.delete.workspace=Delete @{alias.case}
com.visallo.remove.workspace=Remove from @{alias.case}
//Case ID, Delete Case, Remove from Case```

PR #698, opened 2016-10-25
Fixed: UI date parsing of IE formatted strings

joeferner:
- [x] @joeferner
- [x] @dsingley @EvanOxfeld 

In some cases IE formats strings in 2016-10-24T22:12:13.012+0000 format
this commit fixes the parser to handle this format.

### DIFF
--- a/web/war/src/main/webapp/js/util/formatters.js
+++ b/web/war/src/main/webapp/js/util/formatters.js
@@ -482,8 +482,13 @@ define([
                             parsed += (isNegative ? '-' : '+') + hours + ':' + minutes;
                             dateInLocale = new Date(parsed);
                         } else {
-                            console.warn('Unable to parse date', numberOrString);
-                            return '';
+                            var m = moment(numberOrString, 'YYYY-MM-DDTHH:mm:ss.SSSZ', true);
+                            if (m.isValid()) {
+                                dateInLocale = m.local().toDate();
+                            } else {
+                                console.warn('Unable to parse date', numberOrString);
+                                return '';
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


web/war/src/main/webapp/test/unit/spec/util/formattersTest.js
```diff
@@ -226,9 +226,11 @@ define([
                     offset = date.getTimezoneOffset(),
                     dateLocal = new Date(date.getTime() + offset * 60 * 1000);
 
-                f.date.local('2015-06-23 13:16 CDT').getTime().should.equal(dateLocal.getTime())
+                f.date.local('2015-06-23T17:16:00.000+0000').getTime().should.equal(dateLocal.getTime());
```
joeferner: Not 100% sure on these tests. They pass but they don't feel right to me

web/war/src/main/webapp/js/util/formatters.js
```diff
@@ -482,8 +482,13 @@ define([
                             parsed += (isNegative ? '-' : '+') + hours + ':' + minutes;
                             dateInLocale = new Date(parsed);
                         } else {
-                            console.warn('Unable to parse date', numberOrString);
-                            return '';
+                            var m = moment(numberOrString, 'YYYY-MM-DDTHH:mm:ss.SSSZ', true);
```
joeferner: I had to use moment's strict parsing here because of unit tests. I'm wondering if we might just replace this and the above code with moment's loose format handling.

EvanOxfeld:
On second thought, the travis failure may be a legit timezone issue.

jharwig:
Working on this now...

PR #699, opened 2016-10-27
add listener to save all dashboard cards on change

rygim:
- [x] @joeferner
- [ ] @dsingley @EvanOxfeld 

CHANGELOG
Fixed: Added listeners to update the card locations on the server since there were certain instances in which the card locations were not updated and saved to the server.

There were certain cases in which the dashboard items configuration
metrics were not being changed correctly in memory. This change will now
save the changes to the server whenever any of those events are fired

Fixes https://github.com/v5analytics/visallo-lts/issues/1805


joeybrk372:
does this happen when there isn't a table card? seems like we would catch all cases already

web/war/src/main/webapp/js/dashboard/dashboard.js
```diff
@@ -450,6 +450,8 @@ define([
                         });
                     })
             }
+
```
joeferner: Remove extra whitespace

web/war/src/main/webapp/js/dashboard/dashboard.js
```diff
@@ -493,9 +495,28 @@ define([
...
                     $(node).attr('data-item-id', item.id);
+                    self.dashboard.items.push(item);
+                    self.updateAllDashboardItems();
                 });
+        };
 
-            this.dashboard.items.push(item);
+        this.updateAllDashboardItems = function(){
+            var self = this,
+                stackItemMap = [],
```
joeferner: Did you mean for this to be an array?

web/war/src/main/webapp/js/dashboard/dashboard.js
```diff
@@ -544,7 +565,9 @@ define([
                     }
                 },
                 { concurrency: 1 }
-            ).done(function() {
+            )
+                .then(this.updateAllDashboardItems.bind(this))
+                .done(function() {
```
joeferner: This could just be `.done(this.createDashboardItemToGridStackInBatch.bind(this))`

rygim:
@joeybrk372 We were missing some cases where things moved in the gridview on their own. In @nidasid's example movie that she attached to the issue, the welcome card was never getting it's correct configuration set into the state inside of the component which meant that the location was also never getting correctly persisted to the server. That meant that the gridview got to decide where components went when it got rendered and it wasn't consistent. This is a more brute-force solution to save everything all the time when there is a change in the location of the dashboard cards (it will incur more web traffic) so there might be another solution or event that is being called, but this at least fixes the issue.

web/war/src/main/webapp/js/dashboard/dashboard.js
rygim: fixed

web/war/src/main/webapp/js/dashboard/dashboard.js
rygim: I changed it to an object since it makes more sense that way. It worked both ways, but you are right, it looked a little strange

web/war/src/main/webapp/js/dashboard/dashboard.js
rygim: changed

rygim:
Merging since it has been a while and most people have checked it

PR #700, opened 2016-10-28
Give default authorizations to all users, not just users without auths

kunklejr:
Currently, the default authorizations only get returned if the user doesn't have any existing authorizations. The default authorizations should apply to all users, regardless of other authorizations they might have.
- [x] @joeferner 
- [ ] @dsingley @EvanOxfeld 

CHANGELOG
Changed: Default authorizations will be given to all users, not just users without any existing authorizations.


joeferner:
@kunklejr Can you update the unit tests, Travis does not like your change

PR #707, opened 2016-10-29
fix histogram toolbar was not showing up

sfeng88:
Release 2.2: https://github.com/v5analytics/visallo/pull/706, https://github.com/v5analytics/visallo/pull/715


web/war/src/main/webapp/js/util/acl.js
```diff
@@ -22,18 +22,24 @@ define([
...
-                        name: elementPropertyAcl.name,
-                        key: elementPropertyAcl.keys || null
-                    });
-                    if (matches.length === 0) {
-                        propertyAcls.push(elementPropertyAcl);
-                    } else {
-                        _.each(matches, function(r) {
-                            _.extend(r, elementPropertyAcl);
+                var elements = element;
+                if (!(element instanceof Array)) {
```
joeybrk372: This is checking the old to modify the new. Can be simplified to `const elements = element instanceof Array ? element : [element];`

web/war/src/main/webapp/js/util/acl.js
```diff
@@ -22,18 +22,24 @@ define([
...
+                    elements = [element];
+                }
+                _.each(elements, function (e) {
```
joeybrk372: prefer built-ins `elements.forEach((e) => {  `

web/war/src/main/webapp/js/util/acl.js
jharwig: Also use `Array.isArray`

sfeng88:
@jharwig @joeybrk372 fixed on both PRs.

rygim:
Can we merge this into master?

PR #710, opened 2016-10-31
Return formatted display types(2.2)

joeybrk372:
- [x] @joeferner
- [ ] @dsingley @EvanOxfeld 

new pr due to oversight on `FormulaEvaluator`

CHANGELOG
Added: Data/display type formatters now return the element.


jharwig:
Why is this needed, that calling function passes the element. Also any plugins that defined their own formatters here would break?

joeybrk372:
You may not want to render that element immediately. This will allow you to do either. However, I can add some documentation to make this more clear for plugin writers.

jharwig:
I just don't see it. They way forward is a a real extension point for these things that have different semantics than before (that support all cases you need.) Then the old ways should still work without any changes, but will show a warning deprecation. We can't change the way these work as they have been documented as overridable.

PR #714, opened 2016-11-01
Fix dashboard freeze by leftover draggable-wrapper

rygim:
- [x] @joeferner
- [ ] @dsingley @EvanOxfeld 

CHANGELOG
Fixed: Dashboard does not freeze anymore when cards are moved vertically

When moving the dashboard cards in the way that was shown in issue
1807, the draggable-wrapper that was added is never cleared away. By
registering the method to clear on the deactivate event as well as the drop
event, the wrapper is cleared and the dashboard can be arranged as
normal with no freezing.

Fixes https://github.com/v5analytics/visallo-lts/issues/1807

### DIFF
--- a/web/war/src/main/webapp/js/data/withWorkspaceVertexDrop.js
+++ b/web/war/src/main/webapp/js/data/withWorkspaceVertexDrop.js
@@ -58,7 +58,7 @@ define([], function() {
                                 }
                             })
                     };
-                    
+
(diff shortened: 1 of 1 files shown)


rygim:
OBE from jasons changes

rygim:
Closed in lieu of https://github.com/v5analytics/visallo/pull/736 and being OBE on master

PR #717, opened 2016-11-05
Provide options for restricting directory search to people and/or groups (2.2)

kunklejr:
- [x] @joeferner
- [x] @dsingley @EvanOxfeld 

CHANGELOG
Changed: Provide options for restricting directory search to people and/or groups


web/war/src/main/webapp/js/data/web-worker/services/directory.js
```diff
@@ -14,10 +14,18 @@ define([
...
+            options = options || {};
+            var params = {
                 search: search
-            });
+            };
+            if (typeof options.people !== undefined) {
+                params.people = options.people;
+            }
+            if (typeof options.groups !== undefined) {
+                params.groups = options.groups;
```
EvanOxfeld: `typeof` returns a string. While you could remove the `typeof` to make it work, instead I'd recommend replacing the ifs with `_.extend` to copy any properties from the `options` object over to `params`:

```
search: function(search, options) {
  var params = _.extend({}, options, { search: search });
  return ajax('GET', '/directory/search', params);
}
```

EvanOxfeld: @jharwig Is ES6 syntax supported here? We could simplify the above even further.

kunklejr: Two lines seems pretty simple to me. That's a nice change.

kunklejr:
@EvanOxfeld: Your requested change was pushed.

web/war/src/main/webapp/js/data/web-worker/services/directory.js
```diff
@@ -14,10 +14,9 @@ define([
             });
         },
 
-        search: function(search) {
-            return ajax('GET', '/directory/search', {
-                search: search
-            });
+        search: function(search, options) {
+            var params = _.extend({}, options, { search: search });
```
jharwig: can now do `var params = { ...options, search };`

web/war/src/main/webapp/js/data/web-worker/services/directory.js
jharwig: Yes, see comment above

PR #721, opened 2016-11-08
Multiple Work products (Graph, Map) per Workspace

jharwig:
- [x] @joeferner
- [x] @dsingley @EvanOxfeld 

[Video Walkthrough (3:28)](https://dl.dropboxusercontent.com/u/7953598/tempShare/workproducts-1.mp4)

screen shot 2016-11-08 at 5 46 25 pm

CHANGELOG
Added: Graph and Map are now "Case Work Products" within a workspace. Multiple of each type can be created within one workspace.
Removed: Workspace search is removed since workspaces don't have entities
Removed: Add to Workspace URL scheme has been removed as entities can't be added to a workspace.


core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.List;
+
+public abstract class WorkProductElements implements WorkProduct, WorkProductHasElements {
+
+    public static final String WORKSPACE_PRODUCT_TO_ENTITY_RELATIONSHIP_IRI = "http://visallo.org/workspace/product#toEntity";
+
+
+        @Override
+        public void update(JSONObject params, Graph graph, Vertex workspaceVertex, VertexBuilder vertexBuilder, User user, Visibility visibility, Authorizations authorizations) {
```
joeferner: Can you make params a Jackson POJO instead of JSONObject

core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java
```diff
@@ -0,0 +1,93 @@
...
+            JSONArray removeVertices = params.optJSONArray("removeVertices");
+            if (removeVertices != null) {
+                JSONUtil.toList(removeVertices).stream().forEach(id -> {
+                    graph.softDeleteEdge(getEdgeId(vertexBuilder.getVertexId(), (String) id), authorizations);
+                });
+            }
+        }
+
+        @Override
+        public JSONObject get(JSONObject params, Graph graph, Vertex workspaceVertex, Vertex productVertex, User user, Authorizations authorizations) {
```
joeferner: Can you make the results and params a Jackson POJO instead of JSONObject.

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
```diff
@@ -841,6 +830,34 @@ private Dashboard dashboardVertexToDashboard(
...
+        Property previewDataUrlProperty = WorkspaceProperties.PRODUCT_PREVIEW_DATA_URL.getProperty(productVertex, user.getUserId());
+        InputStream previewDataUrl = null;
+        String md5 = null;
+        if (previewDataUrlProperty != null) {
+            Metadata.Entry entry = previewDataUrlProperty.getMetadata().getEntry("http://visallo.org/product#previewImageMD5");
+            if (entry != null) {
+                md5 = (String) entry.getValue();
+            }
+            StreamingPropertyValue previewValue = (StreamingPropertyValue) previewDataUrlProperty.getValue();
+            previewDataUrl = previewValue.getInputStream();
```
joeferner: This InputStream needs to be closed at some point

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
```diff
@@ -887,12 +904,251 @@ public String addOrUpdateDashboard(String workspaceId, String dashboardId, Strin
...
+            try {
+                md = MessageDigest.getInstance("MD5");
+                byte[] array = md.digest(imageData);
+                StringBuffer sb = new StringBuffer();
+                for (int i = 0; i < array.length; i++) {
+                    sb.append( String.format( "%02x", array[i]));
+                }
+                previewImageMD5 = sb.toString();
+            } catch (NoSuchAlgorithmException e) {
+                e.printStackTrace();
```
joeferner: I know this should never happen but can you LOGGER.error it

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
```diff
@@ -887,12 +904,251 @@ public String addOrUpdateDashboard(String workspaceId, String dashboardId, Strin
...
+
+        Vertex workspaceVertex = getVertex(workspaceId, user);
+        Authorizations authorizations = getAuthorizationRepository().getGraphAuthorizations(
+                user,
+                VISIBILITY_STRING,
+                workspaceId
+        );
+        Visibility visibility = VISIBILITY.getVisibility();
+        Graph graph = getGraph();
+        VertexBuilder productVertexBuilder = graph.prepareVertex(productId, visibility);
```
joeferner: This is going to generate more and more history every time it is called. You should look up the previous product and update it. You can use GraphRepository#beginGraphUpdate then call getOrCreateVertexAndUpdate (https://github.com/v5analytics/visallo/blob/master/core/core/src/main/java/org/visallo/core/model/graph/GraphUpdateContext.java#L196).

See example... https://github.com/v5analytics/visallo/blob/master/core/core/src/main/java/org/visallo/core/model/graph/GraphUpdateContext.java#L24

web/war/src/main/webapp/js/components/Attacher.jsx
```diff
@@ -0,0 +1,46 @@
...
+                this.reattach(nextProps);
+            }
+        },
+
+        componentWillUnmount() {
+            this.attacher.teardown();
+        },
+
+        render() {
+            return <div ref="node" />;
```
joeybrk372: can we have `div` as the default but leave this open to be specified? Maybe something like:

```js
getDefaultProps() {
     return {
        wrapper: 'div'
    }
},
render() {
    return <this.props.wrapper ref="node" />;
}
````

File: core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java (hunk 1 above)
Comment (jharwig) [inline core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java]: How can I if plugins could put whatever they want in here?

File: core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java (hunk 3 above)
Comment (jharwig) [inline core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java]: thanks, fixing...

File: core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java (hunk 4 above)
Comment (jharwig) [inline core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java]: ok

File: web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Cytoscape.jsx [hunk 7]
```diff
@@ -0,0 +1,812 @@
...
+        componentDidMount() {
+            this.moving = {};
+            this.updatePreview = _.debounce(this._updatePreview, PREVIEW_DEBOUNCE_SECONDS * 1000)
+            this.previousConfig = this.prepareConfig();
+            const cy = cytoscape(this.previousConfig);
+
+            cytoscape('layout', 'bettergrid', betterGrid);
+
+            this.clientRect = this.refs.cytoscape.getBoundingClientRect();
+            this.setState({ cy })
```
jharwig: There's a few edge cases where the parent component (graph) grabs this object. This is never set more than once cause the object is shared when switching graphs until teardown

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -0,0 +1,903 @@
...
+        },
+
+        onMenuExport(componentPath) {
+            var exporter = _.findWhere(
+                    this.props.registry['org.visallo.graph.export'],
+                    { componentPath }
+                );
+
+            if (exporter) {
+                const cy = this.refs.cytoscape.state.cy;
```
jharwig: Right, this is a case of a legacy component needing cy. I didn't want to put this in Cytoscape, because that component would have to know about exporting products.

core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java
jharwig: How can I if plugins could put whatever they want in here?

PR #730, opened 2016-11-14
Display offline on network loss

joeybrk372:
- [x] @joeferner
- [x] @dsingley @EvanOxfeld 

We show the offline overlay based on the websocket being closed. However, atmosphere triggers the close event by listening for the `offline` event on `window` which is not available inside a worker.

CHANGELOG
Fixed: Offline overlay not appearing when the browser loses network connection.

### DIFF
--- a/web/war/src/main/webapp/js/data/withWebsocket.js
+++ b/web/war/src/main/webapp/js/data/withWebsocket.js
@@ -27,6 +27,11 @@ define(['util/websocket'], function(websocketUtils) {
                 }
             });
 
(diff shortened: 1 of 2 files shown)


jharwig:
Does this work when the browser/server doesn't support websockets (does it downgrade to comet successfully)?

Also I think we can drop support for browsers that can't connect to web socket in web worker. That was Firefox 17.

joeybrk372:
Yes, it will still downgrade correctly. This change does not affect the atmosphere connection itself.

That is probably right, I will submit a separate pr to remove the legacy websocket.

PR #731, opened 2016-11-14
Use yarn instead of npm

jharwig:
- [ ] @joeferner
- [x] @dsingley @EvanOxfeld 

Faster, more deterministic versions of dependencies. Uses npm registry, package.json. Also retries downloads, and works offline

CHANGELOG
Changed: Webapp and plugins install dependencies with [yarn](https://yarnpkg.com) instead of npm


jharwig:
Needs testing on windows. `mvn compile` on windows should confirm

joeferner:
I'm trying to run `yarn install` in a firewalled environment and I'm having problems downloading taffydb. It appears to be trying to get it from github which my firewall blocks.

jharwig:
Did you get a yarn-error log?

My taffydb resolves to this:
```
taffydb@2.6.2:
    resolved "https://registry.yarnpkg.com/taffydb/-/taffydb-2.6.2.tgz#7cbcb64b5a141b6a2efc2c5d2c67b4e150b2a268"
```

joeferner:
I think I'm going to need to package up my yarn cache and bring it behind the firewall, I'm getting all sorts of errors even after this about not being able to find registry.yarnpkg.com

joeferner:
Do we want to change graph-product and map-product as well?

jharwig:
@joeferner graph and map products are switched to yarn in this PR. Is that what you are asking?

joeferner:
@jharwig sorry, forgot which branch I was on and didn't see yarn.lock files. Yes, this PR does indeed address this.

PR #735, opened 2016-11-15
Add map options extension point that mirrors the graph.options

jharwig:
- [x] @joeferner
- [ ] @dsingley @EvanOxfeld 

Syntax is the same as `org.visallo.graph.options` except instead of getting `cy` as a prop/attribute you get:
```
ol: OpenLayers object
map: OpenLayers map instance
cluster: Map cluster pins source, layers, instances
```

CHANGELOG
Added: Add map configuration UI with [`org.visallo.map.options`](http://docs.visallo.org/extension-points/front-end/mapOptions/) extension point

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -49,6 +49,7 @@
         * [Layout Component](extension-points/front-end/layout/component.md)
         * [Layout Type](extension-points/front-end/layout/type.md)
         * [Logout](extension-points/front-end/logout/index.md)
+        * [Map Options](extension-points/front-end/mapOptions/index.md)
         * [Menubar](extension-points/front-end/menubar/index.md)
         * [Search Filters](extension-points/front-end/searchFilters/index.md)
         * [Search Toolbar](extension-points/front-end/searchToolbar/index.md)
--- a/docs/extension-points/front-end/mapOptions/index.md
+++ b/docs/extension-points/front-end/mapOptions/index.md
@@ -1,0 +1,79 @@
+Map Options Plugin
+=================
+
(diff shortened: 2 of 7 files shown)


joeferner:
Will this allow me to add KML layers? https://github.com/v5analytics/visallo-lts/issues/1860

jharwig:
@joeferner Hmm... 

You could create a map option extension that has upload button/or drag drop support that accepts kml. Since it has access to the map and open layers api, it could add a new vector layer with xml format (http://openlayers.org/en/v3.7.0/examples/kml.html). The component is loaded on map load even if the options popover isn't shown.

Here's a quick test that works (if you enable mixed content cause it downloads from http)

```
require(['configuration/plugins/registry'], function(registry) {    
    define('myplugins/kml', ['react'], function(React) {
        return React.createClass({
            render() {                
                var map = this.props.map;
                
                // TODO: check if map.getLayers() already has kml layer

                map.addLayer(
                    new ol.layer.Vector({
                      source: new ol.source.Vector({
                        url: 'http://openlayers.org/en/v3.7.0/examples/data/kml/2012-02-10.kml',
                        format: new ol.format.KML()
                      })
                    })
                );

                // TODO
                return React.createElement('div', {}, "Drop KML Here")
            }
        });
    });
    registry.registerExtension('org.visallo.map.options', {
        identifier: 'hello-kml',
        optionComponentPath: 'myplugins/kml'
    });
});
```

jharwig:
@joeferner That said, I think some more explicit extensions for layers might be better. So as to not collide with other plugins, and ease of development.

jharwig:
this PR is waiting on updated extension point docs, working on them now.

PR #738, opened 2016-11-16
Changed: Treat zero as a special case for details histogram

joeferner:
- [x] @joeferner
- [ ] @dsingley @EvanOxfeld 

This change is based on user feedback. Thoughts?

Many times when dealing with numbers zero is a special case, this
commit addresses this by moving zero values into their own histogram
bin.

Before
visallo-hist-before

After
visallo-hist-after

Combined with PR https://github.com/v5analytics/visallo/pull/737 this can create a nicer histogram.

### DIFF
--- a/web/war/src/main/webapp/js/detail/properties/histograms.js
+++ b/web/war/src/main/webapp/js/detail/properties/histograms.js
@@ -241,17 +241,31 @@ define([
 
                                                 d.values = values;
 
+                                                var createBin = function(props) {
+                                                    return _.reject(
+                                                        d3.layout.histogram().value(_.property('value'))(props), function(bin) {
+                                                            return bin.length === 0;
+                                                        });
+                                                };
+
                                                 if (ontologyProperty &&
                                                     ~BINABLE_TYPES.indexOf(ontologyProperty.dataType) &&
                                                    !ontologyProperty.possibleValues) {
 
-                                                    var bins = _.reject(
-                                                        d3.layout.histogram().value(_.property('value'))(d[1]), function(bin) {
-                                                        return bin.length === 0;
-                                                    });
+                                                    var lessThanZeroBins = createBin(_.filter(d[1], function(prop) {
+                                                        return prop.value < 0;
+                                                    }));
+                                                    var greaterThanZeroBins = createBin(_.filter(d[1], function(prop) {
+                                                        return prop.value > 0;
+                                                    }));
+                                                    var zeroBins = createBin(_.filter(d[1], function(prop) {
+                                                        return prop.value === 0;
+                                                    }));
 
-                                                    d.bins = bins;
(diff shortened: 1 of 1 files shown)


kunklejr:
Why is zero a special case? I don't particularly like bins that cross zero because it doesn't provide an easy way to drill into **only** positive or negative values.

joeferner:
@kunklejr This was from customer feedback. I can think of a couple examples that could support this...

* In finance an account balance of $0 is probably interesting because it might denote the account is not being used. Verses having an account balance of even $0.01.
* In social situations a duration of 0 seconds might denote a connection, conversation, or call attempt was made but was unsuccessful or the other person didn't respond.
* If you are tracking inventory 0 is more interesting than 1 because you don't have any in stock.

That being said I'm still not 100% convinced that adding special handling makes sense 100% of the time but I could see scenarios where it does.

jharwig:
Seems like I wouldn't want this to be default. Ideally:

1. System-wide config (either in ontology, or in config) per property that turns on this behavior
2. User specific preference per property that allows users to toggle the behavior

For this to merge, I'd say 1 is required. We need to re-write the histogram anyway in react at some point. I'd like to discuss some other ideas for how to improve it.

joeferner:
I was thinking that initially too but I'm finding it hard to think of a time where you wouldn't want negative, zero, and positive values in separate buckets.

jharwig:
I think you're zoned in on your use case, though. Think about numbers where 0 has less meaning. Especially with decimal types:

Temperatures, Angles / Direction, Wavelengths

joeferner:
I was thinking of numbers like that but many of those types of natural numbers rarely go to zero (wavelengths, height, weight, length, distance, hear rate, blood pressure, etc) and are rarely exactly zero (temperature, angles, direction). I also think many of the numbers that could go to zero (temperature, angles, direction) are typically not stored in Visallo. For this reason I might argue the default should be to separate negative, zero, positive and make exceptions for the others.

Just my thoughts. Anyone else have opinions?

jharwig:
We did store heading in visallo, and as a histogram it wouldn't make sense to have that have a zero bucket

jharwig:
I guess it's just weird to me that you'd make all the other numbers bucket based on the results but remove all that exactly equal 0 and put those separately?

What about decimals that are < 1 ? I find it hard to accept that it should be default

jharwig:
What it seems you really need is a way for you to describe the scale, interval for all properties. But also being able to have special intervals for some ranges. Seems pretty complex to me.

kunklejr:
People at one of our financial customers would ask for the ability to customize the bucket sizes all the time. That was mainly in reference to graphs created on the dashboard though.

rygim:
@joeferner what should we do about this PR?

joeferner:
@rygim Lets talk about it at x-team

PR #744, opened 2016-11-18
Changed: Multi-select detail histogram bottom margin

joeferner:
- [x] @joeferner
- [ ] @dsingley @EvanOxfeld 

When adding a multi-select details plugin the margin of the histogram pushes the additional plugin way down. This removes this gap.

Anyone have any idea why this margin was there in the first place? @jharwig 

Before
image

After
image


kunklejr:
No idea.

PR #756, opened 2016-11-22
Gpw find right property (2.1)

rygim:
- [x] @joeferner
- [ ] @dsingley @EvanOxfeld 

CHANGELOG
Fixed: Find correct gpw property to process

Whenever the value of a published property has been changed on a workspace, those changes were not being put through the graph property workers. This change checks all of the properties that can be returned and uses the workspace ID to decide which property needs to be put through the graph property workers.

I will cherry pick this to release-3.0 and master if this is accepted.

### DIFF
--- a/core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
+++ b/core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
@@ -1,6 +1,7 @@
 package org.visallo.core.ingest.graphProperty;
 
 import com.google.common.collect.Lists;
+import com.google.common.collect.Queues;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import org.apache.commons.io.FilenameUtils;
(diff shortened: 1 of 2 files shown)


core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
```diff
@@ -263,27 +262,49 @@ private boolean doesExist(Element element) {
...
+            properties.addAll(Lists.newArrayList(element.getProperties(message.getPropertyName())));
+        } else {
+            properties.addAll(Lists.newArrayList(element.getProperties(message.getPropertyKey(), message.getPropertyName())));
+        }
+
+        if (properties.isEmpty()) {
+            return null;
+        }
+
+        List<Property> propertyList = Lists.newArrayList();
```
joeferner: You can probably tweak this bit of code to be more efficient. Maybe something like this...

```java
Property result;
for (Property property : properties) {
  if (message.getWorkspaceId() != null && property.getVisibility().hasAuthorization(message.getWorkspaceId())) {
    result = property;
  } else if (result == null) {
    result = property;
  }
}
return result;
```

This would eliminate creating a new array list and the potentially expensive insert at index of ArrayList.

With this change you can also eliminate the `properties.isEmpty` check above if you want.

rygim: Good call. I did a little refactoring as well as putting this in. See changed files for updated code

joeferner:
Looks good, would you mind running `git rebase release-2.1 -i` and squashing down some of the commits. Then go ahead and merge

rygim:
Sure. just curious, squashing from here doesnt squash them down like we want? I always figured it did

jharwig:
I think it appends all the commits like:

* commit 1 message
* commit 2 message

PR #772, opened 2016-11-29
Remove pluginDevMode configuration (3.0)

jharwig:
- [x] @joeferner
- [ ] @dsingley @EvanOxfeld 

Different settings could cause plugins that were expecting minification to fail, so removing the configuration setting. 3.0 contains a new helper function for disabling on a per-file basis with `registerCompiledJavaScript` which will automatically pull in `.map` file if available.

The original purpose of the configuration was for debugging when browsers source map support wasn't good. Now it is, and better to run more like production.

CHANGELOG
Removed: `pluginDevMode` is no longer available. Turn off compilation on a per-file basis with `registerCompiledJavaScript` or `skipCompile`


joeferner:
I have a patch to an enterprise plugin which is broken without `pluginDevMode=true`. I'll submit it tonight (11/30) and we can merge this

PR #776, opened 2016-11-30
Added: methods and routes to link two products

joeferner:
- [ ] @joeferner
- [ ] @dsingley @EvanOxfeld 

Linking two products will update the linked product when the source
product is updated

This PR is for discussion purposes right now...
* Currently these routes and methods are not exposed to the user for linking.
* If existing work products which have existing data get linked that data does not get updated to the linked product. Should it?
* If you link a map to a graph and drag an item onto the map it gets placed at 0,0 since it doesn't have coordinates. What should happen?

CHANGELOG
Added: methods and routes to link two products

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/workspace/DefaultWorkspaceListener.java
+++ b/core/core/src/main/java/org/visallo/core/model/workspace/DefaultWorkspaceListener.java
@@ -47,4 +47,14 @@ public class DefaultWorkspaceListener implements WorkspaceListener {
     public void workspaceDeleteUser(Workspace workspace, String userId, User user) {
 
     }
+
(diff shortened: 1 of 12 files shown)


rygim:
Response to your questions: 
- I dont really see a current use-case for linking but it is definitely a nice to have since it seems like you are trying to restore the pre-work product functionality of visallo.
- Probably should add all the separate data from both to the other linked ones. Also should probably notify the user that they are about to do that.
- Not sure what would be the best. When we automatically added many elements to the graph from unrolling based on signal prop, we just started laying them out below the element we clicked on. Unfortunately, you dont have that so I would defer to @jharwig

PR #797, opened 2016-12-03
Fixed: Priority not being propagated correctly

joeferner:
- [x] @joeferner
- [x] @kunklejr @sfeng88 @diegogrz
- [x] @mwizeman @EvanOxfeld 
- [x] @joeybrk372 @rygim @jharwig 

CHANGELOG
Fixed: Priority not being propagated correctly

PR #798, opened 2016-12-03
Added: methods to update work product data and extended data

joeferner:
- [x] @joeferner
- [x] @mwizeman @EvanOxfeld 

Data is always brought back when requesting work products. Extended data
is brought back when requesting a single work product passing the
includeExtended flag.

CHANGELOG
Added: methods to update work product data and extended data


sfeng88:
@joeferner It looks like some of the changes in this PR is also in https://github.com/v5analytics/visallo/pull/797? Is there an order we need to merge these PRs? Do we need both?

joeferner:
@sfeng88 the commit exists in both so #797 should probably go first

rygim:
What do we need the extra extended data object for? Can we just use the regular data object to encapsulate it?

web/plugins/map-product/src/main/resources/org/visallo/web/product/map/Map.jsx
```diff
@@ -31,6 +31,7 @@ define([
                     tools={this.getTools()}
                     generatePreview={generatePreview}
                     panelPadding={this.props.panelPadding}
+                    product={this.props.product}
```
jharwig: don't pass the product to open layers component. There's already a merged PR that gives the product to the map options extension

web/plugins/map-product/src/main/resources/org/visallo/web/product/map/OpenLayers.jsx
```diff
@@ -564,11 +564,12 @@ define([
 
         injectExtraPropsToTools() {
             const { map, cluster } = this.state;
+            const { product } = this.props;
```
jharwig: Remove (merged PR already addresses passing the property)

web/war/src/main/webapp/js/data/web-worker/store/product/reducer.js
```diff
@@ -125,7 +127,7 @@ define(['updeep'], function(u) {
 
     function hydrateExtendedData(item) {
         if (item.extendedData && _.isString(item.extendedData)) {
-            return { ...item, extendedData: JSON.parse(item.extendedData) };
+            return { ...item, extendedData: item.extendedData };
```
jharwig: This doesn't do anything anymore. Remove all references to `hydrate...` and remove this function

web/war/src/main/webapp/js/data/web-worker/store/product/reducer.js
```diff
@@ -140,4 +142,35 @@ define(['updeep'], function(u) {
         }, state);
     }
 
+    function updateData(state, {workspaceId, productId, key, value}) {
+        const updates = {};
+        updates[key] = value;
```
jharwig: Do you want `u.constant(value)` here? That replaces updates[key] with value. Currently it will only merge, so if value is an object it will do something like: `updates[key] = Object.assign({}, updates[key], value)`

web/war/src/main/webapp/js/data/web-worker/store/product/reducer.js
```diff
@@ -140,4 +142,35 @@ define(['updeep'], function(u) {
...
+                        }
+                    }
+                }
+            }
+        }, state);
+    }
+
+    function updateExtendedData(state, {workspaceId, productId, key, value}) {
+        const updates = {};
+        updates[key] = value;
```
jharwig: Same as above. Want `u.constant(value)`?

web/war/src/main/webapp/js/data/web-worker/store/product/reducer.js
joeferner: Is there an easy way to unit test this method? If not how about just test this method in isolation to verify it works like I'm expecting?

web/plugins/map-product/src/main/resources/org/visallo/web/product/map/OpenLayers.jsx
jharwig: I updated this for you since I added the test (and master had test fixes)

joeferner:
@rygim extended data is only pulled when the product is requested by itself. When the list of products is retrieved only data is returned.

web/war/src/main/webapp/js/data/web-worker/store/product/reducer.js
joeferner: done

PR #830, opened 2016-12-14
Default to allow all acl if property can't be found(3.0)

joeybrk372:
- [ ] @joeferner
- [x] @mwizeman @EvanOxfeld 

If property acl's can't be found then default to allow all.

CHANGELOG
Changed: If an element has a property not in it's domain, let the property still be addable/updateable/deletable.

### DIFF
--- a/web/war/src/main/webapp/js/util/acl.js
+++ b/web/war/src/main/webapp/js/util/acl.js
@@ -3,6 +3,12 @@ define([
 ], function(withDataRequest) {
     'use strict';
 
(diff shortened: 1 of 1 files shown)


rygim:
Would saving on the backend fail if a user tries to save something that is not in the ontology? I am just curious since it would be great to edit/add things like this on the front end but I am not sure how the backend would deal with it.

joeybrk372:
No it does not fail, you can save these properties on the backend.

PR #832, opened 2016-12-16
Fix browser extension import(3.0)

joeybrk372:
- [ ] @joeferner
- [x] @mwizeman @EvanOxfeld 

CHANGELOG
Fixed: Images imported through the Chrome browser extension could not publish the image's url property.


joeferner:
Using the default visibility doesn't look right. Shouldn't this be handled up stream. It looks like the visibility is coming from the property's visibility source, can we make the browser extension pass this as open or default instead.

joeybrk372:
we assign default visibilities to property metadata values, the problem here was that since they were taking the visibility of the property instead of the default we would be looking for/updating the wrong metadata values

EvanOxfeld:
@joeferner Are you ok with these changes?

PR #834, opened 2016-12-16
Don't show unresolve option if resolved entity is public

joeybrk372:
- [x] @joeferner
- [ ] @mwizeman @EvanOxfeld 

CHANGELOG
Fixed: Unresolve option was being displayed for term mentions/detected objects that were public entities which can't be unresolved.


EvanOxfeld:
Should validation be performed by the server as well?

joeybrk372:
it is, https://github.com/v5analytics/visallo/blob/master/web/web-base/src/main/java/org/visallo/web/routes/vertex/UnresolveTermEntity.java#L68

PR #837, opened 2016-12-20
Add search filter option to include child concepts

joeybrk372:
- [x] @joeferner
- [x] @mwizeman @EvanOxfeld 

filter-preview

CHANGELOG
Added: Optional search filter to include child concepts in search results if filtering by a specific concept (defaults to true).
Fixed: Dashboard card reports, such as the charts for entity counts and saved search aggregations, now return only the concept of the slice that is clicked in a popup search and not their child concepts.
Fixed: Search filters no longer initially show both concept and edge type filters.


jharwig:
👍 

Couple minor design changes:

* Since it's not just immediate children maybe "Include Descendants" is better with a title attribute of "Match entities with only this concept" when unchecked and "Match entities with this concept and any descendants" when checked.
* Can you hide it until selection and only show if there are descendants for the selected concept?
* Make it smaller font, and the margin between this element and the top should be closer than the next section to tie it together more.

decendants

jharwig:
Also make sure to test "Find related/Similar/Common" from graph. Changes to search sometimes break these. Test with search panel closed after browser refresh and after it's opened.

web/war/src/main/webapp/js/search/filters/filters.js
```diff
@@ -299,6 +302,13 @@ define([
             }
         };
 
+        this.setIncludeChildNodes = function(include) {
+            include = include !== undefined ? include : true;
```
EvanOxfeld: This could be simplified with default parameters: `this.setIncludeChildNodes = function(include = true) { }`

joeybrk372:
All changes requested complete and double-checked to work with from graph searches

PR #840, opened 2016-12-23
Fixed: Dashboard horizontal bar text clip colors in FF

joeferner:
- [x] @joeferner
- [ ] @mwizeman

FF doesn't scope the clip-path defs like Chrome does. This commit adds
the component id into the id to create a globally uniq id

CHANGELOG
Fixed: Dashboard horizontal bar text clip colors in FF

Before
before

After
after


joeferner:
see https://github.com/v5analytics/visallo-lts/issues/2035

PR #848, opened 2016-12-29
Added: context menu for edges

joeferner:
- [x] @joeferner
- [ ] @kunklejr @sfeng88 @diegogrz
- [x] @mwizeman

This commit changes the vertex context menu to react. It also adds
support for context menus for edges

Testing Instructions:
Requires testing all context menu items. Selecting multiple vertices with context menu. Context menu for edges.

CHANGELOG
Added: context menu for edges


web/war/src/main/webapp/js/app.js
```diff
@@ -318,20 +319,23 @@ define([
...
                 .then(function(_F) {
                     F = _F;
-                    return F.vertex.getVertexIdsFromDataEventOrCurrentSelection(data, { async: true });
+                    return F.vertex.getVertexAndEdgeIdsFromDataEventOrCurrentSelection(data, { async: true });
                 })
                 .then(function(elementIds) {
-                    return self.dataRequest('vertex', 'store', { vertexIds: elementIds })
+                    return Promise.all([
+                        self.dataRequest('vertex', 'store', { vertexIds: elementIds.vertexIds }),
+                        self.dataRequest('edge', 'store', { edgeIds: elementIds.edgeIds })
```
EvanOxfeld: Is there any downside to firing off both data requests regardless of whether vertex or edge IDs are present?

web/war/src/main/webapp/js/util/vertex/formatters.js
```diff
@@ -80,6 +92,47 @@ define([
...
 
+            getEdgeIdsFromDataEventOrCurrentSelection: function(data, opts) {
+                // Normalize the edgeIds sent from a edge menu event,
+                // also checking the current object selection
+                var edgeIds = [],
+                    options = opts || {},
+                    // Return a promise (will return more accurate list that
+                    // isn't susceptible to selectObjects -> objectsSelected
+                    // race condition
+                    async = options.async;
```
EvanOxfeld: Does it make sense to provide a sync option for this new call? IIRC `getVertexIdsFromDataEventOrCurrentSelection` only includes the option to avoid breaking old code or existing customer extensions but neither is the case here.

jharwig:
Does this add the menu to element lists? search, etc

web/war/src/main/webapp/js/app.js
joeferner: From what I can tell these methods look in the local cache first, fills in all those elements then if there are any left over will make the request base to the server (https://github.com/v5analytics/visallo/blob/master/web/war/src/main/webapp/js/data/web-worker/store/element/actions-impl.js#L28)

web/war/src/main/webapp/js/util/vertex/formatters.js
joeferner: fixed

web/war/src/main/webapp/js/util/vertex/menu.js
```diff
@@ -1,217 +1,50 @@
-
```
jharwig: create a new file `element/menu.js` and just put a stub in this file to redirect...

```
define(['../element/menu'], function(elementMenu) {
    console.warn('Component is deprecated, use element menu instead: "util/element/menu"');
    return elementMenu;
})

**Comment by jharwig** [comment]
Also how does the vertex menu extension work with this?

File: web/war/src/main/webapp/js/util/vertex/menu.js (hunk 3 above)
Comment (joeferner) [inline web/war/src/main/webapp/js/util/vertex/menu.js]: Not really sure what you mean by this. Where am I creating this file?

**Comment by joeferner** [comment]
@jharwig I added/changed docs for edge context menu items. They work the same except you use `org.visallo.edge.menu`

File: web/war/src/main/webapp/js/util/vertex/menu.js (hunk 3 above)
Comment (jharwig) [inline web/war/src/main/webapp/js/util/vertex/menu.js]: you have a flight stub called `vertex/menu` that calls to a react element menu. We should move the stub to better name.

File: docs/extension-points/front-end/elementMenu/index.md [hunk 4]
```diff
@@ -1,7 +1,7 @@
 Vertex Menu Plugin
```
sfeng88: Should we change the title to be `Element Menu Plugin` since it talks about both Vertex and Edges?

web/war/src/main/webapp/js/util/vertex/menu.js
joeferner: fixed

docs/extension-points/front-end/elementMenu/index.md
joeferner: updated

joeferner:
@jharwig @EvanOxfeld @sfeng88 I think I got all the comments addresses can you please re-review

PR #863, opened 2017-01-06
Graph Performance Improvements

jharwig:
- [x] @joeferner
- [x] @kunklejr @diegogrz
- [x] @mwizeman @sfeng88

Testing Instructions: Graph with many (~hundreds) items should be acceptable when select -> all.

Improve graph performance, especially when selecting
* Toolbar was making acl requests for all elements, even though we don't
need those for multi selection
* Element list was slow with many items – add a local infinite scroll
* Break up some functions so JITs can better optimize.
* Use native array methods for property formatting for less memory usage
* No force directed animation if many elements

CHANGELOG
Fixed: Performance improved when selecting large amounts of items


sfeng88:
@jharwig should this have a changelog?

joeybrk372:
please cherry-pick to 3.1 branch after merge

PR #868, opened 2017-01-06
Added: message when pasting elements

joeferner:
- [x] @joeferner
- [ ] @kunklejr @diegogrz
- [x] @mwizeman @sfeng88

This commit adds a message similar to the copy message when pasting
elements. This is added to let the user know the paste was successful when
viewing the map where it is not not always obvious especially when some
or all entities may not include geolocations

CHANGELOG
Added: message when pasting elements

### DIFF
--- a/core/core/src/main/resources/MessageBundle.properties
+++ b/core/core/src/main/resources/MessageBundle.properties
@@ -441,6 +441,8 @@ vertex.contextmenu.remove.subtitle=Remove from View
 
 element.clipboard.copy.one=Copied element
 element.clipboard.copy.some=Copied {0} elements
+element.clipboard.paste.one=Pasted element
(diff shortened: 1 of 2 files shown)


core/core/src/main/resources/MessageBundle.properties
```diff
@@ -441,6 +441,8 @@ vertex.contextmenu.remove.subtitle=Remove from View
 
 element.clipboard.copy.one=Copied element
 element.clipboard.copy.some=Copied {0} elements
+element.clipboard.paste.one=Pasted element
+element.clipboard.paste.some=Pasted  {0} elements
```
jharwig: looks like extra space here, probably won't show when rendered, though

joeferner: fixed

PR #877, opened 2017-01-12
Modified: WorkProductElements ontology update to only occur once

joeferner:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

For performance reasons updating the ontology should only occur at most
once per running of the application. It is also not necissary to clear
the ontology cache and is the resposibility of the ontology repository
to mange the state.


EvanOxfeld:
@joeferner @sfeng88 Is this fix worth cherry-picking to 3.1?

PR #886, opened 2017-01-18
Fix some visibility/updating issues(3.1)

joeybrk372:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

fixed issue where created edges won't show up on the graph until refreshing

CHANGELOG
Fixed: When a vertex on a map product was changed to have a visibility you are not authorized to see, the map would not open.
Fixed: When the visibility of a published element changed to something you were not authorized to see and on a case you weren’t actively in, it would not disappear/reappear until the browser was refreshed.

### DIFF
--- a/web/plugins/map-product/src/main/resources/org/visallo/web/product/map/Map.jsx
+++ b/web/plugins/map-product/src/main/resources/org/visallo/web/product/map/Map.jsx
@@ -125,7 +125,7 @@ define([
             const elementsSelectedById = { ..._.indexBy(this.props.selection.vertices), ..._.indexBy(this.props.selection.edges) };
             const productVertices = _.pick(this.props.elements.vertices, _.pluck(vertices, 'id'));
             const productEdges = _.pick(this.props.elements.edges, _.pluck(edges, 'id'));
-            const elements = _.compact(Object.values(productVertices).concat(Object.values(productEdges)));
+            const elements = Object.values(productVertices).concat(Object.values(productEdges));
             const geoLocationProperties = _.groupBy(this.props.ontologyProperties, 'dataType').geoLocation;
 
             return elements.map(el => {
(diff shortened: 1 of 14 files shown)


web/war/src/main/webapp/js/data/web-worker/store/element/actions-impl.js
```diff
@@ -121,21 +121,28 @@ define(['../actions', '../../util/ajax'], function(actions, ajax) {
...
+                } else if (isEdge) {
+                    const { inVertexId, outVertexId } = change;
+                    if (!inVertexId || !outVertexId || vertexInStore(inVertexId, outVertexId)) {
+                        dispatch(api.get({ workspaceId, edgeIds: [graphEdgeId], invalidate: true }));
+                    }
                 }
-            }
+            };
+
+            workspaceIds.forEach(updateOnWorkspace);
```
jharwig: Since visibility changes aren't sandboxed do we need to update all workspace caches on a visibility change?

web/plugins/map-product/src/main/resources/org/visallo/web/product/map/Map.jsx
```diff
@@ -126,7 +126,7 @@ define([
             const elementsSelectedById = { ..._.indexBy(this.props.selection.vertices), ..._.indexBy(this.props.selection.edges) };
             const productVertices = _.pick(this.props.elements.vertices, _.pluck(vertices, 'id'));
             const productEdges = _.pick(this.props.elements.edges, _.pluck(edges, 'id'));
-            const elements = Object.values(productVertices).concat(Object.values(productEdges));
+            const elements = _.compact(Object.values(productVertices).concat(Object.values(productEdges)));
```
joeferner: I just ran across this today. I'm wondering should `productSelectors.getElementsInProduct` (the source of this.props.elements) be fixed to not return null elements instead?

joeferner:
Is this just for 3.1, I see this on master as well.

web/war/src/main/webapp/js/data/web-worker/store/element/actions-impl.js
joeybrk372: Changed to update all workspaces

web/plugins/map-product/src/main/resources/org/visallo/web/product/map/Map.jsx
joeybrk372: No, the responsibility should fall on the product to determine how it handles null elements

jharwig: https://github.com/v5analytics/visallo/blob/master/web/war/src/main/webapp/js/data/web-worker/store/product/selectors.js#L51

PR #890, opened 2017-01-18
Graph loading showing unauthorized nodes

jharwig:
- [x] @joeferner
- [ ] @diegogrz
- [x] @mwizeman @sfeng88

Let the client know which product elements are unauthorized/deleted so they don't show loading nodes

Testing Instructions:
Create a new vertex on a workspace, then delete it. Refreshing shouldn't show a loading placeholder for that node. Use network devtool to throttle connection slower to see it.
Also test visibility changes using a shared workspace to a user that doesn't have a visibility. the node should appear/reappear live depending on visibility.

CHANGELOG
Fixed: Don't show placeholder nodes loading for unauthorized or deleted entities

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java
+++ b/core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java
@@ -12,9 +12,13 @@ import org.visallo.core.model.properties.VisalloProperties;
 import org.visallo.core.model.workspace.WorkspaceProperties;
 import org.visallo.core.user.User;
 import org.visallo.core.util.JSONUtil;
+import org.visallo.core.util.StreamUtil;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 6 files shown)


core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java
```diff
@@ -104,9 +104,12 @@ public JSONObject getExtendedData(
                     authorizations
             );
             for (Edge propertyVertexEdge : productVertexEdges) {
-                String other = propertyVertexEdge.getOtherVertexId(id);
+                Vertex otherVertex = propertyVertexEdge.getOtherVertex(id, authorizations);
```
joeferner: This call is going to be quite a bit slower than just getting the id. You should probably grab all the ids in a loop beforehand and grab them all at once using `graph.getVertices`.

jharwig: Done

PR #894, opened 2017-01-19
Fixes for graph selection when changing selection outside of graph

jharwig:
- [x] @joeferner
- [ ] @diegogrz
- [x] @mwizeman @sfeng88

Graph selection state could get in unknown state when selection changes are made outside of graph. (It uses a coalescing internal selection state to manage the many selection changes from cytoscape, and wasn't updating itself from outside changes)

Testing Instructions: Select multiple items, from multiple detail pane select a bar that will change the selection. Then try select-all in the graph, and clicking items in the graph. Selection state should work as expected

CHANGELOG
Fixed: Graph selection state could be unstable after changing the selection in the graph, then in external component, then in graph again.

### DIFF
--- a/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
+++ b/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
@@ -124,6 +124,9 @@ define([
         },
 
         componentWillReceiveProps(nextProps) {
+            if (nextProps.selection !== this.props.selection) {
+                this.resetQueuedSelection(nextProps.selection);
+            }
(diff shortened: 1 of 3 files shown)


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -124,6 +124,9 @@ define([
         },
 
         componentWillReceiveProps(nextProps) {
+            if (nextProps.selection !== this.props.selection) {
```
joeybrk372: `_.isEqual`

PR #918, opened 2017-01-27
Support unauthorized edge ends on the UI(3.1)

joeybrk372:
- [x] @joeferner
- [x] @diegogrz
- [ ] @mwizeman @sfeng88

Testing Instructions: Everywhere you can interact with edges when one or both vertices are not visible to the user.

CHANGELOG
Fixed: When one or more of a relationship's vertices were not visible to the user, the element inspector would appear blank when trying to inspect that edge.
Changed: When the other end of a relationship is not visible to the user, inspecting the relationships of a vertex will not include that vertex in the list.

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/termMention/TermMentionRepository.java
+++ b/core/core/src/main/java/org/visallo/core/model/termMention/TermMentionRepository.java
@@ -650,6 +650,9 @@ public class TermMentionRepository {
     ) {
         Authorizations authorizationsWithTermMentions = getAuthorizations(authorizations);
         Vertex vertex = graph.getVertex(vertexId, authorizationsWithTermMentions);
+
+        if (vertex == null) { return null; }
+
         Iterable<Vertex> termMentions = vertex.getVertices(
                 Direction.IN,
                 VisalloProperties.TERM_MENTION_LABEL_RESOLVED_TO,
--- a/core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
+++ b/core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
@@ -33,10 +33,7 @@ import org.visallo.web.clientapi.model.*;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.Locale;
+import java.util.*;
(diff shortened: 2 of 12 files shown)


joeybrk372:
will cherry-pick to master

core/core/src/main/java/org/visallo/core/model/termMention/TermMentionRepository.java
```diff
@@ -650,6 +650,9 @@ private Vertex findTermMention(
     ) {
         Authorizations authorizationsWithTermMentions = getAuthorizations(authorizations);
         Vertex vertex = graph.getVertex(vertexId, authorizationsWithTermMentions);
+
+        if (vertex == null) return null;
```
joeferner: Can you add `{}` around the return statement.

core/core/src/main/java/org/visallo/core/model/workspace/WorkspaceRepository.java
```diff
@@ -667,7 +664,7 @@ private void publishProperty(
             }
 
             if (foundProperty) {
-                Iterable<Vertex> termMentions;
+                Iterable<Vertex> termMentions = Collections.emptyList();
```
joeferner: Move new collection to else of the if.

web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexEdges.java
```diff
@@ -58,91 +54,41 @@ public ClientApiVertexEdges handle(
...
+        result.setTotalReferences(count);
+
+        return result;
     }
 
     /**
      * This is overridable so web plugins can modify the resulting set of edges.
      */
-    protected ClientApiVertexEdges.Edge convertEdgeToClientApi(Edge edge, String otherVertexId, Vertex otherVertex, String workspaceId, Authorizations authorizations) {
+    protected ClientApiVertexEdges.Edge convertEdgeToClientApi(Edge edge, Vertex otherVertex, String workspaceId, Authorizations authorizations) {
```
sfeng88: By modifying this method signature it might break other projects that are overriding this method. 

@EvanOxfeld Is this something we care about since we don't have a formal API?

sfeng88: after talking with @EvanOxfeld, can we depreciate the old method and add the new one?

PR #921, opened 2017-01-31
Shade org.xerial.snappy dependency to fix parquet when used with hadoop (3.1)

jharwig:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Testing Instructions: Try parquet import with open source and enterprise

NO CHANGELOG: first parquet release


web/plugins/structured-ingest/parquet/pom.xml
```diff
@@ -50,11 +50,78 @@
...
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <version>3.0.0</version>
```
sfeng88: Should this match the version in visallo/root? (https://github.com/v5analytics/visallo/blob/release-3.1/root/pom.xml#L134)

jharwig: oops, that shouldn't be included

PR #929, opened 2017-02-02
Fix map element visibility + add some tests(3.1)

joeybrk372:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Initially thought it was a product selector error which is why the tests got started. Changed the map product to also check the element store when deciding which elements to include.

no changelog


web/plugins/map-product/src/main/resources/org/visallo/web/product/map/Map.jsx
```diff
@@ -130,8 +130,8 @@ define([
         mapElementsToFeatures() {
             const { vertices, edges } = this.props.productElementIds;
             const elementsSelectedById = { ..._.indexBy(this.props.selection.vertices), ..._.indexBy(this.props.selection.edges) };
-            const productVertices = _.pick(this.props.elements.vertices, _.pluck(vertices, 'id'));
-            const productEdges = _.pick(this.props.elements.edges, _.pluck(edges, 'id'));
+            const productVertices = _.omit(_.pick(this.props.elements.vertices, _.pluck(vertices, 'id')), (vertex) => !vertex);
+            const productEdges = _.omit(_.pick(this.props.elements.edges, _.pluck(edges, 'edgeId')), (edge) => !edge);
```
jharwig: Why is this necessary? isn't it using the selector which removes these already?

joeybrk372: the selector checks both the element store and the product's extendedData to find the element. The element might have changed since the last time we requested extendedData.

jharwig: Still not understanding. `this.props.elements` are the vertices and edges from element store from the ids in the products extendedData. 

The selector that drives [`this.props.elements`](https://github.com/v5analytics/visallo/blob/4d975b2edb18b85319eaef797b259e23ffc05d19/web/war/src/main/webapp/js/data/web-worker/store/product/selectors.js#L56) shouldn't return null values, right? Since it's filtered by viewable?

jharwig: don't we just do something similar to the graph? https://github.com/v5analytics/visallo/blob/4d975b2edb18b85319eaef797b259e23ffc05d19/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx#L564

jharwig: We want to move these state transformations to selectors when possible for performance

joeybrk372: the graph also checks both: https://github.com/v5analytics/visallo/blob/4d975b2edb18b85319eaef797b259e23ffc05d19/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx#L605

the problem is that we don't request the product's extendedData again when we are receiving element updates through the websocket, I didn't change that since the extendedData might be a lot to keep sending over the network.

jharwig: It might check but it shouldn't need to anymore. That condition is not possible after the selector change

jharwig: An element update will invalidate the selector. So still not seeing why this is needed.

PR #933, opened 2017-02-02
Requery ontology to get parent properties

rygim:
Properties that were created from generated classes were unable to be imported due to errors that the ingest repository threw because it does not pull back parent concept properties. This behavior is different between the VertexiumOntologyRepository and the InMemoryOntologyRepository. This fix can be made in the VertexiumOntologyRepository by querying the getConceptsWithProperties() while trying to pull back the parent concept, but I was split on where I wanted to make the fix.

- [ ] @joeferner
- [ ] @diegogrz
- [ ] @mwizeman @sfeng88

Testing Instructions:
(These instructions may not be reasonable to actually test)
 - Generate code for entities of ontology that has two concepts, one parent and one child, with a property on the parent.
 - Ensure that you are using the VertexiumOntologyRepository 
 - Try and ingest the entities using the generated code by setting a property on the child-concept that is in on the parent concept. Use the IngestRepository to save the created data. There will be no validation errors.

Points of Regression:
N/A

CHANGELOG
Fixed: IngestRepository now correctly traverses the concept heirarchy and pulls back the parent concept properties when deciding if an entity should be ingested.

### DIFF
--- a/core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepository.java
+++ b/core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepository.java
@@ -468,7 +468,10 @@ public class VertexiumOntologyRepository extends OntologyRepositoryBase {
         if (parentConceptVertex == null) {
             return null;
         }
-        return createConcept(parentConceptVertex);
+
+        Concept parentConcept = createConcept(parentConceptVertex);
+
+        return getConceptByIRI(parentConcept.getIRI());
     }
 
     private List<Concept> toConcepts(Iterable<Vertex> vertices) {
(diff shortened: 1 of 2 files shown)


rygim:
@joeferner I can make this fix here or I can add something like 

``` Concept parentConcept = createConcept(parentConceptVertex);

        for(Concept conceptWithProperties : getConceptsWithProperties()){
            if(conceptWithProperties.equals(parentConcept)){
                return conceptWithProperties;
            }
        }```

to here https://github.com/v5analytics/visallo/blob/master/core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepository.java#L470-L471. Both are valid fixes to the problem. Do you have a preference?

joeferner:
@rygim I like the second fix in the ontology repository since it handles this problem everywhere not just the ingest repo. Could the fix in `getParentConcept` just call `getConceptByIRI` like your original fix?

rygim:
Closed in lieu of https://github.com/v5analytics/visallo/pull/936

PR #936, opened 2017-02-03
Fix ontology repository to return parent properties

rygim:
Whenever the the vertexium ontology repository would query for a parent concept, it would not return the properties that were associated with that concept. This is important when checking inheritance in the case of the generated ontology code, and would cause the IngestRepository to start throwing failures because it tries to prevent invalid data from going into Visallo. By returning a more fully inflated parent concept the VertexiumOntologyRepository not only matches what is expected, but now works the same as in in memory implementation

- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Testing Instructions:
(These instructions may not be reasonable to actually test)

Generate code for entities of ontology that has two concepts, one parent and one child, with a property on the parent.
Ensure that you are using the VertexiumOntologyRepository
Try and ingest the entities using the generated code by setting a property on the child-concept that is in on the parent concept. Use the IngestRepository to save the created data. There will be no validation errors.

Points of Regression: 
N/A

CHANGELOG
Fixed: VertexiumOntologyRepository returns properties on the parent concept when returning the parent concept


core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepository.java
```diff
@@ -468,7 +468,10 @@ public Concept getParentConcept(final Concept concept) {
         if (parentConceptVertex == null) {
             return null;
         }
-        return createConcept(parentConceptVertex);
+
+        Concept parentConcept = createConcept(parentConceptVertex);
+
+        return getConceptByIRI(parentConcept.getIRI());
```
EvanOxfeld: Is it worth adding a unit test?

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepository.java
```diff
@@ -468,7 +468,10 @@ public Concept getParentConcept(final Concept concept) {
         if (parentConceptVertex == null) {
             return null;
         }
-        return createConcept(parentConceptVertex);
+
+        Concept parentConcept = createConcept(parentConceptVertex);
```
joeferner: You can probably by pass the step of `createConcept` and just grab the IRI off the vertex if you want.

rygim: Fixed

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepository.java
rygim: Added

PR #941, opened 2017-02-03
Web: Fix saving user UI preferences (3.1)

EvanOxfeld:
Resolves issues saving snap to grid and find path settings. Will cherry-pick to master when approved.

- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Testing Instructions: Confirm snap to grid and most recent find path settings are stored and reused in the current user session and across user sessions.

Points of Regression:

CHANGELOG
Fixed: Saving snap to grid and find path settings

### DIFF
--- a/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/popovers/FindPathPopoverContainer.jsx
+++ b/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/popovers/FindPathPopoverContainer.jsx
@@ -19,7 +19,12 @@ define([
         },
         (dispatch, props) => {
             return {
-                setUserPreferences: (value) => dispatch(userActions.setUserPreference(PREFERENCE_NAME, JSON.stringify(value)))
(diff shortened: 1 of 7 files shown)


joeferner:
@EvanOxfeld what are the benefits of passing all the arguments like this `({user})` or `({name,value})` seems a bit silly to me. I can understand if the method takes 4+ arguments but for something like `setUserPreference` it seems a bit over kill and just adds `{` all over the place.

EvanOxfeld:
@joeferner The need for actions vs actions-impl pretty much forces this. If the action and the impl both destructure a single argument they can always have the same function signature, otherwise 2+ arguments cause different signatures or the need for some arguments object trickery. Although I can personally go either way when the action takes a single argument, for 2+ it's especially awful if the action function and the impl have different signatures, and in this case the difference caused the  UI preferences defect fixed by this pull request.

@jharwig @joeybrk372 Thoughts?

PR #948, opened 2017-02-09
Web: Snap to grid graph position fixes

EvanOxfeld:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Fixes issues dropping an element onto the graph when snap to grid is enabled and elements snapping back to the current position if not moved to a new position on the grid.

Moves position change logic when snap to grid is on out of the reducer and into the action creator function to prevent side effects.

Will cherry-pick to master when approved.

Testing Instructions: Verify moving entities on the graph with and without snap to grid enabled. Verify that when snap to grid is enabled, small entities movements cause the entity to snap back to its original position. Verify that elements can be added to the graph with and without snap to grid enabled.

Points of Regression: None known

CHANGELOG
Fixed: Entities can be added and moved when snap to grid is enabled.

### DIFF
--- a/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/__tests__/worker/actionsImplTest.js
+++ b/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/__tests__/worker/actionsImplTest.js
@@ -1,0 +1,309 @@
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+import actions from '../../worker/actions-impl';
+
+jest.mock('data/web-worker/store/actions', () => ({
+    protectFromMain: jest.fn()
+}));
+jest.mock('data/web-worker/util/ajax', () => () =>
+    Promise.resolve({
+        extendedData: {
+            edges: []
+        }
+    }));
+jest.mock('data/web-worker/store/product/actions-impl', () => ({
+    update: jest.fn(() => ({ type: 'MOCK_PRODUCT_UPDATE' }))
+}));
+jest.mock('data/web-worker/store/element/actions-impl', () => ({
+    get: jest.fn(() => ({ type: 'MOCK_ELEMENT_GET' }))
+}));
+
+const middlewares = [ thunk ];
+const mockStore = configureMockStore(middlewares);
+
+describe('graph plugin actions', () => {
+    describe('PRODUCT_GRAPH_SET_POSITIONS', () => {
+        it('should not dispatch if the workspace is read only', () => {
+            const store = mockStore({
+                workspace: {
+                    currentId: 'WORKSPACE1',
+                    byId: {
+                        WORKSPACE1: {
+                            editable: false
+                        }
+                    }
+                },
+                product: {
+                    workspaces: {
+                        WORKSPACE1: {
+                            products: {
+                                PRODUCT1: {}
+                            }
+                        }
+                    }
+                }
+            });
+            expect(
+               store.dispatch(
+                actions.setPositions({
+                    productId: 'PRODUCT1'
+                })
+               )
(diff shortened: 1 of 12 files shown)


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/package.json
```diff
@@ -31,13 +31,23 @@
...
   "jest": {
     "moduleNameMapper": {
       "components/NavigationControls": "<rootDir>/__mocks__/components/NavigationControls.jsx",
       "components/RegistryInjectorHOC": "<rootDir>/__mocks__/components/RegistryInjectorHOC.jsx",
+      "data/web-worker/store/actions": "<rootDir>/__mocks__/data/web-worker/store/actions.js",
+      "data/web-worker/store/element/actions-impl": "<rootDir>/__mocks__/data/web-worker/store/element/actions-impl.js",
+      "data/web-worker/store/product/actions-impl": "<rootDir>/__mocks__/data/web-worker/store/product/actions-impl.js",
+      "data/web-worker/store/product/selectors": "<rootDir>/__mocks__/data/web-worker/store/product/selectors.js",
+      "data/web-worker/store/selection/actions-impl": "<rootDir>/__mocks__/data/web-worker/store/selection/actions-impl.js",
+      "data/web-worker/store/user/actions-impl": "<rootDir>/__mocks__/data/web-worker/store/user/actions-impl.js",
```
EvanOxfeld: Jest warnings about duplicate manual mocks can be ignored -- currently Jest only considers the module name, not the full path.

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/worker/actions-impl.js
```diff
@@ -61,35 +63,39 @@ define([
...
+            const addingNewVertices = Object.keys(updateVertices)
+                .some(id => !(id in byId));
 
-            updateVertices = _.mapObject(updateVertices, pos => {
-                return _.mapObject(pos, v => Math.round(v));
-            });
+            updateVertices = _.mapObject(updateVertices, pos =>
+                _.mapObject(pos, v => Math.round(v))
+            );
             dispatch({
```
joeybrk372: dispatch is getting called twice if snapToGrid is true

EvanOxfeld: Yes, that's intentional -- state has to update twice in the case where the entity needs to snap back to its original position. In an attempt to make it clearer I added a test named should `dispatch PRODUCT_GRAPH_SET_POSITIONS twice when snap to grid is enabled`. Would it be even clearer if the second action had a different type?

joeybrk372: Right thanks for the clarification, seems fine as is.

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/__tests__/worker/actionsImplTest.js
```diff
@@ -0,0 +1,309 @@
...
+                        }
+                    },
+                    snapToGrid: true
+                })
+            ).then(() => {
+                expect(store.getActions()).toEqual(expectedActions);
+            });
+        });
+
+        it('should dispatch PRODUCT_GRAPH_SET_POSITIONS twice on a new vertex',  () => {
```
EvanOxfeld: Test name should be `should dispatch PRODUCT_GRAPH_SET_POSITIONS twice on a new vertex when snap to grid is enabled`

PR #956, opened 2017-02-13
Allow structured ingest to map multiple of the same property(3.1)

joeybrk372:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

CHANGELOG
Fixed: Mapping the same properties to multiple columns during structured ingest would not create multiple properties.


rygim:
One thing that we may have lost when we created this is the ability to have dependent properties. Take for example the following file (github doesnt like csvs so it is a txt file here):

[structured-file-dependent-properties.csv.txt](https://github.com/visallo/visallo/files/791459/structured-file-dependent-properties.csv.txt)

and look at what happens when I try to import it with two first names (you will have to unzip it since github doesnt like me uploading just movie files):

[structured-import.mov.zip](https://github.com/visallo/visallo/files/791465/structured-import.mov.zip)

In the element inspector, every property gets its own key so each name gets its own key, so it shows on its own line but we don't necessarily want that (like at time index 1:00 where Jeff's name is split as "Jeff" and "Kunkle" rather than what it should likely be: "Jeff Kunkle"

Fortunately/unfortunately this isn't even worse than what it previously did because when I ran this on the old code, I can get it to do this even importing it correctly:
image

which is also incorrect.

Unfortunately, I dont have any great solutions for you other than possibly some more UI fanciness. I'm not necessarily saying that it is a blocker anyways, but I just wanted to see what you think about it because  it makes it a little bit interesting when importing data when some properties need to go together and some don't.

joeybrk372:
Yeah, seems like there is more work involved to support this. I'm closing this pr and will take another look at what is appropriate for 3.1

PR #960, opened 2017-02-15
Upgrade Cytoscape from 2.7.11 to 2.7.14(3.1)

joeybrk372:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Resolves an issue with requestAnimationFrame in IE which was present in `2.7.11`

no changelog


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/cytoscapeCorsFix.js
```diff
@@ -25,24 +28,28 @@ define(['underscore'], function(_) {
         var cache = imageCache[ url ];
 
         if( cache ){
-            if( !cache.image.complete ){
-                cache.image.addEventListener('load', onLoad);
-            }
+          if( !cache.image.complete ){
```
joeferner: 4 spaces?

joeybrk372: This is a monkeypatch of cytoscape's `getCachedImage`. I would rather keep source formatting to distinguish it

joeferner: Can you put it in a new function (maybe called `applyCytoscapeGetCacheImageFix` or something) to comment it and distinguish it.

joeybrk372: the function is called `fixedGetCachedImage` and is in cytoscapeCorsFix.js which returns `fixCytoscapeCorsHandling`

jharwig: This isn't needed anymore. My PR was merged. We need to remove this fix and set the `background-crossorigin` style in the default css. Then test in IE to confirm

joeybrk372: Yep, your fix went into cytoscape 2.7.12. I looked at it and set `background-image-crossorigin = 'use-credentials'` here already, but I believe the pr you submitted isn't quite correct so we still need to override this function

PR #962, opened 2017-02-15
Redux undo support and restoration of graph undo/redo 3.1

EvanOxfeld:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

See individual commits for more detail. We should probably add map product undo/redo before finalizing the APIs. Will cherry-pick to master when approved.

Testing Instructions: Verify undo and redo of moving vertices on the graph, adding vertices to the graph, removing vertices from the graph. Verify that these operations can be performed in various combinations.

Points of Regression: Potentially other graph updates that result in position changes

CHANGELOG
Fixed: Restore graph undo/redo using keyboard shortcut


web/war/src/main/webapp/js/data/withWorkspaces.js
```diff
@@ -16,6 +16,51 @@ define(['util/undoManager'], function(UndoManager) {
             this.on('switchWorkspace', this.onSwitchWorkspace);
             this.on('updateWorkspace', this.onUpdateWorkspace);
 
+            this.on(window, 'keydown', function(event) {
```
joeybrk372: please register these keyboard shortcuts and then listen for the event here, like: https://github.com/v5analytics/visallo/blob/master/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/GraphContainer.jsx#L198-L206

EvanOxfeld: The below code is copy/pasted from Visallo 2.x with the exception of dispatching the redux actions. I'm not sure the keyboard shortcut extension point supports everything necessary here -- does it support different shortcuts based on OS or a three key combination with the shift and meta keys? /cc @jharwig

joeybrk372: Yes, the extension point normalizes meta keys and will do a three key combination as well.

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/plugin.js
```diff
@@ -17,6 +17,8 @@ require(['configuration/plugins/registry'], function(registry) {
     })
 
     $(document).on('applicationReady currentUserVisalloDataUpdated', function() {
+        const redoKey = ~navigator.userAgent.indexOf('Mac OS X') ? 'shift-meta-z' : 'meta-y';
```
jharwig: Should we expose `isMac` in [formatters](https://github.com/visallo/visallo/blob/master/web/war/src/main/webapp/js/util/formatters.js#L19) as `F.platform.isMac` and use that instead?

Or should we update `registerKeyboardShortcuts` to accept `redo` and `undo`. We already accept `delete`, etc. And put the checking in there

web/war/src/main/webapp/js/data/web-worker/store/undo/reducer.js
```diff
@@ -0,0 +1,79 @@
...
+                    undo,
+                    redo
+                };
+                const scopedState = state[scope] || { undos: [], redos: [] };
+                const { undos, redos } = scopedState;
+
+                return {
+                    ...state,
+                    [scope]: {
+                        undos: [ ...undos, newUndo ],
```
jharwig: Is there a stack limit?

jharwig: talking with @EvanOxfeld there is no limit, but I don't think it's an issue as we aren't storing much at all yet. When we open this up we should add some checking on object size (maybe only in `devMode` using something like [object-sizeof](https://www.npmjs.com/package/object-sizeof) to prevent huge undo stacks

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/plugin.js
EvanOxfeld: `registerKeyboardShortcuts` seems like the most appropriate place for a consistent redo shortcut across plugins.

PR #965, opened 2017-02-16
Docs: Update makefile clone URL

EvanOxfeld:
- [x] @joeferner
- [x] @diegogrz
- [ ] @mwizeman @sfeng88

`make clone` only works for me with the https git URL.

Testing Instructions: None

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -25,7 +25,7 @@ clean:
 
 clone:
 	@rm -rf docs.visallo.org
-	@git clone -n git@github.com:v5analytics/docs.visallo.org
+	@git clone -n https://github.com/visallo/docs.visallo.org.git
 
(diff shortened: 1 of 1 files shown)


docs/Makefile
```diff
@@ -25,7 +25,7 @@ clean:
 
 clone:
 	@rm -rf docs.visallo.org
-	@git clone -n git@github.com:v5analytics/docs.visallo.org
+	@git clone -n https://github.com/v5analytics/docs.visallo.org.git
```
sfeng88: @EvanOxfeld we will need to change this to `visallo` instead of `v5analytics`

EvanOxfeld: Good catch. I submitted this PR prior to the org name change.

PR #973, opened 2017-02-17
Fix: vertex image upload metadata(3.1)

joeybrk372:
- [ ] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Testing Instructions: After adding an image to a vertex, the image's history should show Created Image, as well as Title and Source property modifications by user.

CHANGELOG
Fixed: The properties automatically added to an image when you upload the image to another concept weren't showing the correct author and timestamp.

### DIFF
--- a/web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexUploadImage.java
+++ b/web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexUploadImage.java
@@ -102,6 +102,8 @@ public class VertexUploadImage implements ParameterizedHandler {
 
         Metadata metadata = new Metadata();
         VisalloProperties.VISIBILITY_JSON_METADATA.setMetadata(metadata, visibilityJson, visibilityTranslator.getDefaultVisibility());
+        VisalloProperties.MODIFIED_DATE_METADATA.setMetadata(metadata, new Date(), visibilityTranslator.getDefaultVisibility());
+        VisalloProperties.MODIFIED_BY_METADATA.setMetadata(metadata, user.getUserId(), visibilityTranslator.getDefaultVisibility());
 
         String title = imageTitle(entityVertex);
         ElementBuilder<Vertex> artifactVertexBuilder = convertToArtifact(file, title, visibilityJson, metadata, user, visibility);


web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexUploadImage.java
```diff
@@ -102,9 +106,12 @@ public ClientApiElement handle(
 
         Metadata metadata = new Metadata();
         VisalloProperties.VISIBILITY_JSON_METADATA.setMetadata(metadata, visibilityJson, visibilityTranslator.getDefaultVisibility());
+        VisalloProperties.MODIFIED_DATE_METADATA.setMetadata(metadata, new Date(), visibilityTranslator.getDefaultVisibility());
+        VisalloProperties.MODIFIED_BY_METADATA.setMetadata(metadata, userRepository.getSystemUser().getUserId(), visibilityTranslator.getDefaultVisibility());
+
 
         String title = imageTitle(entityVertex);
-        ElementBuilder<Vertex> artifactVertexBuilder = convertToArtifact(file, title, visibilityJson, metadata, user, visibility);
+        ElementBuilder<Vertex> artifactVertexBuilder = convertToArtifact(file, title, visibilityJson, metadata, userRepository.getSystemUser(), visibility);
```
joeferner: System user doesn't seem right. We should elevate unless absolutely necessary.

joeybrk372: I don't think we should use the actual user unless it was their direct action. The properties/image vertex created are side effects of the user's direct action to add it to the vertex.

sfeng88: @joeybrk372 From what I remember, I think that is inconsistent with the other gpws. Please correct me if I'm wrong.

joeferner: Isn't this a direct action of the user uploading an image?

joeybrk372: I could be convinced into using the user for the image vertex, but I still believe the other added properties should be made by the system. 

@sfeng88 For gpws we get the user from the gpw data, but I'm unsure what connection you are making to this route

sfeng88: @joeybrk372 The reason why I brought up the GPWs is because lets say a user imports a image from the UI, our image metadata gpw automatically created properties. For those properties, I'm pretty sure the user is the user who imported the image. Isn't that the same concept here?

joeybrk372: After looking more closely at the gpw's: no, we run them as system user. Which works out, because an investigator is not being attributed for these automatically generated properties that they have not personally verified.

sfeng88: @joeybrk372  you're right. I vote for consistency so this is fine with me.

joeferner: But doesn't this line here get invoked directly from a user action? It's in a route. I get why the GPWs are doing what they are doing because they are augmenting the data.

joeybrk372: My thought was the additional properties added to the image vertex weren't directly created by the user, but I can also see your reasoning about encompassing those as an effect of the user action. I'm still leaning towards system user for those properties but I can change to the actual user if you feel it makes more sense.

jharwig: I think that since this is a route it should be the user, not system.

especially since the `createArtifact` method uses the user as the `modifiedBy` property. That should be the user.

jharwig: https://github.com/visallo/visallo/blob/4484e75f9981748cb36183d4c58473760aa9e6ea/web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexUploadImage.java#L201

EvanOxfeld: @jharwig Although the user initiated the action, the source and title were set by Visallo without user confirmation -- I'd say that's a system action. One step further, any time Visallo derives a fact the modified by metadata should be set to system user. For example, when creating a raw entity from an upload, the filename should always be set by the system user.

joeybrk372:
Changed to use upload user for now. Will warrant a discussion for whether that is the correct user for derived data like this case long term.

PR #987, opened 2017-02-22
Add scrolling to tabs in saved search table(3.1)

joeybrk372:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

scrollButtons

CHANGELOG
Added: Scroll arrows to page entity and relationship tabs in the saved search table.


jharwig:
I think the arrows should be on right of the tabs, and maybe show how many pages?

joeybrk372:
I can move the arrows to the right, but I'm not convinced for showing number of pages. I think just the arrows are still functional and like the cleaner look.

PR #989, opened 2017-02-22
Fix sharing dropdown overlap (3.1)

EvanOxfeld:
- [x] @joeferner
- [x] @diegogrz
- [ ] @mwizeman @sfeng88

Converts case shareRows to flexbox. Will submit a pull request to master when approved.

Testing Instructions: Verify that the user name displays with an ellipsis when necessary and the permission dropdown does not overlap at various edit case panel widths in the supported browsers.

Points of Regression: None. Changes isolated to the edit case form.

CHANGELOG
Fixed: Case sharing username and permission dropdown overlap issues

### DIFF
--- a/web/war/src/main/webapp/less/workspaces/workspaces.less
+++ b/web/war/src/main/webapp/less/workspaces/workspaces.less
@@ -119,6 +119,7 @@ li {
   .badge {
     cursor: pointer;
     margin: 0.25em 0 0 0.5em;
+    flex: 0 0 auto;
   }
 
   .popover {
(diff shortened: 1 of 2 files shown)


EvanOxfeld:
I'm considering converting the share row template to a react component while I'm here. The template seems isolated to the edit case panel.

EvanOxfeld:
Reopening. Converting the share row template to React requires readjusting the popover as well, which is too intrusive for 3.1. I'll do the conversion in the future.

web/war/src/main/webapp/less/workspaces/workspaces.less
```diff
@@ -119,6 +119,7 @@ li {
   .badge {
     cursor: pointer;
     margin: 0.25em 0 0 0.5em;
+    flex: 0 1 auto;
```
jharwig: should the badge be `0 0 auto`? So it doesn't shrink smaller?

EvanOxfeld: Fixed now -- `0 0 auto` is better. Thanks!

PR #990, opened 2017-02-23
Invalidate multiple property mappings structured ingest(3.1)

joeybrk372:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

ingest-error

CHANGELOG
Fixed: Structured ingest tool shows error when trying to map a property to more than one column.


jharwig:
Would it be better to filter to list to not show properties in the first place?

joeybrk372:
I did consider filtering them, but decided that it might be better to have all properties viewable instead of hiding them if they're trying to make a choice. Ideally I would have liked to have the property select component display disabled options but that was a bigger change because it currently doesn't support that.

jharwig:
okay, fair enough. Can you adjust some alignment? Left align the text in input with the error, and add some margin-bottom so the checkbox line has more room.
slice

PR #996, opened 2017-02-23
Use vertex context menu selection configuration(3.1)

joeybrk372:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Testing Instructions: Certain context menu options should not appear with differing numbers of vertices selected. (e.g. History, Search by Title are both limited to only one vertex)

CHANGELOG
Fixed: Entity context menu options still appeared when the number of selected entities did not match their configuration for the acceptable amount.


jharwig:
We also need to cherry-pick for master. But I think this file is now element/menu

joeybrk372:
Reverted docs change.

I'm holding off on master pr because there are a few changes to the element menu that need to happen in 3.2 that affect how this is done. I made a note in those other issues

PR #1011, opened 2017-02-28
Update docs to coincide with Github transition to visallo

kunklejr:
- [x] @joeferner
- [x] @diegogrz
- [ ] @mwizeman @sfeng88
- [ ] @joeybrk372 @rygim @jharwig @EvanOxfeld

Testing Instructions:
Verify that all occurrences of v5analytics in Github URLs have been replaced with visallo

Points of Regression:

CHANGELOG
Documentation: Updated docs to coincide with Github transition from v5analytics to visallo

PR #1012, opened 2017-02-28
Update docs to coincide with Github transition to visallo (3.1)

kunklejr:
See #1011


kunklejr:
Will redo these changes after major documentation update from @jharwig

PR #1013, opened 2017-02-28
Workers can ack tuples instead of message ids

rygim:
This gives us the ability to use customized worker spouts to implement custom functionality inside of the workers and spouts.

- [ ] @joeferner
- [ ] @diegogrz
- [ ] @mwizeman @sfeng88

Testing Instructions: N/A, there isnt anything to test

Points of Regression: A couple of interfaces have changed so custom code written against custom queues may need to change.

CHANGELOG
Changed: Interfaces relating the Worker Queues


core/core/src/main/java/org/visallo/core/model/WorkQueueNames.java
```diff
@@ -8,16 +8,20 @@
 public class WorkQueueNames {
     private static final String GRAPH_PROPERTY_QUEUE_NAME = "graphProperty";
     private static final String LONG_RUNNING_PROCESS_QUEUE_NAME = "longRunningProcess";
+    private static final String DEADLETTER_RUNNING_PROCESS_QUEUE_NAME = "deadLetter";
```
rygim: This comes from a custom implementation of our queues but was the most appropriate here. Should it be taken out and moved or left in here?

core/core/src/main/java/org/visallo/core/ingest/WorkerSpout.java
```diff
@@ -8,11 +8,11 @@ public void close() {
 
     }
 
-    public void ack(Object msgId) {
+    public void ack(WorkerTuple workerTuple) {
```
sfeng88: please deprecated the old methods and add the new ones

core/core/src/main/java/org/visallo/core/ingest/WorkerSpout.java
```diff
@@ -8,11 +8,11 @@ public void close() {
...
 
     }
 
-    public void fail(Object msgId) {
+    public void fail(WorkerTuple workerTuple) {
```
sfeng88: same comment as above

core/core/src/main/java/org/visallo/core/ingest/WorkerSpout.java
```diff
@@ -12,11 +12,27 @@ public void ack(WorkerTuple workerTuple) {
 
     }
 
+    /**
+     * @deprecated  replaced by {@link #ack(WorkerTuple)} ()}
+     */
+    @Deprecated
+    public void ack(Object msgId) {
+        ack(new WorkerTuple(msgId, new byte[0]));
```
rygim: @sfeng88 Does this look good for the deprecated methods? This class originally had no functionality in it, but since I wanted it to work correctly I added some implementation to call the correct method if super.ack is called. Since I can't be sure if anyone actually overrode the methods (because they are both going to have bodies), I sort of took this approach. It could potentially be confusing until these methods are actually removed. Do you have any suggestions?

core/core/src/main/java/org/visallo/core/ingest/WorkerSpout.java
```diff
@@ -8,12 +8,28 @@ public void close() {
...
+    public void ack(WorkerTuple workerTuple) {
+
+    }
+
+    /**
+     * @deprecated  replaced by {@link #ack(WorkerTuple)} ()}
+     */
+    @Deprecated
     public void ack(Object msgId) {
+        ack(new WorkerTuple(msgId, new byte[0]));
```
rygim: @sfeng88 Does this look good for the deprecated methods? This class originally had no functionality in it, but since I wanted it to work correctly I added some implementation to call the correct method if super.ack is called. Since I can't be sure if anyone actually overrode the methods (because they are both going to have bodies), I sort of took this approach. It could potentially be confusing until these methods are actually removed. Do you have any suggestions?

core/core/src/main/java/org/visallo/core/ingest/WorkerSpout.java
```diff
@@ -8,12 +8,28 @@ public void close() {
...
 
     }
 
+    /**
+     * @deprecated  replaced by {@link #fail(WorkerTuple)} ()}
+     */
+    @Deprecated
     public void fail(Object msgId) {
-
+        fail(new WorkerTuple(msgId, new byte[0]));
```
rygim: Same as above

core/core/src/main/java/org/visallo/core/ingest/WorkerSpout.java
rygim: Once we figure out if this change is ok, I will cherry pick these changes to the other branches

sfeng88: @rygim I think this is fine. we can remove these methods with the next release.

PR #1014, opened 2017-02-28
Workers can ack tuples instead of message ids (release-3.1)

rygim:
This gives us the ability to use customized worker spouts to implement custom functionality inside of the workers and spouts.

Master PR: https://github.com/visallo/visallo/pull/1013

- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Testing Instructions: N/A, there isnt anything to test

Points of Regression: A couple of interfaces have changed so custom code written against custom queues may need to change.

CHANGELOG
Changed: Interfaces relating the Worker Queues


core/core/src/main/java/org/visallo/core/ingest/WorkerSpout.java
```diff
@@ -8,11 +8,11 @@ public void close() {
 
     }
 
-    public void ack(Object msgId) {
+    public void ack(WorkerTuple workerTuple) {
```
sfeng88: please deprecated the old methods and add the new ones

core/core/src/main/java/org/visallo/core/ingest/WorkerSpout.java
```diff
@@ -8,11 +8,11 @@ public void close() {
...
 
     }
 
-    public void fail(Object msgId) {
+    public void fail(WorkerTuple workerTuple) {
```
sfeng88: please deprecated the old methods and add the new ones

PR #1017, opened 2017-02-28
Search interface improvements (3.1)

jharwig:
- [x] @joeferner
- [x] @diegogrz
- [ ] @mwizeman @sfeng88

* Supports searching multiple concepts and edgeLabels (ORed together)
* Unified style between search sections (related, concept, sort, filter)
* Supports the previous search format so no migration necessary to make old searches load in new search tool, or execute them.

Testing Instructions: Test searching for entities using multiple concepts. Similar for edges. The results should be one of the selected concept/edgeLabels.

Points of Regression: None

CHANGELOG
Added: Multiple concepts and edge labels can be added to entity/relation search for an `IN` search

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepository.java
+++ b/core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepository.java
@@ -133,7 +133,11 @@ public interface OntologyRepository {
 
     void addConceptTypeFilterToQuery(Query query, String conceptTypeIri, boolean includeChildNodes);
 
+    void addConceptTypeFilterToQuery(Query query, String[] conceptTypeIri, boolean[] includeChildNodes);
+
     void addEdgeLabelFilterToQuery(Query query, String edgeLabel, boolean includeChildNodes);
+
+    void addEdgeLabelFilterToQuery(Query query, String[] edgeLabel, boolean[] includeChildNodes);
 
     void updatePropertyDependentIris(OntologyProperty property, Collection<String> dependentPropertyIris);
 
--- a/core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepository.java
+++ b/core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepository.java
@@ -9,6 +9,7 @@ import org.vertexium.Authorizations;
 import org.vertexium.query.Query;
 import org.visallo.core.model.properties.types.VisalloProperty;
 import org.visallo.core.security.VisalloVisibility;
+import org.visallo.web.clientapi.model.ClientApiObject;
 import org.visallo.web.clientapi.model.ClientApiOntology;
 
 import java.io.File;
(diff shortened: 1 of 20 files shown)


core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepository.java
```diff
@@ -133,8 +133,12 @@ Relationship getOrCreateRelationshipType(
 
     void addConceptTypeFilterToQuery(Query query, String conceptTypeIri, boolean includeChildNodes);
 
+    void addConceptTypeFilterToQuery(Query query, String[] conceptTypeIri, boolean[] includeChildNodes);
```
joeferner: Instead of two arrays needing to be matched can we make an object to store conceptType and includeChildNodes.

core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepositoryBase.java
```diff
@@ -1266,46 +1268,63 @@ protected boolean determineSearchable(
...
-                for (Concept c : childConcepts) {
-                    conceptIds[count] = c.getIRI();
-                    count++;
-                }
-                query.has(VisalloProperties.CONCEPT_TYPE.getPropertyName(), Contains.IN, conceptIds);
+    @Override
+    public void addConceptTypeFilterToQuery(Query query, String[] conceptTypeIri, boolean[] includeChildNodes) {
+        checkNotNull(query, "query cannot be null");
+        checkArgument(conceptTypeIri.length == includeChildNodes.length, "Must include same size arrays");
+        if (conceptTypeIri.length == 0) return;
```
joeferner: code style: all if's get `{}` -> `if (...) { return; }`

web/war/src/main/webapp/js/search/types/withSearch.js
```diff
@@ -68,11 +68,12 @@ define([
                     this.trigger($searchResults, 'paneResized');
                 }
             });
-            this.on('clearSearch', function() {
+            this.on('clearSearch', function(event, data) {
                 this.hideSearchResults();
+                console.log(event.target, data)
```
joeferner: remove

joeferner:
If I select two concepts of differing types what properties are listed in the filters, union or intersection?

This is probably out of the scope of this PR but should we remove "match" type and allow users to search both entities and relationships at the same time (using https://github.com/visallo/vertexium/blob/master/core/src/main/java/org/vertexium/query/Query.java#L26). This would remove one more decision point the user needs to make. Then the type picker could list both edge types and concept types in the same list with top level items in the hierarchy of "Entites" and "Relationships" if you really want to limit to just entities you could select "Entities".

I'm currently working on a PR to add a third type "Events"/"Extended Data" and it would be nice to search all three types at the same time.

jharwig:
@joeferner the properties shown will be the union

jharwig:
And yes I think an element search should be default. With an action allowing the user to select entity/relation. But that's a different pr

core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepository.java
jharwig: Yup, I figured you'd want that when I was doing this. 🤓

I thought the signature looked nice with the previous, but I'll change it. Because we might add other options too

core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepositoryBase.java
jharwig: ok

PR #1022, opened 2017-03-02
Fix search related not showing type filter initially (3.1)

jharwig:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

also fix type change when searching related not updating search

Testing Instructions: Search related -> The Filter by type should appear, and changes to it should re-run the search

Points of Regression: None

CHANGELOG
Fixed: Changes to type filter when searching related will re-run the search


sfeng88:
@jharwig does this need a change log?

PR #1026, opened 2017-03-02
Element list improvements for listItemRenderer extensions (3.1)

jharwig:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

* Scope list styles to the default components so extensions don't have to
undo css styles
* Fix some multiple selection list visibility bugs
* Fleshed-out docs for item list renderer with less trivial example

Testing Instructions: The element list should work as before along with the listItemRenderer in doc-examples.

Points of Regression: Previous listItemRenderers might layout incorrectly as the styles have changed, but it wasn't documented/supported prior to 3.1

CHANGELOG
Fixed: Issue where multiple selection element list could disappear
Documentation: Added tutorial and documented [`org.visallo.entity.listItemRenderer`](http://docs.visallo.org/extension-points/front-end/entityListItemRenderer/)


sfeng88:
@jharwig should we add changelog items for the multiple selection list visibility bugs that was fixed? For the documentation updates, were they updates to documentation that was released in other versions?

PR #1032, opened 2017-03-03
Add task to more easily run link-check and force it to pass before pushing (3.1)

jharwig:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

screen shot 2017-03-03 at 1 11 22 pm

Previously you had to run the gitbook server and then `make link-check`. The new link-check task builds the book, starts a simple python web server, executes a link-check, then kills the server. This task is now required for pushing docs. Uses [node broken link-checker](https://github.com/stevenvachon/broken-link-checker).

Also fix some links that were broken.

Testing Instructions: None


docs/Makefile
```diff
@@ -97,5 +99,21 @@ push: build clone
...
+	@cd _book; \
+		python -m SimpleHTTPServer $(LINK_CHECK_PORT) > /dev/null & \
+		PID="$$!"; \
+		../node_modules/.bin/blc \
+			http://localhost:$(LINK_CHECK_PORT)/index.html \
+			--recursive \
+			--ordered \
+			--exclude-external \
+			--exclude "java/" \
+			--exclude "https://myvisalloinstance/" \
```
jharwig: These aren't needed when `--exclude-external` is enabled. But I left them in here because it's handy to check external links also. But that can show false errors when we're linking to new pages from jsdoc api so I disable it by default.

jharwig: actually, I'll just create another task for running external...

docs/getting-started/help.md
```diff
@@ -5,6 +5,6 @@ Having trouble? We'd like to help!
 * Read this developer guide. This guide is the most comprehensive documentation available and we're continually working to improve upon it.
 * Look for a `README.md` file in the root directory of a component for component-specific information. Keeping the documentation close to the code it describes makes it more likely to stay up to date.
 * Search for information in the archives of the [visallo mailing list](https://groups.google.com/d/forum/visallo). If you don't find what you're looking for, post a new message.
-* Visit the [Visallo blog](http://visallo.org/blog). We regularly post examples of using Visallo and announcing new features.
+* Visit the [Visallo blog](https://www.visallo.com/blog/). We regularly post examples of using Visallo and announcing new features.
```
EvanOxfeld: `announcing` should be `announce`.

PR #1038, opened 2017-03-06
History enhancements

diegogrz:
- [ ] @joeferner
- [x] @diegogrz
- [ ] @mwizeman @sfeng88

Requires Vertexium 2.5.4. 
History property value reporting now calculates the deltas between consecutive events server side (instead of reporting events). This alleviates the amount for processing visallo enterprise history plugin has to do. If no change occurs, the event is logged as a warning and not reported to the UI (usually due to a metadata changes). We can incorporate metadata changing reporting in the future. Visibility change reporting is optional. Added 'delete' event support.

Do not merge until 2.5.4 released and pom moved from SNAPSHOT to release version.
Requires: 
https://github.com/visallo/vertexium/pull/129
Related to: 
https://github.com/visallo/visallo-lts/issues/1035
https://github.com/visallo/visallo-lts/issues/1013

CHANGELOG
Changed: Modified history property value reporting to include delete events and avoid non-reportable events


core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -280,36 +283,105 @@ private static ClientApiProperty findProperty(
         return null;
     }
 
+
+    /**
+     * Sort HistoricalPropertyValue chronologically and generate events by looking at the
+     * deltas between the historical values.
+     */
     public static ClientApiHistoricalPropertyResults toClientApi(
```
joeferner: Can we split this method up a bit, it is very large.

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -280,36 +283,105 @@ private static ClientApiProperty findProperty(
...
+                if (value instanceof StreamingPropertyValue) {
+                    value = readStreamingPropertyValueForHistory((StreamingPropertyValue) value, locale, resourceBundle);
+                }
+                event.value = toClientApiValue(value);
+                if(withVisibility) {
+                    event.visibility = hpv.getPropertyVisibility().getVisibilityString();
+                }
+                propertyChange = true;
+                cachedValues.put(key, hpv);
+            }
```
joeferner: Formatting. Should be `} else if...` on one line. If you use IntelliJ the default formatting options should format this for you.

web/client-api/pom.xml
```diff
@@ -93,6 +94,11 @@
             <artifactId>jackson-databind</artifactId>
             <version>${jackson.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
```
joeferner: Doesn't look like this is being used. Can you remove this dependency. We've been trying to keep the dependencies in client API as light a possible to allow other to embed into other applications with minimal version conflicts.

web/client-api/src/main/java/org/visallo/web/clientapi/model/ClientApiHistoricalPropertyResults.java
```diff
@@ -1,19 +1,57 @@
 package org.visallo.web.clientapi.model;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import com.google.common.collect.Maps;
```
joeferner: Remove not used

root/pom.xml
```diff
@@ -77,7 +77,7 @@
         <lesscss.version>1.3.3</lesscss.version>
         <log4j.version>1.2.17</log4j.version>
         <metrics.version>3.0.2</metrics.version>
-        <vertexium.version>2.5.3</vertexium.version>
+        <vertexium.version>2.5.4-SNAPSHOT</vertexium.version>
```
mwizeman: We normally depend on non-snapshot versions of Vertexium since we only publish releases. This should be 2.5.4 and the PR should not merge until that version is released.

sfeng88:
@diegogrz this requires a release of vertexium before we can merge this PR. Please update the vertexium iterators as well in a separate PR.

diegogrz:
Changes will be merged to master as vertexium 2.5.4 will not make it to 3.1.

PR #1045, opened 2017-03-08
fix pushing hidden and deleted properties to the queue (3.1)

sfeng88:
- [x] @joeferner
- [x] @diegogrz
- [x] @mwizeman @sfeng88

Testing Instructions: None, covered by another PR

Points of Regression: None

CHANGELOG
Added: Pushing onto the gpw queue when a published property has been changed
Fixed: Added an additional validation of the status when the graph property runner checks for whether or not any GPWs are interested

### DIFF
--- a/core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
+++ b/core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
@@ -661,23 +661,21 @@ public class GraphPropertyRunner extends WorkerBase<GraphPropertyWorkerItem> {
         return this.shouldRun();
     }
 
-    public boolean canHandle(Element element, String propertyKey, String propertyName) {
+    public boolean canHandle(Element element, Property property, ElementOrPropertyStatus status) {
         if (!this.isStarted()) {
             //we are probably on a server and want to submit it to the architecture
             return true;
         }
 
-        Property property = element.getProperty(propertyKey, propertyName);
-
         for (GraphPropertyWorker worker : this.getAllGraphPropertyWorkers()) {
             try {
-                if (worker.isHandled(element, property)) {
+                if (status == ElementOrPropertyStatus.DELETION && worker.isDeleteHandled(element, property)) {
                     return true;
-                } else if (worker.isDeleteHandled(element, property)) {
+                } else if (status == ElementOrPropertyStatus.HIDDEN && worker.isHiddenHandled(element, property)) {
                     return true;
-                } else if (worker.isHiddenHandled(element, property)) {
+                } else if (status == ElementOrPropertyStatus.UNHIDDEN && worker.isUnhiddenHandled(element, property)) {
                     return true;
-                } else if (worker.isUnhiddenHandled(element, property)) {
+                } else if (worker.isHandled(element, property)) {
                     return true;
                 }
             } catch (Throwable t) {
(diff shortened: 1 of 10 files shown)


core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyWorkData.java
```diff
@@ -129,9 +129,17 @@ public VisibilityJson getVisibilityJson() {
         return getVisibilitySourceJson();
     }
 
+    public VisibilityJson getPropertyVisibilityJson() {
+        VisibilityJson propertyVisibilityJson = VisalloProperties.VISIBILITY_JSON_METADATA.getMetadataValue(property);
```
joeybrk372: property can be null

sfeng88: fixed

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
```diff
@@ -661,23 +661,21 @@ public boolean isStarted() {
         return this.shouldRun();
     }
 
-    public boolean canHandle(Element element, String propertyKey, String propertyName) {
+    public boolean canHandle(Element element, Property property, String propertyKey, String propertyName, ElementOrPropertyStatus status) {
```
joeferner: If we have property we shouldn't need `propertKey` and `propertyName`

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
```diff
@@ -699,6 +697,17 @@ public boolean canHandle(Element element, String propertyKey, String propertyNam
         return false;
     }
 
+    public boolean canHandle(Element element, String propertyKey, String propertyName, ElementOrPropertyStatus status) {
```
joeferner: Do we want to pass the property here too to be consistent?

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyWorkData.java
```diff
@@ -120,7 +120,7 @@ public VisibilityJson getVisibilitySourceJson() {
         return visibilityJson;
     }
 
-    public VisibilityJson getVisibilityJson() {
+    public VisibilityJson getElementVisibilityJson() {
```
joeferner: Can we keep the old `getVisibilityJson` and deprecate it.

core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
```diff
@@ -403,15 +439,26 @@ public void pushGraphPropertyQueue(
...
+            return true;
+        }
+        if (property == null) {
+            return false;
+        }
+
+        return this.graphPropertyRunner.canHandle(element, property, property.getKey(), property.getName(), status);
+    }
+
+    private boolean canHandle(final Element element, String propertyKey, final String propertyName, ElementOrPropertyStatus status) {
```
joeferner: remove `final` keywords

core/core/src/main/java/org/visallo/core/ingest/graphProperty/GraphPropertyRunner.java
rygim: This method actually calls the method above after inflating the property inside of it so it needs to stay the same.

joeferner:
Can you rebase/squash and merge

PR #1055, opened 2017-03-12
History enhancement master

diegogrz:
- [x] joeferner
- [x] diegogrz
- [x] mwizeman sfeng88

Notes:
Do not merge until 2.5.4 released
This copy from #1038 PR. Release 3.1 merge was pulled due to vertexium dependency.

Fixes:
visallo/visallo-lts#1035
visallo/visallo-lts#1013

CHANGELOG
Changed: Modified history property value reporting to include delete events and avoid non-reportable events


core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -24,13 +26,16 @@
...
 public class ClientApiConverter extends org.visallo.web.clientapi.util.ClientApiConverter {
+    private static final VisalloLogger LOGGER = VisalloLoggerFactory.getLogger(ClientApiConverter.class);
     public static final EnumSet<FetchHint> SEARCH_FETCH_HINTS = EnumSet.of(
             FetchHint.PROPERTIES,
             FetchHint.PROPERTY_METADATA,
             FetchHint.IN_EDGE_LABELS,
             FetchHint.OUT_EDGE_LABELS
     );
     private static final int HISTORICAL_PROPERTY_MAX_SPV_SIZE = 2000;
+    private static final String MODIFIED_BY_PROPERTY = "http://visallo.org#modifiedBy";
```
joeferner: You can get both of these from `VisalloProperties`

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -280,38 +285,170 @@ private static ClientApiProperty findProperty(
...
+        event.modifiedBy = ((modifiedByEntry != null) ? toClientApiValue(modifiedByEntry.getValue()).toString() : null);
+        return event;
+    }
+    
+    private static ClientApiHistoricalPropertyResults.Event generatePropertyAddedEvent(HistoricalPropertyValue hpv,
+        Locale locale, ResourceBundle resourceBundle, boolean withVisibility) {
+        
+        ClientApiHistoricalPropertyResults.Event event = generateGenericEvent(hpv); 
+        event.setEventType(ClientApiHistoricalPropertyResults.EventType.PROPERTY_ADDED);
+        Map<String, String> fields = Maps.newHashMap();
```
joeferner: Can we use `new HashMap<>()`. `Maps.newHashMap` in my opinion is from the good old days when you had to do `new HashMap<String, String>()` but with newer java versions `new HashMap<>()` seems just as good.

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -280,38 +285,170 @@ private static ClientApiProperty findProperty(
...
+        
+        Object value = hpv.getValue();
+        if (value instanceof StreamingPropertyValue) {
+            value = readStreamingPropertyValueForHistory((StreamingPropertyValue) value, locale, resourceBundle);
+        }
+        fields.put("value", toClientApiValue(value).toString());
+        
+        if(withVisibility) {
```
joeferner: Can you format this file. Should be `if (...)` with the space

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -280,38 +282,170 @@ private static ClientApiProperty findProperty(
...
+        Iterable<HistoricalPropertyValue> historicalPropertyValues, Locale locale, ResourceBundle resourceBundle,
+            boolean withVisibility) {
+       
         ClientApiHistoricalPropertyResults result = new ClientApiHistoricalPropertyResults();
-        for (HistoricalPropertyValue historicalPropertyValue : historicalPropertyValues) {
-            result.events.add(toClientApi(historicalPropertyValue, locale, resourceBundle));
+        
+        // Sort chronologically
+        List<HistoricalPropertyValue> sortedHistoricalValues = Lists.newArrayList(historicalPropertyValues);
+        Collections.reverse(sortedHistoricalValues);
```
mwizeman: This code seems to be assuming that they are already in sorted order, just backwards. Since the method is passed an Iterable, it may be working now but I don't think you can make that guarantee. The `HistoricalPropertyValue` class implements `Comparable` so can we run it through a sort instead of a simple reverse?

diegogrz: That is correct. I misread the documentation and thought "reverse" did a reverse sort instead of just a reversed. It just happen to work. Good catch. This may work Collections.sort(sortedHistoricalValues, Collections.reverseOrder()). Or simply do Collections.sort(sortedHistoricalValues); Collections.reverse(sortedHistoricalValues);

diegogrz: Change also need to be made to #1102

mwizeman: Updated both PR's

PR #1056, opened 2017-03-13
Update getting started guide to run compile before server

jharwig:
- [x] joeferner
- [x] diegogrz
- [x] mwizeman sfeng88

I've noticed that the run server command doesn't collect all the dependencies (notably webapp module) so the app won't be built without first running `mvn compile`

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -10,6 +10,10 @@ Change directories to the checked out code. This is your `$PROJECT_DIR` director
 
       cd visallo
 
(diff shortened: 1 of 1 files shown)


jharwig:
There's probably a way to make the command work, but I tried looking at the mvn dependency:tree when the `dev-tomcat-run` profile is active and the `web/war` is not included. I think it's due to the `-pl` option, but not sure how to fix it. Running `mvn compile` always works though...

PR #1058, opened 2017-03-14
Changed: cursor to pointer in decoration if an onClick handler is specified

joeferner:
- [x] joeferner
- [x] diegogrz
- [x] mwizeman sfeng88

To give the user a visual clue that the graph decoration is clickable
change the mouse cursor to a pointer

CHANGELOG
Changed: cursor to pointer in decoration if an onClick handler is specified

### DIFF
--- a/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
+++ b/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
@@ -306,6 +306,13 @@ define([
                      */
                     mouseout: 'onMouseOut'
                 }[event.type];
+                if (_.isFunction(decoration.onClick)) {
+                    if (handlerName === 'onMouseOver') {
+                        this.refs.node.style.cursor = 'pointer';
+                    } else if (handlerName === 'onMouseOut' || handlerName === 'onClick') {
+                        this.refs.node.style.cursor = null;
+                    }
+                }
                 if (_.isFunction(decoration[handlerName])) {
                     decoration[handlerName].call(cyTarget, event, {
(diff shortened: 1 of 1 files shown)


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -306,6 +306,13 @@ define([
                      */
                     mouseout: 'onMouseOut'
                 }[event.type];
+                if (_.isFunction(decoration['onClick'])) {
```
jharwig: `decoration.onClick`

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -306,6 +306,13 @@ define([
...
+                    if (handlerName === 'onMouseOver') {
+                        this.refs.node.style.cursor = 'pointer';
+                    } else if (handlerName === 'onMouseOut') {
```
jharwig: should a click also reset it? In case cytoscape never sends it

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
joeferner: done

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
joeferner: done

PR #1059, opened 2017-03-14
Fixed: Tomcat exceptions in Atmosphere when session expires

joeferner:
- [x] joeferner
- [x] diegogrz
- [x] mwizeman sfeng88

Tomcat deletes the session before Atmosphere's disconnect gets called
causing the user's id to not be found resulting in an exception thrown.
The user's session count is depricated by web/web-base/src/main/java/org/visallo/web/SessionUser.java
valueUnbound

Testing:
Login as a user and let the session expire. This should not throw an exception and exceptions should not show up in the log.

CHANGELOG
Fixed: Tomcat exceptions in Atmosphere when session expires

### DIFF
--- a/web/war/src/main/webapp/WEB-INF/web.xml
+++ b/web/war/src/main/webapp/WEB-INF/web.xml
@@ -6,4 +6,10 @@
         <listener-class>org.visallo.web.ApplicationBootstrap</listener-class>
     </listener>
     <distributable/>
+
+    <!--
+    <session-config>
+        <session-timeout>1</session-timeout>
+    </session-config>
+    -->
 </web-app>
--- a/web/web-base/src/main/java/org/visallo/web/ApplicationBootstrap.java
+++ b/web/web-base/src/main/java/org/visallo/web/ApplicationBootstrap.java
@@ -2,10 +2,7 @@ package org.visallo.web;
 
 import com.google.inject.Injector;
 import org.atmosphere.cache.UUIDBroadcasterCache;
-import org.atmosphere.cpr.AtmosphereHandler;
-import org.atmosphere.cpr.AtmosphereInterceptor;
-import org.atmosphere.cpr.AtmosphereServlet;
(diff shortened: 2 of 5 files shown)


web/web-base/src/main/java/org/visallo/web/ApplicationBootstrap.java
```diff
@@ -160,16 +157,17 @@ private void addAtmosphereServlet(ServletContext context, Configuration config)
         servlet.setAsyncSupported(true);
         servlet.setLoadOnStartup(0);
         servlet.setInitParameter(AtmosphereHandler.class.getName(), Messaging.class.getName());
-        servlet.setInitParameter("org.atmosphere.cpr.sessionSupport", "true");
-        servlet.setInitParameter("org.atmosphere.cpr.broadcastFilterClasses", MessagingFilter.class.getName() + "," +
+        servlet.setInitParameter(ApplicationConfig.PROPERTY_SESSION_CREATE, "false");
```
joeferner: Added this so that Atmosphere doesn't try to create empty sessions.

web/war/src/main/webapp/WEB-INF/web.xml
```diff
@@ -6,4 +6,10 @@
         <listener-class>org.visallo.web.ApplicationBootstrap</listener-class>
     </listener>
     <distributable/>
+
+    <!--
```
joeferner: Kept this in to make it easier to test by uncommenting these lines

EvanOxfeld:
@joeferner Is the issue logging only or do the Tomcat exceptions cause incorrect behavior?

joeferner:
@EvanOxfeld This PR fixes the logging issue. Trying to get rid of all the ERROR logging so I can actually monitor the logs.

EvanOxfeld:
Sounds like a good plan. 👍  Just making sure this change isn't needed in the 3.1 branch.

PR #1068, opened 2017-03-16
Added: configuration to set web session expiration

joeferner:
- [x] joeferner
- [x] diegogrz
- [x] mwizeman sfeng88

Java web containers typically require the configuration of web session
expiration using the web.xml file. Since Visallo ships a pre-built war
file this commit allows software to configure the web session timeout
without repackaging the war file.

CHANGELOG
Added: configuration to set web session expiration

### DIFF
--- a/web/war/src/main/webapp/WEB-INF/web.xml
+++ b/web/war/src/main/webapp/WEB-INF/web.xml
@@ -6,10 +6,7 @@
         <listener-class>org.visallo.web.ApplicationBootstrap</listener-class>
     </listener>
     <distributable/>
-
-    <!--
-    <session-config>
-        <session-timeout>1</session-timeout>
-    </session-config>
-    -->
+    <listener>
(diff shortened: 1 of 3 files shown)


core/core/src/main/java/org/visallo/core/config/Configuration.java
```diff
@@ -91,6 +91,8 @@
     public static final String STATUS_ENABLED = "status.enabled";
     public static final boolean STATUS_ENABLED_DEFAULT = true;
     public static final String SYSTEM_PROPERTY_PREFIX = "visallo.";
+    public static final String WEB_MAX_SESSION_INACTIVE_INTERVAL_SECONDS = "web.maxSessionInactiveIntervalSeconds";
```
jharwig: Should we just put these in WebConfiguration and give it the `web.ui` prefix? Then the UI could have access to that number to show users in the session timeout error message.

joeferner: Moved

PR #1075, opened 2017-03-16
Link check docs 3.1

jharwig:
- [ ] joeferner
- [x] diegogrz
- [ ] mwizeman sfeng88

Runs server and external link checker before deploying docs

### DIFF
--- a/travis/script.sh
+++ b/travis/script.sh
@@ -10,7 +10,7 @@ if [ "$BUILD_DOCS" ]; then
     # Check if this branch should be built as part of documentation
     if [[ $VERSION_LIST =~ (^|[[:space:]])$TRAVIS_BRANCH($|[[:space:]]) ]]; then
       source ~/.nvm/nvm.sh
-      VERSION_CURRENT="$TRAVIS_BRANCH" make -C docs build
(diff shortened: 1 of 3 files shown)


travis/install.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 if [ "$BUILD_DOCS" ]; then
-  if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
+  #if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
```
joeferner: Should this be commented out?

travis/script.sh
```diff
@@ -5,18 +5,18 @@ set -e
 if [ "$BUILD_DOCS" ]; then
 
   # only build if merging into real branch
-  if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
+  #if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
```
joeferner: How about this?

travis/install.sh
jharwig: yup, I disabled that check to test the process. It worked, so they're back

PR #1081, opened 2017-03-22
Web: Create workspace edge when adding vertex (3.1)

EvanOxfeld:
- [x] joeferner
- [ ] diegogrz
- [x] mwizeman sfeng88

Adds both workspace and work product edges when adding vertices to a product.

Testing Instructions: Verify that newly created relationships are displayed in the diff panel

CHANGELOG
Fixed: Creates workspace edge when adding vertex

### DIFF
--- a/web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexNew.java
+++ b/web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexNew.java
@@ -13,7 +13,6 @@ import org.visallo.core.model.ontology.OntologyRepository;
 import org.visallo.core.model.properties.VisalloProperties;
 import org.visallo.core.model.workQueue.Priority;
 import org.visallo.core.model.workQueue.WorkQueueRepository;
-import org.visallo.core.model.workspace.Workspace;
 import org.visallo.core.model.workspace.WorkspaceHelper;
 import org.visallo.core.model.workspace.WorkspaceRepository;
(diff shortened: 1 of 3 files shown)


web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexNew.java
```diff
@@ -135,10 +137,9 @@ public ClientApiElement handle(
         );
 
         if (workspaceId != null) {
-            Workspace workspace = workspaceRepository.findById(workspaceId, user);
-            workspaceRepository.updateEntityOnWorkspace(workspace, vertex.getId(), user);
-            workQueueRepository.pushUserCurrentWorkspaceChange(user, workspaceId);
-            this.graph.flush();
+            Collection<String> vertexIds = new ArrayList<>();
+            vertexIds.add(vertex.getId());
+            workspaceHelper.updateEntitiesOnWorkspace(workspaceId, vertexIds, user);
```
rygim: Just a nitpick, so feel free to merge it without changing, but it is a line or two less of code to use [Arrays.asList()](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#asList-T...-) or [Collections.singletonList()](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#singletonList-T-) to name a few helper methods to turn values into lists.

PR #1084, opened 2017-03-22
Added: collapsing of graph nodes into single nodes

joeferner:
- [x] joeferner
- [x] diegogrz
- [x] mwizeman sfeng88

This commit only effects the current work product. It collapses multiple
nodes into a single node keeping the edges.

[Video](https://drive.google.com/open?id=0BwAYHbIO38hXSWpoMWdtWHBxX2c)

CHANGELOG
Added: collapsing of graph nodes into single nodes

### DIFF
--- a/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
+++ b/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
@@ -6,6 +6,7 @@ define([
     './GraphEmpty',
     './GraphExtensionViews',
     './popovers/index',
+    './collapsedNodeImageHelpers',
     'util/vertex/formatters',
     'util/retina',
     'components/RegistryInjectorHOC'
@@ -17,6 +18,7 @@ define([
     GraphEmpty,
     GraphExtensionViews,
     Popovers,
+    CollapsedNodeImageHelpers,
     F,
     retina,
     RegistryInjectorHOC) {
@@ -29,14 +31,16 @@ define([
     const noop = function() {};
     const generateCompoundEdgeId = edge => edge.outVertexId + edge.inVertexId + edge.label;
     const isGhost = cyElement => cyElement && cyElement._private && cyElement._private.data && cyElement._private.data.animateTo;
-    const isValidElement = cyElement => cyElement && cyElement.is('.v,.e,.partial') && !isGhost(cyElement);
-    const isValidNode = cyElement => cyElement && cyElement.is('node.v,node.partial') && !isGhost(cyElement);
+    const isValidElement = cyElement => cyElement && cyElement.is('.c,.v,.e,.partial') && !isGhost(cyElement);
+    const isValidNode = cyElement => cyElement && cyElement.is('node.c,node.v,node.partial') && !isGhost(cyElement);
     const edgeDisplay = (label, ontology, edges) => {
         const display = label in ontology.relationships ? ontology.relationships[label].displayName : '';
         const showNum = edges.length > 1;
         const num = showNum ? ` (${F.number.pretty(edges.length)})` : '';
         return display + num;
     };
+    const COLLAPSED_EXTENDED_DATA_KEY = 'org.visallo.web.product.graph.collapsed';
+    const COLLAPSED_ID_PREFIX = '__COLLAPSED_';
     const Graph = React.createClass({
 
         propTypes: {
@@ -59,7 +63,8 @@ define([
                 initialProductDisplay: true,
                 draw: null,
                 paths: null,
-                hovering: null
+                hovering: null,
+                collapsedImageDataUris: {}
             }
         },
 
@@ -98,6 +103,9 @@ define([
                             connectionType
                         }
                     });
+                },
+                uncollapse: (event, { collapsedItemId }) => {
+                    this.uncollapse(collapsedItemId);
                 },
                 menubarToggleDisplay: { node: document, handler: (event, data) => {
                     if (data.name === 'products-full') {
@@ -185,7 +193,8 @@ define([
                 menuHandlers = {
                     onMenuCreateVertex: this.onMenuCreateVertex,
                     onMenuSelect: this.onMenuSelect,
-                    onMenuExport: this.onMenuExport
+                    onMenuExport: this.onMenuExport,
+                    onCollapseSelectedVertices: this.onCollapseSelectedVertices
                 },
                 cyElements = this.mapPropsToElements(editable),
                 extensionViews = registry['org.visallo.graph.view'];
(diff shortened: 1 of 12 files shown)


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -826,16 +1002,184 @@ define([
                 this.removeEvents.push({ node, func, events });
                 $(node).on(events, func);
             })
+        },
+
+        updateImageDataUrisForCollapsedItems() {
```
jharwig: I think all these image get functions should be in a collapsed utility file

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -826,16 +1002,184 @@ define([
...
+                });
+            });
+            if (changed) {
+                this.setState({
+                    collapsedImageDataUris: newCollapsedImageDataUris
+                });
+            }
+        },
+
+        generateImageDataUriForCollapsedItem(collapsedItemId, collapsedItem) {
```
jharwig: Am I right in that this is only called once? unless the vertices inside change? What about vertex icons changing? 

Just worried that this seems very slow

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -826,16 +1002,184 @@ define([
...
+            const circleLineWidth = 5;
+            const circleColor = '#a5a5a5';
+            const circleRadius = (imageSize / 2) - circlePaddingTop;
+            const countTextFont = '50px Arial';
+            const countTextBorderRadius = 22;
+            const countTextBackgroundStyle = '#a5a5a5';
+            const countTextForegroundStyle = '#ffffff';
+
+            Promise.all(collapsedItem.vertexIds.map(vertexId => this.loadVertexImage(vertexId)))
+                .then(images => {
```
jharwig: should you unique the images?

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -826,16 +1002,184 @@ define([
...
+                    const canvas = document.createElement('canvas');
+                    canvas.height = canvas.width = imageSize;
+                    const canvasCtx = canvas.getContext('2d');
+
+                    // draw circle
+                    canvasCtx.beginPath();
```
jharwig: this function is very large, what about drawCircle(ctx), drawImages(ctx), drawCount(ctx) functions...

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -826,16 +1002,184 @@ define([
...
+                                layoutInfo.size,
+                                layoutInfo.size
+                            );
+                        }
+                    });
+
+                    // draw count bubble
+                    canvasCtx.font = countTextFont;
+                    const countText = collapsedItem.vertexIds.length.toString();
+                    const countTextMeasurements = canvasCtx.measureText(countText);
```
jharwig: probably want to cache/memoize the calculation of this measurement. `measureText` is kinda slow

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -826,16 +1002,184 @@ define([
...
+                });
+        },
+
+        loadVertexImage(vertexId) {
+            return new Promise((resolve, reject) => {
+                const vertex = this.props.elements.vertices[vertexId];
+                if (!vertex) {
+                    return reject(new Error(`Could not find vertex: ${vertexId}`));
+                }
+                const vertexImageUrl = F.vertex.selectedImage(vertex, null, 150);
```
jharwig: why selected image instead of `image`?

jharwig: also 150 feels kinda large compared to the collated node size. How big are they rendered? x2 that number here

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/GraphContainer.jsx
```diff
@@ -236,6 +236,19 @@ define([
...
+     * Register a function that can add or remove classes from cytoscape collapsed nodes for custom styling.
+     *
+     * @param {org.visallo.graph.collapsed.class~classFn} config
+     */
+    registry.documentExtensionPoint('org.visallo.graph.collapsed.class',
+        'Function that can change cytoscape classes of nodes',
+        function(e) {
+            return _.isFunction(e);
+        },
+        'http://docs.visallo.org/extension-points/front-end/graphCollapsedNode/class.html'
```
jharwig: Need to actually create `docs/extension-points/front-end/graphCollapsedNode/class.md`

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
joeferner: done

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
joeferner: I don't think so. The way it works if there are lets say 4 airplanes grouped together it will show 4 airplane in the circle

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
joeferner: done

EvanOxfeld:
Conceptually I like this idea -- not only is it a good way to help an investigator make sense of a large graph, but also it potentially could help with large graph performance.

I'd like to try out this feature. I checked out the branch, but I don't see the collapse option. Is there a setting required to enable collapsing entities? Also, what happens if you collapse entities of different concept types?

joeferner:
@EvanOxfeld did you webpack graph and web/war? Did you right click in open space or on an entity (btw, I think this behavior is strange in general. If I have multiple entities selected and I right click an entity I should probably get the multi-select context menu)?

Collapsing different concept types works well, you get different little icons in the collapse bubble.

EvanOxfeld:
Thanks, collapsing looks good. ✨  My issue was right clicking on one of the selected entities instead of open space. I believe fixing the multi-select context menu behavior is covered by #2353.

One issue I found was that when adding additional entities to a collapsed node, the histogram doesn't appear to update correctly:

collapsed-histogram-wrong-count

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
joeferner: done

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/GraphContainer.jsx
joeferner: Oops, do we want a full example here? It will be nearly equivalent to node.class so I refer them to those docs.

EvanOxfeld:
@joeferner Are you still fixing the histogram issue?

joeferner:
@EvanOxfeld Yes, once I find some time to do it

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
joeferner: this only gets called if the entity image URLs change.

PR #1095, opened 2017-04-03
Web: Secure work product edges (3.1.1)

EvanOxfeld:
- [x] joeferner
- [x] diegogrz
- [x] mwizeman sfeng88

Updates ProductGet endpoint to return a lists of authorized edges and unauthorized edge IDs.

Points of Regression: Possibly edges moving between the sandbox and public or collaborating on edges with users with different authorizations.

CHANGELOG
Fixed: Brief display of some edges breaking the case sandbox when switching graphs

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java
+++ b/core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java
@@ -136,6 +136,7 @@ public abstract class WorkProductElements implements WorkProduct, WorkProductHas
 
         if (params.optBoolean("includeEdges")) {
             JSONArray edges = new JSONArray();
+            JSONArray unauthorizedEdgeIds = new JSONArray();
             Authorizations systemAuthorizations = authorizationRepository.getGraphAuthorizations(
                     user,
                     VisalloVisibility.SUPER_USER_VISIBILITY_STRING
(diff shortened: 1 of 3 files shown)


core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductElements.java
```diff
@@ -145,16 +146,28 @@ public JSONObject getExtendedData(
...
-                edge.put("edgeId", relatedEdge.getEdgeId());
-                edge.put("label", relatedEdge.getLabel());
-                edge.put("outVertexId", relatedEdge.getOutVertexId());
-                edge.put("inVertexId", relatedEdge.getInVertexId());
-                edges.put(edge);
+            Iterable<RelatedEdge> productRelatedEdges = graph.findRelatedEdgeSummaryForVertices(productVertices, authorizations);
+            List<String> ids = StreamUtil.stream(productRelatedEdges)
+                    .map(edge -> edge.getEdgeId())
+                    .collect(Collectors.toList());
+            Map<String, Edge> relatedEdgesById = StreamUtil.stream(graph.getEdges(ids, FetchHint.NONE, authorizations))
```
joeferner: Can use `graph.doEdgesExist` instead. https://github.com/visallo/vertexium/blob/master/core/src/main/java/org/vertexium/Graph.java#L585

PR #1112, opened 2017-04-13
Fix history "Created Thing"(3.1.2)

joeybrk372:
- [x] joeferner
- [x] diegogrz
- [ ] mwizeman sfeng88

We are iterating through chronological history events and in some cases the concept type may be set before the modifiedBy property or vice versa. Now this will find the modifiedBy property if we haven't already viewed it.

Points of Regression: Creation history for any concept/edge.

CHANGELOG
Fixed: Importing files and viewing their history would show "Created Thing" instead of "Created Image/Document/etc."


core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -308,16 +308,25 @@ public static ClientApiHistoricalPropertyResults calculateHistoricalPropertyDelt
             if (cached == null) { // Add
                 if (hpv.getPropertyName().equals(VisalloProperties.CONCEPT_TYPE.getPropertyName())) {
                     conceptTypeHpv = hpv;
+                    HistoricalPropertyValue modifiedByHpv = cachedValues.get(VisalloProperties.MODIFIED_BY.getPropertyName());
+
+                    int i = 0;
```
diegogrz: You should only need to look for the ModifiedBy property from this point forward since it wasnt cached. Consider keeping the index "i" in the outer loop and use it to look forward in sortedHistoricalValues

core/core/src/main/java/org/visallo/core/util/ClientApiConverter.java
```diff
@@ -308,16 +308,25 @@ public static ClientApiHistoricalPropertyResults calculateHistoricalPropertyDelt
...
+                    while (modifiedByHpv == null) {
+                        HistoricalPropertyValue v = sortedHistoricalValues.get(i);
+                        if (v.getPropertyName().equals(VisalloProperties.MODIFIED_BY.getPropertyName())) {
+                            modifiedByHpv = v;
+                        }
+                        i++;
+                    }
+
+                    // Use the ModifiedBy property to complete the ConceptType event
+                    ClientApiHistoricalPropertyResults.Event modifiedByEvent = generatePropertyAddedEvent(modifiedByHpv, locale, resourceBundle, withVisibility);
```
diegogrz: Lets add a check in case modifiedByHpv was not found

PR #1117, opened 2017-04-25
Allow generated code to add visibilities (master)

rygim:
It is difficult with the current api to set different visibilities for data that is ingested. These modifications alter the api to allow entities to create properties with different visibilities.

- [x] joeferner
- [x] diegogrz
- [x] mwizeman sfeng88

Testing Instructions: Use the code generator to generate ontology classes

Points of Regression: None

CHANGELOG
Changed: Can set visibility to ontology generated class properties


EvanOxfeld:
Are these APIs included in the codgen documentation? Would you please add some documentation if not?

PR #1120, opened 2017-04-28
Selection fixes

jharwig:
- [x] joeferner
- [x] diegogrz
- [ ] mwizeman sfeng88

* Removes duplicate `objectsSelected` event from firing
  * Fixes Timeline selection window from disappearing, now also highlights the selected in blue
  * Fixes element inspector double render
* Set the selection `options` object in redux
* Move `focus` selectObjects parameter to options. Includes console
warning if it notices old plugin code that might use it.

Testing Instructions: 
Selection of objects should work correctly from search, work products, multiple selection.
Also create a property source info reference to a document and click the reference link –> the selected text that was referenced should flash and the text body should scroll to it.

CHANGELOG
Fixed: Element inspector "double-render" when selecting from search results.
Fixes: Timeline selection window disappearing after selecting


jharwig:
TODO: create cherry-pick to 3.1.3

PR #1127, opened 2017-05-05
Fix image selection box and resolution(3.1.3)

joeybrk372:
- [x] joeferner
- [x] diegogrz
- [x] mwizeman sfeng88

We flash the product menu bar icon when dragging data, which was failing when trying to drag the image selection box.

Also fixed an issue with the resolved edge creation and the detected objects form getting torn down every time you moved/resized the selection box. Now the form will only update if the object is changed.

Testing Instructions: Resolve/unresolve detected objects and your own selections. You should be able to switch between detected objects and resolving new selections.

CHANGELOG
Fixed: Image selection box could not be moved if the Work Products List was closed.
Fixed: Image object resolution form no longer reopens every time you resize/move the selection box.
Fixed: Resolving object would sometimes fail.

### DIFF
--- a/web/web-base/src/main/java/org/visallo/web/routes/vertex/ResolveDetectedObject.java
+++ b/web/web-base/src/main/java/org/visallo/web/routes/vertex/ResolveDetectedObject.java
@@ -104,13 +104,15 @@ public class ResolveDetectedObject implements ParameterizedHandler {
         Visibility defaultVisibility = visibilityTranslator.getDefaultVisibility();
         VisalloProperties.VISIBILITY_JSON_METADATA.setMetadata(metadata, visibilityJson, defaultVisibility);
 
+        Date modifiedDate = new Date();
+
         Vertex resolvedVertex;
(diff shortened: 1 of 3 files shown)


web/web-base/src/main/java/org/visallo/web/routes/vertex/ResolveDetectedObject.java
```diff
@@ -134,6 +134,7 @@ public ClientApiElement handle(
         }
 
         Edge edge = graph.addEdge(artifactVertex, resolvedVertex, artifactContainsImageOfEntityIri, visalloVisibility.getVisibility(), authorizations);
+        graph.flush();
```
joeferner: This is a strange place for a flush. Did you mean for this to be here?

joeybrk372: Yes, the edge was not available when trying to set the visibilityJson property. But updated to use an edge mutation because it was also missing other metadata properties.

PR #1134, opened 2017-05-06
Changed: OWL Import: Compare the previous ontology MD5 and skip import if it has not changed

joeferner:
- [x] joeferner
- [ ] diegogrz
- [x] mwizeman sfeng88

Importing OWL files creates lots of history and takes time to prevent this store
the MD5 and only re-import it when the MD5 has changed

Testing Instructions:
- Start the application with an OWL file defined in the configuration
- Restart the application, the OWL file should be skipped
- Change the OWL file
- Restart the application, the OWL file should import the new changes

CHANGELOG
Changed: OWL Import: Compare the previous ontology MD5 and skip import if it has not changed

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/ontology/OntologyProperties.java
+++ b/core/core/src/main/java/org/visallo/core/model/ontology/OntologyProperties.java
@@ -10,6 +10,7 @@ public class OntologyProperties {
 
     public static final StringSingleValueVisalloProperty TITLE = new StringSingleValueVisalloProperty("http://visallo.org#title");
     public static final StreamingVisalloProperty ONTOLOGY_FILE = new StreamingVisalloProperty("http://visallo.org#ontologyFile");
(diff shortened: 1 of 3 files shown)


core/plugins/model-vertexium-inmemory/src/main/java/org/visallo/vertexium/model/ontology/InMemoryOntologyRepository.java
```diff
@@ -109,6 +109,11 @@ protected void addEntityGlyphIconToEntityConcept(Concept entityConcept, byte[] r
     }
 
     @Override
+    protected boolean hasFileChanged(IRI documentIRI, byte[] inFileData) {
```
mwizeman: Since this is the same implementation as the base class, does it need to define it here?

joeferner: removed

PR #1135, opened 2017-05-08
Remove duplicate setting of Json in VertexiumLongRunningProcessRepository

joeferner:
- [x] joeferner
- [ ] diegogrz
- [x] mwizeman sfeng88

Testing Instructions:
- Submit a long running task such as find path.
- Should work and return results as before.

No Changelog needed


jharwig:
Yeah, I probably messed that because of a merge conflict

PR #1140, opened 2017-05-10
Graph extension point caching

jharwig:
- [x] joeferner
- [ ] diegogrz
- [x] mwizeman sfeng88

For the following extension points don't call the functions until the vertex or registry changes:

* org.visallo.graph.edge.class
* org.visallo.graph.edge.transformer
* org.visallo.graph.node.class
* org.visallo.graph.node.transformer
* org.visallo.graph.node.decoration (applyTo and data)

Previously some state changes (like a selection or opening a panel) would cause all these extension point functions to re-evaluate.

Needed to change the registry to be non-mutated state for simple comparisons, and added some tests to both graph and registry.

Added benefit is new the `RegistryInjectorHOC` now won't trigger a
re-render unless the subscribed to registry changes

Testing Instructions: Register extensions of the above types. Use doc-examples repo for samples.

Points of Regression: None

CHANGELOG
Fixed: Certain graph extension points would get invoked more often than necessary causing slowdowns with large graphs.

### DIFF
--- a/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Cytoscape.jsx
+++ b/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Cytoscape.jsx
@@ -227,7 +227,7 @@ define([
                 if (initialProductDisplay && !hasViewport) return true;
                 const wasEmpty = oldNodes.length === 0;
                 const hasNodes = newNodes.length
-                const positionIsEmpty = node => node.position.x === 0 && node.position.y === 0;
+                const positionIsEmpty = node => !node.position || (node.position.x === 0 && node.position.y === 0);
 
                 if (wasEmpty && hasNodes) {
                     return _.every(newNodes, positionIsEmpty);
--- a/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
+++ b/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
@@ -37,9 +37,51 @@ define([
         const num = showNum ? ` (${F.number.pretty(edges.length)})` : '';
         return display + num;
     };
+    const propTypesElementArrays = { vertices: PropTypes.array, edges: PropTypes.array };
+
+    let memoizeForStorage = {};
+    const memoizeClear = () => { memoizeForStorage = {}; }
+    const memoizeFor = function(key, elements, fn, idFn) {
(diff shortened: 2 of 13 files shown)


joeferner:
After removing my caching, this PR reduced the times from ~22s to ~13s, good work

jharwig:
Testing snippets to add these extensions and log when they are called. Previously any zooming, opening panes would trigger them.

```js
require(['public/v1/api'], visallo => {
    var registry = visallo.registry;
    if ('__reg' in window) __reg.forEach(e => registry.unregisterExtension(e));    
    var reg = window.__reg = [];

    
    
    reg.push(registry.registerExtension('org.visallo.graph.style', function(cytoscapeStylesheet) {
        console.log('invoking stylesheet (should invoke when selection changes)');
        cytoscapeStylesheet.selector('edge.custom_cls_e')
             .style({
                 'width': '5', 'line-color': 'red'
             }) 
        cytoscapeStylesheet.selector('node.custom_cls')
             .style({
                 'background-color': '5', 'line-color': 'red'
             })                         
    }))

    reg.push(registry.registerExtension('org.visallo.graph.edge.transformer', function(d) {
        console.log('invoking transformer (should exec when edge(s) changes)', data.edges);        
    }))

    reg.push(registry.registerExtension('org.visallo.graph.node.transformer', function(v) {
        console.log('invoking transformer (should exec when vertex changes)', v.id);        
    }))

    reg.push(registry.registerExtension('org.visallo.graph.node.class', function(v, c) {
        console.log('invoking node classer (should exec when vertex changes)', v.id);        
        c.push('custom_cls');
    }))  

    reg.push(registry.registerExtension('org.visallo.graph.edge.class', function(edges, type, c) {
        console.log('invoking edge classer (should exec when edge(s) changes)', edges);        
        c.push('custom_cls_e');
    }))   
       
    reg.push(registry.registerExtension('org.visallo.graph.node.decoration', {
        applyTo: function(v) {
            console.log('invoking applyTo (should exec when vertex changes)', v.id);        
            return _.any(v.properties, function(p) {
                return p.name === 'http://visallo.org/comment#entry' &&
                    p.value.indexOf(onlyDecorationAlignmentsText) === -1 &&
                    p.value.indexOf("popover") === -1;
            })
        },
        alignment: { h: 'left', v: 'top' },        
        data: function(v) {
            console.log('invoking data (should exec when vertex changes)', v.id);        
            return {
                label: '!'
            }
        }
    }))
})
```

PR #1143, opened 2017-05-11
High side merge

diegogrz:
CHANGELOG
Changed: 
* Tomcat configuration now uses a generic error page when an exception occurs
* Log email notification as INFO
* Allow search queries to be exact matches


jharwig:
That changelog syntax won't work, please use.

```
CHANGELOG
Changed: change1
Changed: change2
etc...
```

PR #1159, opened 2017-05-18
Convert as many ejs files to handlebars

jharwig:
- [x] joeferner
- [ ] diegogrz
- [x] mwizeman sfeng88

Should speed up load times, especially on slower CPUs. Left some files like alert.ejs which is too crazy for handlebars.

Testing Instructions: Things that changed to test: audio, video entities, workspaces, elements lists, workspace user sharing

Points of Regression: None

CHANGELOG
Changed: Converted legacy `ejs` template files to `hbs` which are precompiled at build time.


web/war/src/main/webapp/js/app.hbs
```diff
@@ -31,9 +31,3 @@
 <div class="workspace-overlay"/>
 <div class="workspace-timeline"/>
 <div class="help-dialog modal hide"/>
-
-<div style="display:none" 
-     class="alert alert-warning context-menu-warning">
-<button type="button" class="close" data-dismiss="alert">×</button>
-<%= i18n('contextmenu.blocked') %>
-</div>
```
jharwig: This is no longer used

web/war/src/main/webapp/js/app.js
```diff
@@ -51,13 +51,6 @@ define([
 
     return defineComponent(App, withFileDrop, withDataRequest);
 
-    function preventPinchToZoom(e) {
-        if (e.ctrlKey) {
-            e.preventDefault();
-            e.stopPropagation();
-        }
-    }
-
```
jharwig: This was preventing a chrome issue with zooming that doesn't seem to be a problem anymore

PR #1161, opened 2017-05-19
Automatically compile plugin template files (hbs) on first request

jharwig:
- [x] joeferner
- [ ] mwizeman sfeng88

Testing Instructions: All Plugins?

Points of Regression: Will warn when using `hbs!` syntax (but will work)

CHANGELOG
Changed: Plugin handlebar templates are precompiled on web server startup.
Deprecated: `hbs!` RequireJs syntax is deprecated for plugins. All templates registered with `registerJavascriptTemplate` should change require path from `hbs![path]` to `[path].hbs`. A warning is displayed, but automatically switched to compiled file.


web/web-base/src/main/java/org/visallo/web/HandlebarsPrecompiledResourceHandler.java
```diff
@@ -0,0 +1,98 @@
...
+
+public class HandlebarsPrecompiledResourceHandler implements RequestResponseHandler {
+    private static final VisalloLogger LOGGER = VisalloLoggerFactory.getLogger(HandlebarsPrecompiledResourceHandler.class);
+
+    private final String path;
+    private String cached;
+    private Long cachedLastModified;
+    private boolean recompileOnModification;
+
+    public HandlebarsPrecompiledResourceHandler(String path) {
```
joeferner: This constructor appears to not be used

PR #1162, opened 2017-05-19
Fix panes getting stuck at zero width(3.1.3)

joeybrk372:
- [x] joeferner
- [x] mwizeman sfeng88

Testing Instructions: Follow instructions in issue, also when drawing an entity resolution box in an image(larger is better)

Points of Regression: panel resizing, image resolution

CHANGELOG
Fixed: Issue where side panels would sometimes become invisible when trying to resize them.


joeferner:
@joeybrk372 I added this bit of code onsite https://github.com/visallo/visallo/commit/d2672f573c28cb6d1d1b95957c760515a28501f8 to fix the user that had broken values. Should we add this commit to this PR?

jharwig:
@joeferner  Yeah, that's a good sanity ✅  although clamping can be a one-liner: `$pane.width(Math.min(maxWidth, Math.max(minWidth, width));`

joeferner:
I don't think minWidth and maxWidth are always specified though.

PR #1164, opened 2017-05-19
Changed: keep search results when reopening the search pane

joeferner:
- [x] joeferner
- [ ] mwizeman sfeng88

Previously a user would need to rerun the search, on complex searches
this could take a while. This change allows the user to close the search
pane and reopen and retain their search results.

Testing Instructions:
- Open the search and find something
- Close the search pane using the "ESC" key (or click the find button again)
- Reopen the search pane, the results should be there

CHANGELOG
Changed: keep search results when reopening the search pane

### DIFF
--- a/web/war/src/main/webapp/js/search/types/withSearch.js
+++ b/web/war/src/main/webapp/js/search/types/withSearch.js
@@ -69,7 +69,7 @@ define([
                 }
             });
             this.on('clearSearch', function(event, data) {
-                this.hideSearchResults();
+                this.hideAndClearSearchResults();
 
                 var filters = this.select('filtersSelector').find('.content')
                 this.trigger(filters, 'clearfilters', data);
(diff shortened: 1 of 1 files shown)


web/war/src/main/webapp/js/search/types/withSearch.js
```diff
@@ -69,7 +69,7 @@ define([
                 }
             });
             this.on('clearSearch', function(event, data) {
-                this.hideSearchResults();
+                this.hideSearchResults(true);
```
EvanOxfeld: Rather than include a boolean argument, I find it more maintainable to split out a second method to clear out the results.

joeferner: @EvanOxfeld how about this?

EvanOxfeld: Better. While not worth holding up this PR, by name `hideAndClearSearchResults` does two things...I find separate `clearSearchResults` and `hideSearchResults` methods more maintainable. The `clearSearch` handler would call both, everywhere else would call just `hideSearchResults`.

PR #1171, opened 2017-05-22
Added: support for react components in advanced search

joeferner:
- [x] joeferner
- [x] mwizeman sfeng88

CHANGELOG
Added: Support for react components in advanced search. `org.visallo.search.advanced` extension point now provides `renderResults`, `updateQueryStatus`, and `initialParameters`
Deprecated: `resultsSelector` parameter for the `org.visallo.search.advanced` extension point . Use `renderResults` instead.

### DIFF
--- a/web/war/src/main/webapp/js/search/search.js
+++ b/web/war/src/main/webapp/js/search/search.js
@@ -105,7 +105,8 @@ define([
             return {
                 type: type,
                 query: '',
-                filters: []
+                filters: [],
+                status: {}
             }
         }), 'type');
 
@@ -172,19 +173,27 @@ define([
          * @event org.visallo.search.advanced#savedQuerySelected
          * @property {object} data
          * @property {object} data.query
-         * @property {object} data.query.url The search endpoint
-         * @property {object} data.query.parameters The search endpoint
-         * parameters
+         * @property {string} data.query.url The search endpoint
+         * @property {object} data.query.parameters The search endpoint parameters
          */
         this.onSavedQuerySelected = function(event, data) {
             if ($(event.target).is(this.currentSearchNode)) return;
-            this.trigger(
-                'searchByParameters',
-                {
-                    ...data.query,
-                    submit: true
-                }
-            );
+
+            if (this.advancedActive) {
+                const params = this.advancedSearchExtension.params();
+                this.advancedSearchExtension.params({
+                    ...params,
+                    initialParameters: data.query.parameters
+                }).attach();
+            } else {
+                this.trigger(
+                    'searchByParameters',
+                    {
+                        ...data.query,
+                        submit: true
+                    }
+                );
+            }
         };
 
         this.onSearchByParameters = function(event, data) {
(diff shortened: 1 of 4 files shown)


web/war/src/main/webapp/js/search/search.js
```diff
@@ -670,27 +672,35 @@ define([
...
+                            resultsSelector: '.search-pane .' + cls,
+                            resultsContainer: $('.search-pane .' + cls)
+                        })
+                        .behavior({
+                            setCurrentSearchForSaving: (evt, query) => {
+                                self.trigger($container, 'setCurrentSearchForSaving', query);
+                            }
+                        })
+                        .attach();
+                    self.trigger($container, 'searchtypeloaded', { type: newSearchType });
```
jharwig: This is semantically different than previously since the `path` isn't loaded yet. This should be in the `attach().then`

web/war/src/main/webapp/js/search/search.js
```diff
@@ -670,27 +672,35 @@ define([
...
+                     * @fires org.visallo.search.advanced#setCurrentSearchForSaving
+                     * @example <caption>Rendering results</caption>
+                     * List.attachTo($(this.attr.resultsSelector), {
+                     *     items: results
+                     * })
+                     */
+                    Attacher()
+                        .path(path)
+                        .node($container)
+                        .params({
```
jharwig: Should react components have to listen for `savedQuerySelected`?

joeferner: Yes, is there a way to denote that?

jharwig: I was thinking we'd pass the saved query as a prop. Since otherwise each component would have to: `$(this.refs.myRootEl.parentNode).on`

web/war/src/main/webapp/js/search/search.js
joeferner: done

web/war/src/main/webapp/js/search/search.js
```diff
@@ -670,27 +672,35 @@ define([
...
+                            resultsSelector: '.search-pane .' + cls,
+                            resultsContainer: $('.search-pane .' + cls)
+                        })
+                        .behavior({
+                            setCurrentSearchForSaving: (evt, query) => {
```
joeferner: @jharwig  I'm trying a flight component and getting

```
jquery.js:2735 Uncaught RangeError: Maximum call stack size exceeded
    at n.fn.init (https://localhost:8889/libs/jquery/dist/jquery.min.js?vslo_v=1495577270581:2:23786)
    at n (https://localhost:8889/libs/jquery/dist/jquery.min.js?vslo_v=1495577270581:2:405)
    at Component.trigger (https://localhost:8889/libs/flightjs/build/flight.js?vslo_v=1495577270581:1267:22)
    at setCurrentSearchForSaving (https://localhost:8889/jsc/search/search.js?vslo_v=1495577270581:1:14813)
```

I'm guessing it's because of this.

What is this `behavior` method for. And I also saw a `legacyMapping` what is that for

web/war/src/main/webapp/js/search/search.js
joeferner: How do we pass it as a prop such that when a new save search is selected it gets updated?

joeybrk372:
I will update docs with new github links after https://github.com/visallo/doc-examples/pull/2 is merged

joeferner:
@joeybrk372 is this good to go?

joeybrk372:
yes, we can go ahead and merge

PR #1173, opened 2017-05-24
Fix timezone conversions when set to new timezone in settings

joeferner:
- [x] joeferner
- [x] mwizeman sfeng88

### DIFF
--- a/web/war/src/main/webapp/js/util/formatters.js
+++ b/web/war/src/main/webapp/js/util/formatters.js
@@ -760,32 +760,17 @@ define([
              */
             local: function(str) {
                 if (_.isUndefined(str)) return '';
+                var tz = FORMATTERS.timezone.currentTimezone();
                 var numberOrString = _.isString(str) && !isNaN(Number(str)) ? Number(str) : str,
                     dateInLocale;
 
                 if (_.isDate(numberOrString)) {
                     dateInLocale = numberOrString;
                 } else {
-                    dateInLocale = new Date(numberOrString);
+                    dateInLocale = new Date(moment.tz(numberOrString, tz.name).toISOString());
                     if (isNaN(dateInLocale.getTime())) {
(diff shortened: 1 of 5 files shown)


web/war/src/main/webapp/js/util/formatters.js
```diff
@@ -760,32 +760,16 @@ define([
...
-                            if (hours.length < 2) hours = '0' + hours;
-                            if (minutes.length < 2) minutes = '0' + minutes;
-
-                            parsed += (isNegative ? '-' : '+') + hours + ':' + minutes;
-                            dateInLocale = new Date(parsed);
-                        } else {
-                            console.warn('Unable to parse date', numberOrString);
-                            return '';
-                        }
+                        console.warn('Unable to parse date', numberOrString);
```
joeferner: I couldn't get into this branch of the code, I'm guessing because moment is just parsing it. When was the other code exercised?

web/war/src/main/webapp/js/util/formatters.js
```diff
@@ -760,32 +760,17 @@ define([
...
                 var numberOrString = _.isString(str) && !isNaN(Number(str)) ? Number(str) : str,
                     dateInLocale;
 
                 if (_.isDate(numberOrString)) {
                     dateInLocale = numberOrString;
                 } else {
-                    dateInLocale = new Date(numberOrString);
+                    dateInLocale = new Date(moment.tz(numberOrString, tz.name).toISOString());
                     if (isNaN(dateInLocale.getTime())) {
-                        var match = numberOrString.match(/(\d{4}-\d{2}-\d{2})\s(\d{2}:\d{2})\s(.*)$/);
```
joeferner: I couldn't get into this branch of the code, I'm guessing because moment is just parsing it. When was the other code exercised?

jharwig: Some older versions of firefox wouldn't be able to parse so we had to do this. Might not be necessary with Firefox 38 now

PR #1179, opened 2017-06-01
Compile less and cache the result

jharwig:
- [x] joeferner
- [x] mwizeman sfeng88

Improves performance of page load.

Testing Instructions: Plugin less files "Display as [name].less.css in network panel" should load correctly.

CHANGELOG
Changed: Plugin Less stylesheets are compiled once and cached. In `devMode=true` the last modified time is checked for recompilation


sfeng88:
can this be merged?

PR #1187, opened 2017-06-09
Changed: Replace breadthfirst graph layout with dagre

joeferner:
- [x] joeferner
- [x] sfeng88
- [ ] mwizeman joeybrk372 rygim jharwig EvanOxfeld

Many times breadthfirst will create a single long line of entities on the graph when laying out nodes. 
Dagre is much better at spacing the nodes out horizontally as well.

Testing Instructions:
Add nodes with high linking to the graph. Right click and layout hierarchical. The nodes should layout better than before. 

CHANGELOG
Changed: Replace breadthfirst graph layout with dagre

### DIFF
--- a/web/war/src/main/webapp/js/visallo.js
+++ b/web/war/src/main/webapp/js/visallo.js
@@ -28,6 +28,8 @@ function(jQuery,
 
     $.ui = { keyCode: { ENTER: 13 } };
 
+    lockDownUnderscore(_);
+
     // Debug retina/non-retina by changing to 1/2
     // window.devicePixelRatio = 1;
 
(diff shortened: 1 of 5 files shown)


web/war/src/main/webapp/js/visallo.js
```diff
@@ -419,4 +421,10 @@ function(jQuery,
         return (/#\s*$/).test(url) || url.indexOf('#') === -1;
     }
 
+    function lockDownUnderscore(_) {
```
jharwig: Can we move this to the graph plugin that actually has the issue? Also some comment on why

joeferner: I added a comment, but I think this still should go here to prevent other plugins from including different versions of underscore or lodash in the future.

PR #1193, opened 2017-06-13
log out in user repository and send message when user logs out

rygim:
The app doesn't set the user's status to inactive so when it is queried in drop downs on the UI, it returns active even if the user has logged out. This sets the user's status correctly on logout. 

Fixes https://github.com/visallo/visallo-lts/issues/2747 (and probably other status bugs since this seems like it would be a problem regardless of auth implementation)

- [ ] joeferner
- [ ] sfeng88
- [x] mwizeman joeybrk372 rygim jharwig EvanOxfeld

Testing Instructions: Open up two separate user windows of Visallo. Share a case with another user. Log out of the user that was shared with, then check the user status again.

Points of Regression: N/A

CHANGELOG
Fixes: User logout now sends user logout message to UI and sets user status to OFFLINE explicitly


joeferner:
I pretty sure status is managed by the web socket connection. https://github.com/visallo/visallo/blob/f11aa654a0e410eb4b565596fa56271f4aefa696/web/web-base/src/main/java/org/visallo/web/Messaging.java#L173

I don't think this PR handles the case of multiple browsers which I believe the web socket connection is trying to handle.

rygim:
Gotcha. Do we want to leave multiple browsers logged in like that even if a user explicitly logs out? Checking now to see why that event doesn't seem to be firing enough to show that a user has logged out.

rygim:
I am intermittently getting a null session on logout, which means that it can't get the user id and doesn't ever decrement the counter based on the websockets closing. Still looking...

joeferner:
I'm getting similar problems on my project. It happens when the users close the tab as well. So what ever we do we'll need to handle that as well. BTW my project doesn't have a logout button.

rygim:
Alternate solution: https://github.com/visallo/visallo-enterprise-plugins/pull/503

PR #1194, opened 2017-06-13
log out in user repository and send message when user logs out

rygim:
See https://github.com/visallo/visallo/pull/1193 for release-3.1 PR

Testing Instructions: Open up two separate user windows of Visallo. Share a case with another user. Log out of the user that was shared with, then check the user status again.

CHANGELOG
Fixes: User logout now sends user logout message to UI and sets user status to OFFLINE explicitly
Testing Instructions:


rygim:
Alternate fix: https://github.com/visallo/visallo-enterprise-plugins/pull/503

PR #1212, opened 2017-06-23
Update to actual browser version requirements

jharwig:
- [x] joeferner
- [ ] sfeng88
- [x] mwizeman joeybrk372 rygim jharwig EvanOxfeld

Testing Instructions: Open Visallo with non-supported browser < ie11 or <firefox 38. Should see a message about not supported.

We may want to update this filter to show a better error page, with maybe a button to continue anyway that sets a cookie.

CHANGELOG
Changed: Updated user agent filter to new browser version requirements


sfeng88:
this is good with me too

PR #1217, opened 2017-06-27
Fix OntologyRepositoryTestBase for OWL files in Jars

mwizeman:
For certain invocations of mvn, when the test was trying to load the OWL files from the classpath they were getting a URI to the file within the compiled jar in the m2 repository. Passing this URI to the File constructor was failing the tests. This change will check to see if the URI is in a jar or on the filesystem. If in a jar, it will first extract to the filesystem before continuing to import the OWL file. 

- [x] joeferner
- [x] sfeng88
- [ ] mwizeman joeybrk372 rygim jharwig EvanOxfeld

Testing Instructions:
Ensure the ontology unit tests pass.

### DIFF
--- a/core/core-test/src/main/java/org/visallo/core/model/ontology/OntologyRepositoryTestBase.java
+++ b/core/core-test/src/main/java/org/visallo/core/model/ontology/OntologyRepositoryTestBase.java
@@ -11,7 +11,12 @@ import org.visallo.core.util.VisalloInMemoryTestBase;
 import org.visallo.web.clientapi.model.PropertyType;
 
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


core/core-test/src/main/java/org/visallo/core/model/ontology/OntologyRepositoryTestBase.java
```diff
@@ -242,19 +246,30 @@ private void validateChangedOwlProperties() throws IOException {
...
+            testOwl = owlFilePath.toFile();
+            InputStream owlFileStream = OntologyRepositoryTestBase.class.getResourceAsStream(owlFileResourcePath);
+            IOUtils.copy(owlFileStream, new FileOutputStream(testOwl));
+
+            InputStream glyphIconStream = OntologyRepositoryTestBase.class.getResourceAsStream(GLYPH_ICON_FILE);
+            IOUtils.copy(glyphIconStream, new FileOutputStream(glyphIconPath.toFile()));
+        } else {
+             testOwl = new File(owlUri);
+        }
+        System.out.println(testOwl);
```
joeferner: remove sout

mwizeman: Done. Good catch!

PR #1223, opened 2017-07-12
Fix context menu selections

joeybrk372:
- [x] joeferner
- [x] sfeng88

The flight wrapper for the context menu didn't teardown it's react children so you would get the previous menu when clicking a different element.

Brought this fix https://github.com/visallo/visallo/pull/996 from 3.1 for selection configuration.

CHANGELOG
Added: You can now use the Select Connected option from the context menu if you are clicking on a single entity on the graph without a selection.


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -140,7 +140,13 @@ define([
                 selectConnected: (event, data) => {
                     event.stopPropagation();
                     const cy = this.cytoscape.state.cy;
-                    const selected = cy.elements().filter(':selected');
+                    let selected = cy.elements().filter(':selected');
+
+                    if (selected.length === 0) {
+                        const id = data.collapsedNodeId || data.vertexId || date.edgeIds;
+                        selected = cy.elements().filter(cyElement => cyElement.id() === id);
```
jharwig: Can't you use cy.getElementById here?

PR #1225, opened 2017-07-12
Changed: disable select all in search results if all elements not loaded

joeferner:
- [x] joeferner
- [x] sfeng88

Being able to select all elements before they are all loaded in the search
results panel is misleading. This commit adds a message indicating that
all results are not loaded yet

see https://github.com/visallo/visallo-lts/issues/1799

CHANGELOG
Changed: disable select all in search results if all elements not loaded

### DIFF
--- a/core/core/src/main/resources/MessageBundle.properties
+++ b/core/core/src/main/resources/MessageBundle.properties
@@ -648,6 +648,8 @@ search.help.scope=Search
 search.help.toggle=Show search pane
 search.query.invalid=Invalid Query
 
+search.noSelectAllWhileLoading=Select all is not available while results are loading
+
 properties.groups.count.hover.property=property
 properties.groups.count.hover.property.plural=properties
 properties.groups.count.hover.value=value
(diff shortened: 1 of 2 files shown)


core/core/src/main/resources/MessageBundle.properties
```diff
@@ -648,6 +648,8 @@ search.help.scope=Search
 search.help.toggle=Show search pane
 search.query.invalid=Invalid Query
 
+search.noSelectAllWhileLoading=Select all is not available while results are loading
```
joeferner: I'm open to suggestions on better wording here

joeferner:
What would be the purpose of selecting all when it's not really all?

jharwig:
Very useful to load get the first 20 or so elements on the graph without scrolling until it's done

kunklejr:
I suggest we remove select all entirely. I agree with Joe that it's misleading even with a warning message. And that misleading of users could have significant consequences when it results in an incomplete investigation.

PR #1232, opened 2017-07-17
Override bootstrap typeahead methods to fix scrolling in IE

joeybrk372:
- [x] joeferner
- [x] sfeng88

CHANGELOG
Fixed: In IE, some dropdown menus that provided typeahead suggestions could not be scrolled by clicking or dragging the scrollbar.


sfeng88:
@joeybrk372 can you please submit an different issue for the related bug that we discussed?

PR #1236, opened 2017-07-21
Added: method to graph plugins to support clearing the graph style cache

joeferner:
- [x] joeferner
- [ ] sfeng88

This allows graph plugins which rely on the style cache to clear the cache if any changes are made that are not picked up by the redux state change such as user preference changes

CHANGELOG
Added: method to graph plugins to support clearing the graph style cache


jharwig:
I'd rather not have plugins do this. What's not working?

joeferner:
@jharwig when a user preference is changed (ie toggle plugin) the graph is not refreshed

jharwig:
I think we previously had something like `refreshGraphStylehsheet` event. Is it style changes that aren't updated?

PR #1243, opened 2017-07-25
Added: VisibilityTranslatorBase

joeferner:
- [x] joeferner
- [x] sfeng88

When creating new VisibilityTranslators instead of extending from a mostly empty VisibilityTranslator and the overly filled out DirectVisibilityTranslator. VisibilityTranslatorBase gives a middle ground by providing implementations for most of the methods.

CHANGELOG
Added: VisibilityTranslatorBase


core/core/src/main/java/org/visallo/core/security/VisibilityTranslator.java
```diff
@@ -1,16 +1,12 @@
 package org.visallo.core.security;
 
-import org.visallo.web.clientapi.model.VisibilityJson;
 import org.vertexium.Visibility;
-
-import java.util.Map;
+import org.visallo.web.clientapi.model.VisibilityJson;
 
 public abstract class VisibilityTranslator {
```
mwizeman: Should this be an interface? Changing it would impact backwards compatibility for anyone that's extended it in their code.

joeferner: It should be an interface but didn't want to mess with it for that reason (breaking code)

PR #1247, opened 2017-08-02
(Waiting on Dynamic Ontology) DiffPanel performance improvements

jharwig:
- [x] joeferner
- [x] sfeng88

*_IMPORTANT: Set base against master after dynamic-ontology merges_*

## Back-End
* remove title eval for non-deleted diffs
* remove unnecessary vertex request for non-deleted diffs

## Front-End
* Use virtualized list to render
* Title evaluation is lazy and initiated when items in scroll view
* Prevent dup react renders
* Add loading

Tested with 15K diff changes, back-end response 2s, front-end 1s.

Testing Instructions: Use structured import to import 1k+ items into case. Inspect diff count load time and diff panel opening load times.

Points of Regression: Visibility display in diff panel won't wrap if it can't fit, will clip.

CHANGELOG
Fixed: Case Publish/Undo list (DiffPanel) is more performant for large datasets in imported into case through structured import.


web/war/src/main/webapp/js/workspaces/diff/DiffPanel.jsx
```diff
@@ -246,72 +322,125 @@ define([
...
             );
         },
 
+        getTitles(startIndex, stopIndex) {
+            if (this.titlesRequest) {
+                this.titlesRequest.cancel();
+            }
+            const { titles } = this.state;
+            const { flatDiffs } = this.props;
+            const ids = Object.keys(flatDiffs.slice(startIndex, stopIndex + 1).reduce((ids, diff) => {
```
joeybrk372: should we check to see if the elements have changed here? Otherwise the titles we are storing in the component state could become outdated

web/war/src/main/webapp/js/workspaces/diff/diff.js
```diff
@@ -124,21 +166,20 @@ define([
...
-                            .map(function(object) { return object.id; })
-                            .reduce(function(selected, id) {
-                                selected[id] = true;
-                                return selected;
-                            }, {}),
-                        previousDiffsById = self.diffsById || {};
+            ]).spread(function(store, selectedObjects) {
+                var state = store.getState(),
+                    workspaceId = state.workspace.currentId,
+                    verticesById = state.element[workspaceId] || {},
```
joeybrk372: `state.element[workspaceId].vertices`

web/war/src/main/webapp/js/workspaces/diff/diff.js
```diff
@@ -124,21 +166,20 @@ define([
...
+                    edgesById = verticesById,
```
joeybrk372: `state.element[workspaceId].edges || {}`

web/war/src/main/webapp/js/workspaces/diff/DiffPanel.jsx
jharwig: We don't need to check if they change. Since when we request titles we are adding them to the vertex cache. So when they are changed, the diff panel re-renders and pulls in the cached version, and ignores that first title.

I tested, and it works (after the change from your other comment with element cache). If works if another user in same case updates it too.

PR #1248, opened 2017-08-07
(Waiting on Dynamic Ontology) Fix saved search table breaking with ontology updates

joeybrk372:
- [x] joeferner
- [x] sfeng88

**_Important: Change base to master before merge_**

Testing Instructions: Verify saved search table still works after adding/deleting concepts, edges, and properties that are part of the results.

no CHANGELOG


sfeng88:
@joeybrk372 does this fix also work if items in the ontology is deleted?

joeybrk372:
yes, it will also handle deletions

PR #1251, opened 2017-08-08
Don't override async load of worker scripts in development

joeybrk372:
- [x] joeferner
- [x] sfeng88

The way we make importScripts async by wrapping it in and then importing a blob URL does not allow dev tools to pick up the sources. Worker scripts would still stop if you put debugger statements, but browsing and setting breakpoints in dev tools is much easier.

no CHANGELOG

### DIFF
--- a/web/war/src/main/webapp/js/data/web-worker/data-worker.js
+++ b/web/war/src/main/webapp/js/data/web-worker/data-worker.js
@@ -163,7 +163,10 @@ function setupRequireJs(data, callback) {
     require.deps = data.webWorkerResources;
     require.callback = callback;
     importScripts(BASE_URL + '/libs/requirejs/require.js?' + data.cacheBreaker);
-    require.load = asyncRequireJSLoader
+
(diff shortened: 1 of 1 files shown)


jharwig:
Would it not be better to just make the source maps work with the loader?

jharwig:
On second thought it seems like too much work to deal with plugin worker files too so it's probably fine as long as we are testing in production mode on ET

web/war/src/main/webapp/js/data/web-worker/data-worker.js
```diff
@@ -163,7 +163,10 @@ function setupRequireJs(data, callback) {
     require.deps = data.webWorkerResources;
     require.callback = callback;
     importScripts(BASE_URL + '/libs/requirejs/require.js?' + data.cacheBreaker);
-    require.load = asyncRequireJSLoader
+
+    if (!visalloEnvironment.dev) {
```
jharwig: `visalloEnvironment.prod`?

PR #1259, opened 2017-08-11
Add ability to expand/collapse long string properties in element inspector

joeybrk372:
- [x] joeferner
- [x] sfeng88

expand and collapse property

The expanded state is stored per property and will persist until you refresh your browser.

Also added scroll overflow for justification in the popover and property form and truncation in element inspector headers and vertex connection popover.

Testing Instructions:

1. Add a new string property with a very long value.
1. The property should be truncated and have an option to expand/collapse when hovering.
1. Expand the property.
1. Close the element inspector then reopen.
1. The property should still be expanded.

CHANGELOG
Added: Ability to collapse and expand long text properties in the element inspector.
Fixed: `formatters.vertex.prop` and `formatters.vertex.propDisplay` can also provide an array of arguments to pass to optional transforms. (e.g. `formatters.vertex.propDisplay('myPropertyIRI', 'property value', { truncate: [5] })`
Fixed: When pasting a large amount of text as justification, display it with a scrollbar rather than filling the screen to fit.


joeybrk372:
removed the twizzle

PR #1260, opened 2017-08-14
Fix user admin authorization plugin adding only a single authorization with commas

joeybrk372:
- [x] joeferner
- [x] sfeng88

Testing Instructions: Attempt to add an authorization to a user through the admin tools user editor with commas. The user should get one authorization for each comma-separated value

CHANGELOG
Fixed: Attempting to add an authorization to a user that contained commas via the user admin authorization tool would prevent the user from logging in. Now, an authorization will be added for each value separated by commas.


web/plugins/admin-user-property-auth/src/main/java/org/visallo/web/plugin/adminUserTools/userPropertyAuth/UserAddAuthorization.java
```diff
@@ -40,7 +41,10 @@ public JSONObject handle(
             throw new VisalloAccessDeniedException("Authorization repository does not support updating", authUser, userName);
         }
 
-        ((UpdatableAuthorizationRepository) authorizationRepository).addAuthorization(user, auth, authUser);
+        for (String authStr : auth.split(SEPARATOR)) {
```
jharwig: should we ignore spaces?

kunklejr: We should definitely ignore spaces. Should be as simple as calling `authStr.trim()` before the `addAuthorization` call?

joeferner: We should probably make that the responsibility of the `addAuthorization` method itself not the caller?

joeybrk372: updated to trim whitespace

sfeng88:
=

PR #1261, opened 2017-08-15
Properly Handle Workspace Id's when Deleting a Workspace

mwizeman:
There are three main changes involved in this fix for the deletion of a workspace.
1. Updated the result writer to not attempt to append ACL values if the response is a simple VisalloResponse.SUCCESS. This fixed the WorkspaceDelete route. The result writer was causing an error since the workspace id available was for the workspace that was just deleted. Since it's a simple SUCCESS response, there's no need to load the ontology and attempt to append the ACL.
2. Updated the result writer to first check if the object being written is a workspace. In that case, the ACL should be appended using its id rather than the id of whatever workspace is selected. This fixes the WorkspaceCreate route.
3. Updated the result writer to prefer workspace id's from request values rather than what is stored as the user's currentWorkspaceId. This fixes a race condition where an old workspace has been deleted and a new one has not yet been persisted as the current workspace.

- [x] joeferner
- [x] sfeng88

Testing Instructions:
1. Create a new user and log in
2. Create a new Case
3. Delete the default Case
4. Ensure no errors occur in the logs
5. Delete the new Case, which should be the user's last Case
6. Ensure no errors occur in the logs
7. Ensure that a new default Case is created and made active


jharwig:
Don't think you need a changelog since this fixes a 4.0 only feature?

mwizeman:
@jharwig Fixed. Thanks!

PR #1267, opened 2017-08-16
Sanitize filenames before writing them to the local filesystem

mwizeman:
When a user drops a file onto Visallo, the web server writes it to a temporary file on its local storage. Depending on the filesystem in use, certain characters are not allowed. This PR changes the code to use a generated temp file name rather than trying to sanitize and use the original filename. As part of this, the system will also set the original filename as the property on the vertex rather than the processed name.

- [ ] joeferner
- [x] sfeng88

Testing Instructions:
1. Upload files with illegal characters and percent encoding sequences.
2. Ensure that the upload succeeds and that the vertex displays the original filename

Points of Regression:
None

CHANGELOG
Fixed: Uploading files with illegal characters in their names
Added: The temporary directory that file uploads are written to can be configured via the property `org.visallo.web.routes.vertex.VertexImport.tempDir`


kunklejr:
Do we care about the temporary file's name? If not, I would suggest just using `File.createTempFile` and avoid all the stripping and character replacing altogether.

mwizeman:
@kunklejr Good point! I've updated the PR to use a java tempfile.

@sfeng88 and @joeferner The code differs from what you originally signed off on so I unchecked your boxes. Please take a look when you have time.

@joeferner Jeff also mentioned that you may need to move the temp directory location around, so I made that configurable via the property `org.visallo.web.routes.vertex.VertexImport.tempDir`

PR #1270, opened 2017-08-17
Structured ingest defaults to text property visibility

joeybrk372:
- [x] joeferner
- [x] sfeng88

1. The default value for visibility when mapping items is now set as the text property's visibility. 
1. Fixed an issue where you could not set the visibility of mapped elements/properties to public if the structured file had a visibility itself.
1. Added missing i18n
1. Fixed css issue where top buttons for 'Create Entities', 'Create Relationships', etc. would wrap

CHANGELOG
Changed: In the structured file importer, if the source text has a visibility that will be the default value for all mappings.
Fixed: Elements set to have a public visibility in the structured file importer would get the visibility of the file if that file had its own visibility.


sfeng88:
@joeybrk372  please fix merge conflicts and then merge

PR #1271, opened 2017-08-18
Cleanup User currentWorkspace settings on Case Deletion

mwizeman:
Fix multiple issues related to users having a CURRENT_WORKSPACE that is deleted:

1. Update the workspace repository code to unset the CURRENT_WORKSPACE property for any user that has their current workspace deleted.
2. Update the MeGet route so that if it has to create a new workspace for a user it sets it as the CURRENT_WORKSPACE
3. Update the WorkspaceCreate route to not include the CURRENT_WORKSPACE as an authorization (happened automatically when declaring auths as a parameter). This was a problem in the event a user had just deleted their only case. The front end will send a request to create a new case with the id of the old one as a header.

- [x] joeferner
- [x] sfeng88

Testing Instructions:
1. Create and delete cases and ensure that the user interface does not encounter any problems
2. Share cases with between users and delete them both while a user is online an offline

CHANGELOG
Fixed: Occasional issue related to users having a current workspace preference saved that is but that workspace is deleted.

### DIFF
--- a/web/web-base/src/main/java/org/visallo/web/routes/user/MeGet.java
+++ b/web/web-base/src/main/java/org/visallo/web/routes/user/MeGet.java
@@ -1,11 +1,9 @@
 package org.visallo.web.routes.user;
 
-import com.google.common.collect.Iterables;
 import com.google.inject.Inject;
 import com.v5analytics.webster.ParameterizedHandler;
 import com.v5analytics.webster.annotations.Handle;
 import com.v5analytics.webster.handlers.CSRFHandler;
-import org.vertexium.util.IterableUtils;
 import org.visallo.core.model.user.UserRepository;
 import org.visallo.core.model.workspace.Workspace;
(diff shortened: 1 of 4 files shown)


web/web-base/src/main/java/org/visallo/web/routes/user/MeGet.java
```diff
@@ -74,6 +71,7 @@ public ClientApiUser handle(
                 workspace = workspaceRepository.add(user);
             }
 
+            userRepository.setCurrentWorkspace(user.getUserId(), workspace.getWorkspaceId());
```
kunklejr: Does this cause a write on every request?

mwizeman: Since it only happens when the user vertex doesn't already have a current workspace, it shouldn't cause a write unless they need it.

PR #1275, opened 2017-08-23
notifications go from pre to pre-line if expanded (2.2.4)

sfeng88:
- [x] joeferner
- [x] sfeng88

CHANGELOG
Changed: When notifications are expanded collapse multiple whitespaces to one


jharwig:
The changelog and the changes don't seem to match. They were never `pre` so they would always collapse whitespace. I don't see any reason to change to pre (or pre-line.) We were switching from no wrapping with an ellipsis to a wrapping line on expanding.

What issue are you trying to solve?

PR #1276, opened 2017-08-23
notifications go from pre to pre-line if expanded (3.1.x)

sfeng88:
2.2.4 PR: https://github.com/visallo/visallo/pull/1275


jharwig:
Why do we need to back port this to 3.1?

PR #1278, opened 2017-08-23
Use @visallo/updeep fork

joeybrk372:
- [x] joeferner
- [x] sfeng88

We patched a forked version of updeep to:
1. Recursively return the same instances of nested arrays if no changes are made.
1. `u.constant(value)` will return the same instance if there are no changes

This solves the problem of the store getting unnecessary updates. An example of this: the element inspector having to re-render when search results of the same elements are returned.

Also added element reducer tests.

Testing Instructions:
1. Open vertex in element inspector
1. Search for the vertex so it is loaded in search results.
1. The element inspector should not flash(re-render)

CHANGELOG
Changed: Switched to use a forked version of substantial/updeep.


joeferner:
@joeybrk372 have we sent a PR to updeep?

jharwig:
@joeferner We will try, but I'm not sure they will accept since the `u.constant` change could have performance impact. It doesn't matter to us since we'd have to do it anyways.

jharwig:
I'm going to enter an issue and see their thoughts before submitting anything. These might be design choices rather than bugs in their mind

PR #1281, opened 2017-08-28
Allow saved search table to read configuration by concept type in addition to by saved search ID

sfeng88:
- [x] joeferner
- [x] sfeng88

Testing Instructions: N/A

Points of Regression: N/A

CHANGELOG
Added: Table configuration by concept type

### DIFF
--- a/web/plugins/table/src/main/resources/org/visallo/web/table/js/card/SavedSearchTableCard.jsx
+++ b/web/plugins/table/src/main/resources/org/visallo/web/table/js/card/SavedSearchTableCard.jsx
@@ -201,6 +201,20 @@ define([
                         if (_.isUndefined(tableView.showRowNumbers)) {
                             tableView.showRowNumbers = DEFAULT_ROW_NUMBERS;
                         }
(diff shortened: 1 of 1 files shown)


web/plugins/table/src/main/resources/org/visallo/web/table/js/card/SavedSearchTableCard.jsx
```diff
@@ -201,6 +201,20 @@ define([
                         if (_.isUndefined(tableView.showRowNumbers)) {
                             tableView.showRowNumbers = DEFAULT_ROW_NUMBERS;
                         }
+                    } else if (tableSettings && search.parameters.conceptType && tableSettings[search.parameters.conceptType]) {
```
joeybrk372: Element level configuration is only stored by search id inside `tableSettings`. If they are there it is probably an issue with how we save the card. I don't think finding another arbitrary search's configuration for that concept type would be a good improvement here.

sfeng88: How would a plugin customize the type of columns a user can filter by based on concept type if they don't want to display all of the properties defined in the ontology?

PR #1286, opened 2017-08-29
add tooltip when hovering over saved search hamburger menu

sfeng88:
- [x] joeferner
- [x] sfeng88

Testing Instructions: When you hover over the hamburger menu on the saved search table it should have a tooltip with "Filter"

Points of Regression: N/A

CHANGELOG
Added: tooltip for hamburger menu of saved search table


web/plugins/table/src/main/resources/org/visallo/web/table/messages.properties
```diff
@@ -9,3 +9,4 @@ com.visallo.table.open.search.plural= Open in search with {0} more results...
 com.visallo.table.config.search=Configure Saved Search...
 com.visallo.table.config.search.label=Search
 com.visallo.table.config.search.placeholder=Choose Saved Search
+com.visallo.table.config.columns.filter=Filter
```
jharwig: Is it a filter? Maybe `Show, Hide and Re-arrange Columns`

sfeng88: Fixed and rename the message bundle property to `com.visallo.table.config.columns.hover`

PR #1288, opened 2017-08-29
Configurable date time display 

sfeng88:
- [x] joeferner
- [x] sfeng88

Testing Instructions: When `web.ui.showDateTime` configuration property is set to true, comments time should show actual time instead of relative date/time. Tooltip should be the relative date/time

image

Points of Regression: N/A

CHANGELOG
Added: configuration option to change comment to show date time rather than relative date/time


web/war/src/main/webapp/js/comments/comments.js
```diff
@@ -6,6 +6,7 @@ define([
     'util/vertex/formatters',
     'util/withDataRequest',
     'util/popovers/propertyInfo/withPropertyInfo',
+    'data/web-worker/services/config',
```
jharwig: remove this

web/war/src/main/webapp/js/comments/comments.js
```diff
@@ -15,6 +16,7 @@ define([
     F,
     withDataRequest,
     withPropertyInfo,
+    configurations,
```
jharwig: remove this too. should never require things in the web-worker folder structure (there are some exceptions with redux store actions) But never services. data request handles this

web/war/src/main/webapp/js/comments/comments.js
jharwig: You should probably require `util/requirejs/promise!util/service/propertiesPromise` instead of using dataRequest. That will provide a properties map with all the properties at file require time.

By using the data request in the renderCommentLevel you are changing the method from sync to async and I have a feeling some bugs could come of that.

web/war/src/main/webapp/js/comments/comments.js
```diff
@@ -229,39 +231,56 @@ define([
...
-                    var modified = p[0].metadata['http://visallo.org#modifiedDate'],
-                        created = getCreated(p[0]) || modified,
-                        modifiedStr = modified && F.date.dateTimeString(modified) || '',
-                        createdStr = created && F.date.dateTimeString(created) || '';
-
-                    if (isEdited(created, modified)) {
-                        return i18n('detail.comments.date.hover.edited', createdStr, modifiedStr);
-                    }
-                    return modifiedStr;
+            this.dataRequest('config', 'properties')
```
jharwig: See comment above about using the require js promise for config properties. If you do the request here you're making this function async which might break calling code that expects the DOM to be ready after. And that change is easier than changing it to promise based

web/web-base/src/main/java/org/visallo/web/WebConfiguration.java
```diff
@@ -46,6 +46,7 @@
     public static final String MAX_SESSION_INACTIVE_INTERVAL_SECONDS = PREFIX + "maxSessionInactiveIntervalSeconds";
     public static final String TIMEZONE_DEFAULT_DETECT = PREFIX + "timezone.defaults.detect";
     public static final String TIMEZONE_DEFAULT_TIMEZONE = PREFIX + "timezone.defaults.timezone";
+    public static final String SHOW_DATE_TIME = PREFIX + "showDateTime";
```
jharwig: what about `date.default.display=relative` as the default and `exact` for new behavior. Did you add enterprise docs?

sfeng88: @jharwig changed. Enterprise docs are in a separate PR that I need to submit.

web/war/src/main/webapp/js/comments/comments.js
```diff
@@ -256,12 +266,18 @@ define([
...
                     if (isEdited(created, modified)) {
-                        return i18n('detail.comments.date.hover.edited', createdStr, modifiedStr);
+                        if (dateDisplay === 'exact') {
+                            return i18n('detail.comments.date.hover.edited', relativeCreatedStr, relativeModifiedStr);
+                        } else {
+                            return i18n('detail.comments.date.hover.edited', createdStr, modifiedStr);
+                        }
                     }
-                    return modifiedStr;
+                    return dateDisplay === 'exact' ? relativeModifiedStr : modifiedStr;
```
jharwig: Is this backwards?

web/war/src/main/webapp/js/comments/comments.js
```diff
@@ -256,12 +266,18 @@ define([
...
                         created = getCreated(p[0]) || modified,
                         modifiedStr = modified && F.date.dateTimeString(modified) || '',
-                        createdStr = created && F.date.dateTimeString(created) || '';
+                        createdStr = created && F.date.dateTimeString(created) || '',
+                        relativeModifiedStr = modified && F.date.relativeToNow(F.date.utc(modified)) || '',
+                        relativeCreatedStr = created && F.date.relativeToNow(F.date.utc(created)) || '';
 
                     if (isEdited(created, modified)) {
-                        return i18n('detail.comments.date.hover.edited', createdStr, modifiedStr);
+                        if (dateDisplay === 'exact') {
```
jharwig: Seems like these are flipped?

sfeng88: When we are displaying exact date, I wanted the hover text to display the relative date.

web/war/src/main/webapp/js/comments/comments.js
sfeng88: When we are displaying exact date, I wanted the hover text to display the relative date.

web/war/src/main/webapp/js/comments/comments.js
jharwig: ok, got it

PR #1294, opened 2017-09-05
Ellipsize long strings

joeybrk372:
- [x] joeferner
- [x] sfeng88

Testing Instructions: check that long property types, values, and element titles do not overflow their containers in activity display, edge/vertex/property selector inputs, diff panel, element inspector, & search filters.

CHANGELOG
Fixed: Many places where long names and values would overflow their containers.


joeybrk372:
@jharwig I agree with preferring flex/overflow over truncation, but for the LRP titles in the activity panel I wanted to allow multiple lines, the truncating gives ~3. I'd rather leave the truncation here than do a hairier multiline overflow solution

PR #1297, opened 2017-09-07
notification clear all plugin

sfeng88:
- [x] joeferner
- [x] sfeng88

Testing Instructions:
- Clicking on the trash can button should clear all notifications in the dashboard card. Non-system notifications should not reappear after a browser refresh

Points of Regression: N/A

CHANGELOG
Added: Notification clear all plugin for the notification dashboard card


joeferner:
@sfeng88 should this just be part of the notifications instead of a new plugin? If someone doesn't want it we could add a configuration toggle later. BTW I would like this feature.

web/plugins/notifications-clear-all/src/main/resources/org/visallo/web/notifications/plugin.js
```diff
@@ -0,0 +1,35 @@
+define(['public/v1/api'], function(visallo) {
+    'use strict';
+
+    visallo.registry.registerExtension('org.visallo.dashboard.toolbar.item', {
```
joeybrk372: I agree with Joe. You can register this extension at the top of `notifications.js` and then listen for the event directly in the component without having to send an extra request to get all notifications

web/plugins/notifications-clear-all/src/main/resources/org/visallo/web/notifications/plugin.js
```diff
@@ -0,0 +1,35 @@
...
+        identifier: 'org-viasllo-notification-clear-all',
```
joeybrk372: typo in visallo

jharwig:
@joeferner  Yes, I wasn't imagining this would be a (jar) plugin, just that the built-in notifications used the dashboard tools extension point to add the clear button. (what @joeybrk372  said)

sfeng88:
@joeferner @jharwig @joeybrk372  Fixed, please take another look at the PR

PR #1299, opened 2017-09-07
use the property visibility that the text is being extracted from rather than the vertex

sfeng88:
There are cases where the raw property visibility differs from the vertex visibility. Since TikaTextExtractorGPW is extracting text and setting other properties based on the raw property, we should be using the raw property visibility when adding new properties instead of the vertex visibility.

- [x] joeferner
- [x] sfeng88

CHANGELOG
Changed: All properties added by TikaTextExtractor should use raw property visibility instead of the vertex visibility.


sfeng88:
@joeferner are you ok with these changes?

PR #1300, opened 2017-09-07
use property visibility of property used to determine mimetype when adding mimetype

sfeng88:
There are cases where the property used to determine the mime type differs in visibility from the vertex visibility. We should be using the visibility of the property used to determine mime type and not the vertex visibility.

- [x] joeferner
- [x] sfeng88

CHANGELOG
Changed: All properties added by MimeTypeGPW should use property visibility instead of the vertex visibility.


sfeng88:
@joeferner are you ok with these changes?

PR #1301, opened 2017-09-07
Changed: all injected classes to have singleton annotation to clarify and fix issues

joeferner:
- [x] joeferner
- [x] sfeng88

CHANGELOG
Changed: all injected classes to have singleton annotation to clarify and fix issues


core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepositoryBase.java
```diff
@@ -1,4 +1,4 @@
-package org.visallo.core.model.ontology;
+    package org.visallo.core.model.ontology;
```
mwizeman: This looks like an accidental change

joeferner: good catch, fixed

PR #1302, opened 2017-09-08
VisalloPlugin, web set thread, and memory logger

joeferner:
- [x] joeferner
- [x] sfeng88

Added: VisalloPlugin annotation to support disabled by default plugins
    
Adding plugins which where disabled by default used to require either putting them in their own plugin and using plugin toggle or setting up a configuration item which disabled it. This commit allows creating plugins which are opt in.

Added: set thread name and memory logger web listeners
    
Set thread name web listener will set the current thread's name to "http-" follow by the current request URI. This allows profilers to easily see what thread is currently serving what URL.
    
Memory logger web listener uses JMX to determine the memory of the current thread before and after the web request is serviced and allows logging that memory if it exceeds a threshold.

CHANGELOG
Added: VisalloPlugin annotation to support disabled by default plugins
Added: set thread name and memory logger web listeners


web/web-base/src/main/java/org/visallo/web/webEventListeners/SetThreadNameWebEventListener.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+@VisalloPlugin(disabledByDefault = true)
+public class SetThreadNameWebEventListener extends DefaultWebEventListener {
+    public static final int PRIORITY = -1000;
+    private static final String THREAD_NAME_PREFIX = "http-";
+
+    @Override
+    public void before(WebApp app, HttpServletRequest request, HttpServletResponse response) {
+        Thread.currentThread().setName(getNewThreadName(request));
```
mwizeman: Not that it really matters, but do we want to set it back to its original name on the way out?

joeferner: I was thinking about that, but it might be easier to track in the profiler if it stays the name with URI longer.

PR #1303, opened 2017-09-08
New Text View Term UI with Support for Overlapping Terms

jharwig:
- [x] joeferner
- [x] sfeng88

Allow multiple overlapping terms in element inspector text views.

terms

* Switches TextHighlighting to streaming version
* New work queue events for privileges and auth changes (Needed for new justification component)
* Popover utilities now use size observers so no need to call `positionDialog` on component size changes

Testing Instructions:

Resolve/Unresolve terms. resolve terms highlighted by GPW, set justifications on properties, edges, and entities, both in fullscreen and normal inspector. Set properties, comments. Do the same for transcripts on video/audio

Replaces Components with react:
* Justification viewer/editor
* Element Selector
* Element (start of new element component, just underlined and draggable for now)
* Resolve form
* Unresolve form

Points of Regression: Previous edge justification references might show duplicates in text document.

CHANGELOG
Added: Allow multiple overlapping terms in element inspector text views.
Changed: Resolved terms can be longer than 10 words, configurable up to 5 paragraphs.
Security: Fixed possible XSS exploit using justification snippets


core/core/src/main/java/org/visallo/core/EntityHighlighter.java
```diff
@@ -1,94 +1,325 @@
...
+                        long c;
+                        c = o1.getStart() - o2.getStart();
+                        if (c == 0)
+                            c = o2.getEnd() - o1.getEnd();
+                        return c > 0 ? 1 : c < 0 ? -1 : 0;
+                    })
+                    .collect(Collectors.toList());
+
+            if (options == null) {
+                options = DefaultOptions; // EnumSet.noneOf(Options.class);
```
joeferner: remove comment

core/core/src/main/java/org/visallo/core/EntityHighlighter.java
```diff
@@ -1,94 +1,325 @@
...
+        StringBuilder selector = new StringBuilder(".text");
+        int outset = 0;
+        int lineHeight = 18;
+        for (int depth = 1; depth <= maxDepth; depth++) {
+            selector.append(" .res");
+            out.write(selector.toString());
+            out.write("{");
+            out.write("border-image-outset: 0 0 " + outset + "px 0;");
+            if (depth == 1) {
+                //out.write("border-bottom-color: transparent");
```
joeferner: remove comment

core/core/src/main/java/org/visallo/core/EntityHighlighter.java
```diff
@@ -1,94 +1,325 @@
...
+            writeBuffer(opened, all, out, buffer, innerOffset, preLength);
+            innerOffset += preLength;
+        }
+        return innerOffset;
+    }
+
+    private static void writeBuffer(List<OffsetItem> started, List<OffsetItem> all, OutputStreamWriter out, char[] buffer, int offset, int len) throws IOException {
+        String strBuffer = new String(buffer, offset, len);
+
+        /*if (strBuffer.contains("\n") && !started.isEmpty()) {
```
joeferner: remove comment

core/core/src/main/java/org/visallo/core/EntityHighlighter.java
```diff
@@ -1,94 +1,325 @@
...
-            result.append(">");
-            endOffsets.add((int) OffsetItem.getOffset(offsetItem.getEnd()));
-            lastStart = (int) OffsetItem.getOffset(offsetItem.getStart());
+        } else {*/
+            out.write(StringEscapeUtils.escapeXml11(strBuffer));
+//        }
+    }
+
+    private static boolean indexInsideCurrentBuffer(long index, int offset, int innerOffset, int len) {
+        return index >= offset && (index - offset) < BUFFER_SIZE;// < (offset + innerOffset + len);
```
joeferner: remove comment

core/core/src/main/java/org/visallo/core/EntityHighlighter.java
```diff
@@ -1,94 +1,325 @@
...
+
+    private static void addClosingOffsetItemSpan(OutputStreamWriter out, OffsetItem item) throws IOException {
+        addOffsetItemSpan(out, null, null, item, false, true);
+    }
+
+    private static void addOffsetItemSpan(OutputStreamWriter out, List<OffsetItem> opened, List<OffsetItem> all, OffsetItem item, boolean fullInfo, boolean closing) throws IOException {
+        if (!closing) {
+            JSONObject infoJson = item.getInfoJson();
+
+//            if (opened.size() == 0 || opened.size() == 1 && opened.indexOf(item) == 0) {
```
joeferner: remove comment

core/core/src/main/java/org/visallo/core/EntityHighlighter.java
jharwig: oops, yup

web/war/src/main/webapp/js/data/web-worker/services/vertex.js
```diff
@@ -401,11 +403,12 @@ define([
                  propertyName: property.name,
                  value: property.value,
                  visibilitySource: property.visibilitySource,
-                 oldVisibilitySource: property.oldVisibilitySource,
-                 justificationText: property.justificationText
```
joeybrk372: why not allow passing justification and default to the property's current justification value? Looks like the only place we do that currently is in the admin element editor, but could be useful and this way is an api change

web/war/src/main/webapp/js/data/web-worker/services/edge.js
```diff
@@ -99,11 +108,12 @@ define([
                  propertyName: property.name,
                  value: property.value,
                  visibilitySource: property.visibilitySource,
-                 oldVisibilitySource: property.oldVisibilitySource,
-                 justificationText: property.justificationText
```
joeybrk372: see vertex.setProperty comment

web/war/src/main/webapp/js/detail/text/popover/TermContainer.jsx
```diff
@@ -0,0 +1,311 @@
...
+                    startOffset: mentionStart,
+                    endOffset: mentionEnd,
+                    snippet,
+                    vertexId: artifactId,
+                    textPropertyKey: propertyKey,
+                    textPropertyName: propertyName
+                });
+            },
+            onFullscreen(term) {
+                closeDialog();
```
joeybrk372: might be nice to leave this open here?

web/war/src/main/webapp/js/components/visibility/VisibilityEditor.jsx
```diff
@@ -0,0 +1,78 @@
...
+        },
+        onChange(event) {
+            const value = event.target.value;
+            const valid = this.checkValid(value)
+            this.setState({ value, valid })
+            this.props.onVisibilityChanged({ value, valid })
+        },
+        checkValid(value) {
+            var authorizations = this.props.authorizations;
+            return Boolean(!value.length || value in authorizations);
```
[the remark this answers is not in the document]
jharwig: actually this component (and previously) only supported one visibility. Because the default translator doesn't support multiple. Once we create a character separator translator we can add support for those in the front-end

web/war/src/main/webapp/js/data/web-worker/services/vertex.js
jharwig: this isn't an API change. It just doesn't always send justificationText if sourceInfo is given.

web/war/src/main/webapp/js/detail/text/popover/TermContainer.jsx
jharwig: hmm, maybe. will have to test any weird behavior while you're in fullscreen page.

jharwig: done

web/war/src/main/webapp/js/components/visibility/VisibilityEditor.jsx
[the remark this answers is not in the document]
jharwig: @joeferner What do you think. I think we should only allow one to be set by default, and not support the boolean expressions. 

I would like to create a new component that is a simple Any of these/All of these multiselect. That we use as the default when it's built.

PR #1305, opened 2017-09-11
Provide more docs on windows building, and fix postinstall command

jharwig:
- [x] joeferner
- [x] sfeng88

Formula evaluator post install task syntax was failing on windows. Also added some docs for windows setup.

Testing Instructions: `mvn compile` on all supported platforms

Points of Regression: None

NO CHANGELOG
broken in dynamic ontology changes, not released yet


joeferner:
@mwizeman looks like your review is needed

jharwig:
I assigned to Matt because he was looking into it earlier and found another issue. (But I haven't heard from him since)

PR #1308, opened 2017-09-11
Work product controls toolbar/button configuration

joeybrk372:
- [x] joeferner
- [x] sfeng88

Added a product controls toolbar so that plugins which needed larger configuration components could have a better place to display their configuration UI.

toolbar preview

_*options displayed are just placeholders_

**To Do**: will update the example tutorials after this pr is merged

CHANGELOG
Added: Product controls toolbar to add functionality/controls to products. This is similar to old Graph/Map options but now uses the `org.visallo.product.toolbar.item` extension point.
Removed: Panning controls in the graph/map. Click and drag to pan the viewport instead.
Deprecated: `org.visallo.graph.options` and `org.visallo.map.options` extension points, use `org.visallo.product.toolbar.item` instead


web/plugins/map-product/src/main/resources/org/visallo/web/product/map/OpenLayers.jsx
```diff
@@ -581,28 +581,27 @@ define([
             el.addEventListener(type, handler, false);
         },
 
-        injectExtraPropsToTools() {
+        /**
+         * Map Work Product
+         *
+         * @typedef org.visallo.product.toolbar.item~Component
```
jharwig: Not sure how docs will render when component is defined twice with different attributes for graph/map. Think we should define 3 types:

`org.visallo.product.toolbar.item~(Component|GraphComponent|MapComponent)`

And have Graph/Map components extends Component

jharwig: will have to check rendered output to ensure it's clear what component will be passed

joeferner:
@joeybrk372 What is the status of this PR?

PR #1309, opened 2017-09-13
Back-end collapse API - Convert generic JSON objects to Java POJOs for strong typing

joeferner:
- [x] joeferner
- [x] sfeng88

This commit gives API users a type safe interface to collapse entities.

CHANGELOG
Changed: Back-end collapse API - Convert generic JSON objects to Java POJOs for strong typing

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductVertex.java
+++ b/core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductVertex.java
@@ -1,0 +1,51 @@
+package org.visallo.core.model.workspace.product;
+
+import org.visallo.web.clientapi.model.ClientApiObject;
+
+public class WorkProductVertex implements ClientApiObject {
+    private String id;
+    private boolean visible;
+    private String title;
+    private String type;
+    private boolean unauthorized;
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    public String getId() {
+        return id;
+    }
+
+    public void setVisible(boolean visible) {
+        this.visible = visible;
+    }
+
+    public boolean isVisible() {
+        return visible;
+    }
+
+    public void setTitle(String title) {
+        this.title = title;
+    }
+
+    public String getTitle() {
+        return title;
+    }
+
+    public void setType(String type) {
+        this.type = type;
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public void setUnauthorized(boolean unauthorized) {
+        this.unauthorized = unauthorized;
+    }
+
+    public boolean isUnauthorized() {
+        return unauthorized;
+    }
+}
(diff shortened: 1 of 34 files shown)


joeferner:
On my project I am generating workspaces and I need a way to collapse vertices using the new API. Typing in weakly types JSON seemed very fragile.

jharwig:
@joeferner would this prevent plugins from adding to extendedData?

joeferner:
@jharwig It works with the graph styling plugin (I assume it should work with others as well). Vertices, edges, and compound nodes are a little different in the extended data structure in that they aren't stored anywhere but are computed when asked for so this change doesn't effect the other parts of the extended data because of this.

Existing plugin extended data is retrieved in the getProductExtendedDataJson method https://github.com/visallo/visallo/pull/1309/files#diff-a0f439d5c2ad6a598890917b6f2915daR893 which still loads the property as JSON and appends to the JSON.

The saving is handled here https://github.com/visallo/visallo/blob/bd264f6b90f702c78da05ffc3046233af65962f3/core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java#L1073 which isn't effected by these changes.

core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductVertex.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.visallo.web.clientapi.model.ClientApiObject;
+import org.visallo.web.clientapi.model.GraphPosition;
+
+import java.util.List;
+
+public class WorkProductVertex implements ClientApiObject {
+    private String id;
+    private boolean visible;
+    private String parent;
```
jharwig: These are specific to graph. How do other products changes these?

web/plugins/graph-product/src/main/java/org/visallo/web/product/graph/GraphWorkProductService.java
```diff
@@ -528,52 +522,57 @@ private GraphPosition calculatePositionFromParents(
...
+            if (children != null) {
+                GraphProductOntology.NODE_CHILDREN.updateProperty(elemCtx, children, visibility);
+            }
         }
     }
 
-    protected void setEdgeJson(Edge propertyVertexEdge, JSONObject vertex) {
-        JSONObject position = ENTITY_POSITION.getPropertyValue(propertyVertexEdge);
+    @Override
+    protected void populateVertexWithWorkspaceEdge(Edge propertyVertexEdge, WorkProductVertex vertex) {
```
jharwig: Maybe we can change the signature to return a `WorkProductVertex` subclass here like `GraphWorkProductVertex`?

joeferner: I moved parent, pos, and children to a subclass.  Leaving id, visible, title, type, unauthorized in the base class.

jharwig:
Go ahead and merge after travis succeeds

PR #1316, opened 2017-09-15
Changed: Deprecate ApplicationBootstrapInitializer in favor of VisalloProcess

joeferner:
- [x] joeferner
- [x] sfeng88

ApplicationBootstrapInitializer requires running in a web container because as it's first
argument to the initialize method is a ServletContext.

CHANGELOG
Added: VisalloProcess
Deprecated: ApplicationBootstrapInitializer


web/web-base/src/main/java/org/visallo/web/ApplicationBootstrap.java
```diff
@@ -84,6 +79,25 @@ public void run() {
         }
     }
 
+    private void startVisalloProcesses(Configuration config) {
+        Iterable<VisalloProcess> processes = ServiceLoaderUtil.load(VisalloProcess.class, config);
```
joeferner: Last night, I thought of one quick thing I need to add. An option to turn off all processes in the web process. I'll add this later today before the merge @sfeng88 @mwizeman

PR #1317, opened 2017-09-15
GraphWorkProductService performance improvements

mwizeman:
1.) Added `FetchHint.EDGE_REFS` when loading vertices since they are only used to find edges
2.) Removed the `trimCompoundNodes` as that functionality isn't needed due to changes elsewhere.

- [x] joeferner
- [x] sfeng88

Testing Instructions:
1.) Load a workspace with several vertices
2.) Collapse the vertices
3.) Refresh the page and ensure vertices are still collapsed
4.) Uncollapse the vertices
5.) Use undo to make the vertices collapsed again
6.) Refresh the page and ensure vertices are still collapsed
7.) Uncollapse the vertices
8). Use undo to make the vertices collapsed again
9.) Use redo to make the vertices uncollapsed again
10.) Refresh the page and ensure vertices are not collapsed


web/plugins/graph-product/src/main/java/org/visallo/web/product/graph/GraphWorkProductService.java
```diff
@@ -193,36 +189,6 @@ private Nodes getNodes(
         return edges;
     }
 
-    private void trimCompoundNodes(Graph graph, Vertex productVertex) {
```
joeferner: What did this method do? and why don't we need it anymore?

mwizeman: @joeybrk372 Can elaborate more on this. My understanding was that it was initially intended to clean up nodes that had been uncollapsed during the next request of `/product`. I think that cleanup was once deferred due to support for undo/redo on the front end. This deferral wasn't really practical anyway since the cleanup could be triggered by another using requesting the product.

After talking with @joeybrk372, the suggestion was that if I collapsed/uncollapsed nodes and undo/redo support was in tact it was safe to remove this method. Following the testing instructions above, it looked good to me after removing that method.

joeybrk372: After taking a deeper look I was mistaken. `trimCompoundNodes` is removing collapsed nodes if all their backing vertices have been deleted. We're removing empty collapsed nodes when you request the extended data for a graph because otherwise those collapsed nodes would never be deleted.

Removing this would probably result in a net performance gain in most cases since we are still setting the `visible` property for collapsed nodes when we return the extended data. It would only suffer when a graph product accumulated many empty collapsed nodes.

PR #1321, opened 2017-09-18
Added: admin tool to display list of all active users

joeferner:
- [x] joeferner
- [x] sfeng88

Before I deploy new Visallo software I would like to verify that the web server is quite or at least get a list of users so that I can notify them before taking the system down for an upgrade. This admin plugin will display a list of all users marked as active in the system.

This tool also highlights the fact that our user active status logic is not working since it shows some users as active when they are not in the system. Might be nice for @nidasid to test with.

CHANGELOG
Added: admin tool to display list of all active users


web/plugins/admin-user-tools/src/main/java/org/visallo/web/plugin/adminUserTools/AdminUserToolsWebAppPlugin.java
```diff
@@ -24,7 +24,8 @@ public void init(WebApp app, ServletContext servletContext, Handler authenticati
         app.registerJavaScriptComponent("/org/visallo/web/adminUserTools/WorkspaceList.jsx");
         app.registerJavaScriptComponent("/org/visallo/web/adminUserTools/LoadUser.jsx");
         app.registerJavaScriptComponent("/org/visallo/web/adminUserTools/UserTypeaheadInput.jsx");
-        app.registerCss("/org/visallo/web/adminUserTools/userAdmin.css");
+        app.registerJavaScriptComponent("/org/visallo/web/adminUserTools/ActiveUserList.jsx");
```
jharwig: Don't have to block this PR, but we really need to migrate away from `registerJavaScriptComponent` for our plugins in favor of webpack builds. The slow the web server startup and the client because of all the separate requests.

joeferner: I'm all for that (startup time sucks). Is there an easy way to add webpack or is it just a copy all the webpack files from another place and tweak as needed.

jharwig: Yes, we could create a template and keep it up do date, but right now just use something like what's in graph-plugin. Or the bleeding edge plugins that use es6 modules check product-styling or product-annotations. You need the pom plugins, the .babelrc,  webpack.config.js, and some of the package.json

PR #1322, opened 2017-09-19
sort notification list newest to oldest

sfeng88:
- [x] joeferner
- [x] sfeng88

Testing Instructions:
- When adding new notifications, the dashboard notification card should show notifications newest to oldest
- Notification animations should show in the order the notifications come in (an easy way to test this is to change roles) 
- In the admin pane, list notifications should show newest to oldest

Points of Regression: N/A

CHANGELOG
Changed: Order notification lists to be newest to oldest


web/war/src/main/webapp/js/admin/bundled/notifications/list.js
```diff
@@ -90,16 +90,21 @@ define([
...
-                .data(_.chain(this.notifications)
+                .data(_.chain(sortedNotifications)
                       .groupBy(function(n) {
                           return n.active || n.startDate <= now ? 'Active' : 'Future';
                       })
                       .pairs()
                       .sortBy(function(p) {
                           return p[0];
                       })
+                      .reverse()
```
sfeng88: This is an extra line that is not needed. Will delete.

PR #1327, opened 2017-09-20
Migrate a visallo 3.x instance to visallo 4.0

mwizeman:
With the ontology updates, we now need the `intent` property to be indexed in the search engine. This migration will change the field definition in Vertexium and then reindex all of the ontology vertices.

While I was writing the migration, I noticed that a few other fields (USER_VISIBLE, SEARCHABLE...) were defined improperly as strings so I fixed those too. 

- [x] joeferner
- [x] sfeng88

Testing Instructions:
1. Shut down a running Visallo 3.x instance
2. Upgrade to Visallo 4.x
3. Run the new migration 
4. Start Visallo and ensure that the system starts properly and all data is maintained

CHANGELOG
Added: Migration script to update an existing Visallo 3.x installation to Visallo 4.x

### DIFF
--- a/tools/migrations/core/src/main/java/org/visallo/tools/migrations/MigrationBase.java
+++ b/tools/migrations/core/src/main/java/org/visallo/tools/migrations/MigrationBase.java
@@ -1,0 +1,66 @@
+package org.visallo.tools.migrations;
+
+import org.vertexium.Graph;
+import org.vertexium.GraphFactory;
+import org.visallo.core.cmdline.CommandLineTool;
+import org.visallo.core.config.Configuration;
+import org.visallo.core.exception.VisalloException;
+
+import static org.visallo.core.bootstrap.VisalloBootstrap.GRAPH_METADATA_VISALLO_GRAPH_VERSION_KEY;
+
+public abstract class MigrationBase extends CommandLineTool {
+    private Graph graph = null;
+
+    @Override
+    public final int run(String[] args, boolean initFramework) throws Exception {
+        return super.run(args, initFramework);
+    }
+
+    @Override
+    protected final int run() throws Exception {
+        graph = getGraph();
+        try {
+            Object visalloGraphVersionObj = graph.getMetadata(GRAPH_METADATA_VISALLO_GRAPH_VERSION_KEY);
+            if (visalloGraphVersionObj == null) {
+                throw new VisalloException("No graph metadata version set");
+            } else if (visalloGraphVersionObj instanceof Integer) {
+                Integer visalloGraphVersion = (Integer) visalloGraphVersionObj;
+                if (getFinalGraphVersion().equals(visalloGraphVersion)) {
+                    throw new VisalloException("Migration has already completed. Graph version: " + visalloGraphVersion);
+                } else if (!getNeededGraphVersion().equals(visalloGraphVersion)) {
+                    throw new VisalloException("Migration can only run from version " + getNeededGraphVersion() +
+                            ". Current graph version = " + visalloGraphVersion);
+                }
+            }
+
+            if (migrate(graph)) {
+                graph.setMetadata(GRAPH_METADATA_VISALLO_GRAPH_VERSION_KEY, getFinalGraphVersion());
+            }
+
+            graph.flush();
+            afterMigrate(graph);
+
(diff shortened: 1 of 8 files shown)


tools/migrations/core/src/main/java/org/visallo/tools/migrations/MigrationBase.java
```diff
@@ -0,0 +1,66 @@
...
+                throw new VisalloException("No graph metadata version set");
+            } else if (visalloGraphVersionObj instanceof Integer) {
+                Integer visalloGraphVersion = (Integer) visalloGraphVersionObj;
+                if (getFinalGraphVersion().equals(visalloGraphVersion)) {
+                    throw new VisalloException("Migration has already completed. Graph version: " + visalloGraphVersion);
+                } else if (!getNeededGraphVersion().equals(visalloGraphVersion)) {
+                    throw new VisalloException("Migration can only run from version " + getNeededGraphVersion() +
+                            ". Current graph version = " + visalloGraphVersion);
+                }
+            }
```
joeferner: what if it isn't an integer?

tools/migrations/pom.xml
```diff
@@ -14,7 +14,9 @@
     <name>Visallo: Tools: Migrations</name>
 
     <modules>
+        <module>core</module>
```
joeferner: Do we need a module per migration? Do we do this in-case of dependency version changes?

tools/migrations/predefined-to-dynamic-ontology/src/main/java/org/visallo/tools/migrations/PredefinedToDynamicOntology.java
```diff
@@ -0,0 +1,72 @@
...
+        return true;
+    }
+
+    @Override
+    protected void afterMigrate(Graph graph) {
+        if (graph instanceof GraphWithSearchIndex) {
+            VisalloBootstrap bootstrap = VisalloBootstrap.bootstrap(getConfiguration());
+            SearchIndex searchIndex = ((GraphWithSearchIndex) graph).getSearchIndex();
+            Authorizations authorizations = graph.createAuthorizations(VisalloVisibility.SUPER_USER_VISIBILITY_STRING, OntologyRepository.VISIBILITY_STRING);
+            graph.getVerticesWithPrefix(VertexiumOntologyRepository.ID_PREFIX, authorizations).forEach(vertex -> searchIndex.addElement(graph, vertex, authorizations));
```
joeferner: can you add some logging to show progress and give an indication of what it's working on

tools/migrations/predefined-to-dynamic-ontology/src/main/resources/META-INF/services/org.visallo.core.cmdline.CommandLineTool
```diff
@@ -0,0 +1 @@
+org.visallo.tools.migrations.WorkspacesToWorkProduct
```
joeferner: This is the incorrect class

tools/migrations/core/src/main/java/org/visallo/tools/migrations/MigrationBase.java
mwizeman: I just [added a final else to throw an exception](https://github.com/visallo/visallo/pull/1327/commits/4f3c456791de3228dcff66b2535bc087915540bd#diff-f0aa057bba39fa3356a16ddd57a776d5R35) if the value isn't an Integer.

tools/migrations/pom.xml
mwizeman: There are migrations for the enterprise components as well, so we need to be able to run them in the proper order across the projects. Ultimately, I think we need to move off of the one-up Integer approach as it will get harder to keep the version number properly incrementing across repositories.

tools/migrations/predefined-to-dynamic-ontology/src/main/resources/META-INF/services/org.visallo.core.cmdline.CommandLineTool
mwizeman: Fixed. Good catch!

PR #1337, opened 2017-10-03
Command line tool to perform graph checks and validation

joeferner:
- [ ] joeferner
- [x] sfeng88

DO NOT MERGE.  Waiting for the new Vertexium so that the visit logic can be moved into that. I wanted to get this out there so that we can start the discussion on what other things this might check against. Currently it checks...

- Vertices have concept types
- Edges do not have concept types
- Elements and properties have required metadata: modified date, modified by, visibility JSON
- All elements and properties have "visallo" visibility.
- overly connected vertices >5000 warning >10000 error

CHANGELOG
Added: Command line tool to perform graph checks and validation


tools/graph-check/src/main/java/org/visallo/graphCheck/GraphCheck.java
```diff
@@ -0,0 +1,84 @@
...
+        Iterable<Vertex> vertices = getGraph().getVertices(FetchHint.ALL_INCLUDING_HIDDEN, getAuthorizations());
+        visit(vertices, visitor);
+
+        Iterable<Edge> edges = getGraph().getEdges(FetchHint.ALL_INCLUDING_HIDDEN, getAuthorizations());
+        visit(edges, visitor);
+
+        return 0;
+    }
+
+    private void visit(Iterable<? extends Element> elements, VertexiumObjectVisitor visitor) {
```
mwizeman: Do you need this method or could you just use `org.vertexium.Graph.visit`?

tools/graph-check/src/main/java/org/visallo/graphCheck/VertexiumObjectVisitor.java
```diff
@@ -0,0 +1,17 @@
+package org.visallo.graphCheck;
+
+import org.vertexium.*;
+
+public interface VertexiumObjectVisitor {
```
mwizeman: Do we need this interface? This is just a redefinition of `org.vertexium.GraphVisitor `.

tools/graph-check/src/main/java/org/visallo/graphCheck/rules/HasRequiredMetadataGraphCheckRule.java
```diff
@@ -0,0 +1,48 @@
...
+import org.visallo.core.model.properties.VisalloProperties;
+import org.visallo.graphCheck.DefaultGraphCheckRule;
+import org.visallo.graphCheck.GraphCheckContext;
+
+import java.util.Date;
+
+public class HasRequiredMetadataGraphCheckRule extends DefaultGraphCheckRule {
+    @Override
+    public void visitElement(GraphCheckContext ctx, Element element) {
+        Property modifiedBy = element.getProperty(VisalloProperties.MODIFIED_BY.getPropertyName());
```
mwizeman: This "check if property exists" logic seems to be repeated several times here and in the `HasConceptTypeGraphCheckRule.java` class. Maybe a base class method `checkPropertyExists(Element element, String propertyName)` that they can all use?

tools/graph-check/src/main/java/org/visallo/graphCheck/GraphCheck.java
joeferner: This is here until the Vertexium stuff gets released

tools/graph-check/src/main/java/org/visallo/graphCheck/DefaultVertexiumObjectVisitor.java
```diff
@@ -0,0 +1,35 @@
+package org.visallo.graphCheck;
+
+import org.vertexium.*;
+
+public class DefaultVertexiumObjectVisitor implements VertexiumObjectVisitor {
```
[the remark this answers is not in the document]
joeferner: This is here until the Vertexium stuff gets released

tools/graph-check/src/main/java/org/visallo/graphCheck/VertexiumObjectVisitor.java
joeferner: This is here until the Vertexium stuff gets released

joeferner:
@mwizeman The big question is can you think of other checks. Also currently when I run this I get all sorts of errors and warnings on a new Visallo instances. The errors are mostly ontology items not having required properties. Should internal items such as ontology items have required properties? My current thought is yes.

mwizeman:
@joeferner I agree that it would be good to have the modified date and modified by properties on the ontology entities. Now that they can be properly created from the front end, we should know where all of the entities come from. I'm not sure visibility JSON would be helpful at this point, but I don't think it would hurt us either.

Now that you mention it, one other check it that relationships should NOT have a concept type. There will probably be quite a few of those due to https://github.com/visallo/visallo/pull/1264. Maybe we should add a step to the [PredefinedToDynamicOntology](https://github.com/visallo/visallo/blob/master/tools/migrations/predefined-to-dynamic-ontology/src/main/java/org/visallo/tools/migrations/PredefinedToDynamicOntology.java) migration to clean those up?

mwizeman:
@joeferner I've updated this to Vertexium 3.0.0 and switched out the placeholder classes for the Vertexium ones. I also made two changes to the checks:
1. Edges should not have a concept type
2. modified date, modified by, and visibility JSON properties don't need to have modified date and modified by metadata.

tools/graph-check/src/main/java/org/visallo/graphCheck/rules/HasRequiredMetadataGraphCheckRule.java
```diff
@@ -0,0 +1,32 @@
...
+        checkElementHasProperty(ctx, element, VisalloProperties.MODIFIED_BY.getPropertyName());
+        checkElementHasProperty(ctx, element, VisalloProperties.MODIFIED_DATE.getPropertyName());
+        checkElementHasProperty(ctx, element, VisalloProperties.VISIBILITY_JSON.getPropertyName());
+    }
+
+    @Override
+    public void visitProperty(GraphCheckContext ctx, Element element, Property property) {
+        if (!property.getName().equals(VisalloProperties.MODIFIED_BY.getPropertyName()) &&
+                !property.getName().equals(VisalloProperties.MODIFIED_DATE.getPropertyName()) &&
+                !property.getName().equals(VisalloProperties.VISIBILITY_JSON.getPropertyName())) {
```
mwizeman: @joeferner Should the metadata be set on the concept type property? If not, we should add it into the list here before merging this... We're currently [not setting it in the ElementUpdateContect](https://github.com/visallo/visallo/blob/master/core/core/src/main/java/org/visallo/core/model/graph/ElementUpdateContext.java?utf8=%E2%9C%93#L77).

sfeng88: no metadata on concept type :)

PR #1339, opened 2017-10-03
fix possible values not getting populated from saved search

sfeng88:
- [x] joeferner
- [x] sfeng88

Testing Instructions: 
1) Create a saved search with gender: Male
2) Run the saved search
3) Search filters should be populated with Gender Male

Points of Regression:

NO CHANGELOG


kunklejr:
Why does the `setValues` method not actually set any values? It appears to wipe them out with `this.values = []` but then never sets anything back into `this.values`.

joeybrk372:
The values are being set in `this.values` by `self.addValue()` here: https://github.com/visallo/visallo/pull/1339/files#diff-b98b045ac26391cf2365e27af370e6f2L88

PR #1346, opened 2017-10-04
Set thread count to 1 so Visallo processes run by default

nidasid:
- [x] joeferner
- [x] sfeng88

In the web server, Visallo won't run the GPWs because it had a thread count of 0, or unset. 

Testing Instructions: Verify GPWs, ERWs, and LRPs run

Points of Regression: n/a


joeferner:
I think this config file will be deployed in production. In production these should really be disabled.

nidasid:
@joeferner Would you rather these changes be put in here: https://github.com/visallo/visallo/blob/master/config/x-visallo.properties?

joeferner:
I'm thinking here https://github.com/visallo/visallo-lts/blob/master/environments/development/enterprise/config/y-visallo.properties might be better. This config file should only effect development.

mwizeman:
@joeferner This is affecting people who follow the [Getting Started](http://docs.visallo.org/getting-started.html) instructions. Without having the thread count configs set in this directory somewhere, when you clone and run the open source repo the the VisalloProcess instances don't run.

mwizeman:
I just took a look at the distribution and it looks like we exclude x*, y*, and z* files, so moving these changes to x-visallo.properties would keep it out of production.

nidasid:
@sfeng88 @mwizeman @joeferner Done!

PR #1354, opened 2017-10-10
fix newly added dynamic relationship not setting parent to root ontology parent if parent is not specified

sfeng88:
- [ ] joeferner
- [x] sfeng88

Testing Instructions: 
- Make sure you can add comments to a newly created edge that you don't specify a parent for. This will make sure we are adding the root ontology parent correctly.

Points of Regression: N/A

NO CHANGELOG


core/plugins/model-vertexium-inmemory/src/main/java/org/visallo/vertexium/model/ontology/InMemoryOntologyRepository.java
```diff
@@ -526,6 +526,11 @@ public InMemoryRelationship getRelationshipByIRI(String relationshipIRI, String
     }
 
     @Override
+    public Relationship getTopObjectPropertyRelationship(String workspaceId) {
```
mwizeman: This should be implemented in OntologyRepositoryBase.

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/ontology/VertexiumOntologyRepository.java
```diff
@@ -380,6 +379,11 @@ public Concept getParentConcept(final Concept concept, String workspaceId) {
     }
 
     @Override
+    public Relationship getTopObjectPropertyRelationship(String workspaceId) {
```
mwizeman: This should be implemented in OntologyRepositoryBase.

web/web-base/src/main/java/org/visallo/web/routes/ontology/OntologyRelationshipSave.java
```diff
@@ -57,12 +57,14 @@ public OntologyRelationshipSave(
...
 
-        Relationship parent = null;
+        Relationship parent;
         if (parentIri != null) {
             parent = ontologyRepository.getRelationshipByIRI(parentIri, workspaceId);
             if (parent == null) {
                 throw new VisalloException("Unable to load parent relationship with IRI: " + parentIri);
             }
+        } else {
+            parent = ontologyRepository.getTopObjectPropertyRelationship(workspaceId);
```
mwizeman: Instead of doing this here, you could perform this check in `OntologyRepository.getOrCreateRelationshipType` so that things will be wired up properly all of the time. Otherwise, I'd suggest adding a check to `OntologyRepository.getOrCreateRelationshipType` that parent must not be null.

sfeng88:
@mwizeman fixed.

core/core/src/main/java/org/visallo/core/model/ontology/OntologyRepository.java
```diff
@@ -109,6 +109,8 @@
 
     Iterable<Relationship> getRelationshipsByIRI(List<String> relationshipIRIs, String workspaceId);
 
+    Relationship getTopObjectPropertyRelationship(String workspaceId);
```
joeferner: Is there any need to expose this method as public? If not can you remove this and keep it as protected in `OntologyRepositoryBase`

sfeng88: done

PR #1358, opened 2017-10-11
modified ontology selector to return user visible concepts and properties if showHiddenConcepts or showHiddenProperties is set to true

sfeng88:
Add the ability to send a flag so that the component will render properties and concepts set to `userVisible=false`

- [x] joeferner
- [x] sfeng88

Testing Instructions: 
- Make sure when adding new entities, concepts that are marked `userVisible=false` does not show up in list
- Make sure when adding new properties, properties that are marked `userVisible=false` does not show up in list

Points of Regression: N/A

NO CHANGELOG


joeferner:
@sfeng88 Did you mean to commit the yarn.lock files?

sfeng88:
@joeferner yep, I was told to

web/war/src/main/webapp/js/components/ontology/ConceptSelector.jsx
```diff
@@ -89,7 +89,7 @@ define([
                 otherFilters = rest;
                 const showAdmin = userVisible === null;
                 if (showAdmin) {
-                    concepts = ontologySelectors.getConceptsList(state);
+                    concepts = props.showHiddenConcepts ? ontologySelectors.getConceptsList(state) : ontologySelectors.getVisibleConceptsListRootItemHidden(state);
```
joeybrk372: You don't need to make any changes here, because you are giving`showAdminConcepts = true` to the flight wrapper (`conceptSelect`) which will provide the `userVisible = null` filter.

web/war/src/main/webapp/js/components/ontology/PropertySelector.jsx
```diff
@@ -121,7 +121,7 @@ define([
         (state, props) => {
             return {
                 privileges: userSelectors.getPrivileges(state),
-                properties: ontologySelectors.getVisiblePropertiesWithHeaders(state),
+                properties: props.showHiddenProperties ? ontologySelectors.getPropertiesWithHeaders(state) : ontologySelectors.getVisiblePropertiesWithHeaders(state),
```
joeybrk372: This should match conceptSelect by looking at the `otherFilters.uservisible` prop

web/war/src/main/webapp/js/util/ontology/conceptSelect.js
```diff
@@ -120,7 +120,8 @@ define([
                     creatable: this.attr.creatable !== false,
                     value: this.attr.selectedConceptId,
                     placeholder: this.attr.defaultText,
-                    autofocus: this.attr.focus === true
+                    autofocus: this.attr.focus === true,
+                    showHiddenConcepts: this.attr.showHiddenConcepts === true
```
joeybrk372: this is unnecessary because of `showAdminConcepts`

web/war/src/main/webapp/js/util/ontology/propertySelect.js
```diff
@@ -123,7 +123,8 @@ define([
                     value: this.attr.selectedProperty,
                     autofocus: focus === true,
                     creatable: this.attr.creatable !== false,
-                    placeholder
+                    placeholder,
+                    showHiddenProperties: this.attr.showHiddenProperties === true
```
joeybrk372: this is unnecessary because of `showAdminProperties`

web/war/src/main/webapp/js/data/web-worker/store/ontology/selectors.js
```diff
@@ -186,11 +208,14 @@ define(['reselect'], function(reselect) {
 
     const getConceptsList = createSelector([getConcepts], concepts => {
         return _.chain(concepts)
-            .filter(c => _visible(c, { rootItemsHidden: false }))
             .sortBy('fullPath')
             .value()
     })
 
+    const getVisibleConceptsListRootItemHidden = createSelector([getConceptsList], concepts => {
+        return _.filter(concepts, c => _visible(c, { rootItemsHidden: false}));
```
joeybrk372: should be `rootItemsHidden: true`?

sfeng88: I ended up removing this selector since nobody was using it anymore

PR #1364, opened 2017-10-13
Ensure that a properties Visibility matches the VisibilityJSON

mwizeman:
The visibility of the RAW property was being set to default (PUBLIC) but the VisibilityJSON was being set to that of the vertex. This caused all of the extracted properties to have the same issue.

I also changed the `GraphPropertyWorkData.createPropertyMetadata` method to use the property's VisibilityJSON instead of the value from the vertex. Most GPW's that create new properties are doing it based on an existing property. When saving, they use the visibility of the property so the VisibilityJSON needs to match that.

- [x] joeferner
- [x] sfeng88

Testing Instructions:
1. Upload a new file with a visibility set.
2. Ensure that the TEXT property and any other extracted properties are shown as PUBLIC.

CHANGELOG
Changed: GraphPropertyWorkData.createPropertyMetadata will use the VisibilityJSON of the property and not the vertex


joeferner:
@mwizeman could we add a graph check rule for this?

mwizeman:
@joeferner Good idea! I added an enhancement issue to track that.

PR #1365, opened 2017-10-13
Product Ancillary Data + Map Styles support

joeybrk372:
- [x] joeferner
- [x] sfeng88

CHANGELOG
Added: Extension point to modify the styling of non-clustered features/pins on a map product `org.visallo.map.style`.
Added: Product toolbar items are given a `requestUpdate` callback in their props to request the product re-renders.
Added: Added ancillary vertices to work products hidden with workspace visibility that can still behave like a normal vertex in products except for: being selected, searched, published.
Changed: The map's minimum zoom level is now capped to show the whole Earth. Previously you could zoom out to show multiple wrapped images.


core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
```diff
@@ -1218,6 +1222,63 @@ public Product findProductById(String workspaceId, String productId, GetExtended
         return productVertexToProduct(workspaceId, productVertex, includeExtended, extendedData, authorizations, user);
     }
 
+    public WorkProductAncillaryResponse addOrUpdateProductAncillaryVertex(
+            String workspaceId, String productId, String vertexId, User user,
+            UpdateProductEdgeOptions productEdgeOptions,
+            GraphUpdateContext.Update<Vertex> updateVertexFn) {
+        return addOrUpdateProductAncillaryVertex(workspaceId, productId, vertexId, user, productEdgeOptions, updateVertexFn, null);
+    }
+    public WorkProductAncillaryResponse addOrUpdateProductAncillaryVertex(
```
joeferner: new line

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
```diff
@@ -1218,6 +1222,63 @@ public Product findProductById(String workspaceId, String productId, GetExtended
...
+            String sourceGuid) {
+        Authorizations authorizations = getAuthorizationRepository().getGraphAuthorizations(
+                user,
+                VISIBILITY_STRING,
+                VISIBILITY_PRODUCT_STRING,
+                workspaceId
+        );
+        Vertex productVertex = getGraph().getVertex(productId, authorizations);
+        if (productVertex == null) {
+            return null;
```
joeferner: Should this be an exception

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
```diff
@@ -1218,6 +1222,63 @@ public Product findProductById(String workspaceId, String productId, GetExtended
...
+        }
+
+        String kind = WorkspaceProperties.PRODUCT_KIND.getPropertyValue(productVertex);
+        WorkProductService service = getWorkProductServiceByKind(kind);
+        if (!(service instanceof WorkProductServiceHasElements)) throw new VisalloException("Service doesn't support entities");
```
joeferner: Need to come up with coding style rules but we have been using `{ }` on all blocks

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
```diff
@@ -1218,6 +1222,63 @@ public Product findProductById(String workspaceId, String productId, GetExtended
         return productVertexToProduct(workspaceId, productVertex, includeExtended, extendedData, authorizations, user);
     }
 
+    public WorkProductAncillaryResponse addOrUpdateProductAncillaryVertex(
+            String workspaceId, String productId, String vertexId, User user,
```
joeferner: We haven't been great at this in the past but should we make a small POJO to contain all these arguments. I also like to keep user argument towards the end before any labdas

core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
```diff
@@ -598,13 +598,37 @@ public void broadcastLongRunningProcessChange(JSONObject longRunningProcessQueue
         broadcastJson(json);
     }
 
+    public void broadcastWorkProductAncillaryChange(String workProductId, String workspaceId, String ancillaryId, User user, String skipSourceGuid) {
```
joeferner: For some reason the word "ancillary" doesn't sound right. How is it different from extended data. Does it come back when asking for extended data?

core/core/src/main/java/org/visallo/core/model/workspace/product/WorkProductServiceHasElementsBase.java
```diff
@@ -120,13 +122,60 @@ public void cleanUpElements(Graph graph, Vertex productVertex, Authorizations au
         graph.flush();
     }
 
-    protected abstract void populateVertexWithWorkspaceEdge(Edge propertyVertexEdge, TVertex vertex);
+    @Override
+    public GraphUpdateContext.UpdateFuture<Edge> addOrUpdateProductEdgeToAncillaryEntity(GraphUpdateContext ctx, Vertex productVertex, String entityId, UpdateProductEdgeOptions options, Visibility visibility) {
+        if (options == null) options = new UpdateProductEdgeOptions();
```
joeferner: Add `{ }` around any blocks

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
```diff
@@ -1218,6 +1222,63 @@ public Product findProductById(String workspaceId, String productId, GetExtended
...
+                    vertex.getId(),
+                    productEdgeOptions,
+                    workspaceVisibility
+            ).get();
+
+            getWorkQueueRepository().broadcastWorkProductAncillaryChange(
+                productVertex.getId(), workspaceId, vertex.getId(), user, sourceGuid
+            );
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
joeferner: `RuntimeException` -> `VisalloException`

web/plugins/graph-product/src/main/java/org/visallo/web/product/graph/GraphWorkProductService.java
```diff
@@ -515,8 +481,13 @@ protected void updateProductEdge(
         }
     }
 
+
     @Override
-    protected void populateVertexWithWorkspaceEdge(Edge propertyVertexEdge, GraphWorkProductVertex vertex) {
+    public void populateProductVertexWithWorkspaceEdge(Edge propertyVertexEdge, WorkProductVertex vertex2) {
+        super.populateProductVertexWithWorkspaceEdge(propertyVertexEdge, vertex2);
+
+        // FIXME: generics
```
joeferner: This would be nice to fix as part of this PR

core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
jharwig: Basically it's a vertex (not just extended data) that isn't searchable, diff panel inserted, or selectable in products

core/core/src/main/java/org/visallo/core/model/workQueue/WorkQueueRepository.java
```diff
@@ -598,13 +598,37 @@ public void broadcastLongRunningProcessChange(JSONObject longRunningProcessQueue
...
+        JSONObject dataJson = new JSONObject();
+        dataJson.put("id", ancillaryId);
+        dataJson.put("workspaceId", workspaceId);
+        dataJson.put("productId", workProductId);
+        dataJson.putOpt("sourceGuid", skipSourceGuid);
+        json.put("data", dataJson);
+        broadcastJson(json);
+    }
+
+    public void broadcastWorkProductChange(String workProductId, String workspaceId, User user, String skipSourceGuid) {
```
mwizeman: This method should just call the one below it (or the other way around).

core/plugins/model-vertexium/src/main/java/org/visallo/vertexium/model/workspace/VertexiumWorkspaceRepository.java
jharwig: So should all parameters be in a pojo?

2 pojos one update, on add?

joeferner: @mwizeman What are your thoughts on this? I could really go either way on POJO or not POJO (many parameters), I just know that some of our methods are getting a little out of hand (see OntologyRepository for some of the worst offenders). The thing that got me with this particular case was the three strings next to each other.

mwizeman: @joeferner I agree that having many of the same types next to each other gets confusing. I find that I'm constantly looking at the source to make sure that I have them in the correct order. Three doesn't feel like that many, but I agree we need to think about this moving forward and make an effort to do something less error prone.

sfeng88:
@joeferner @mwizeman looks like all of the PR comments have been addressed, can we merge this PR once Travis finishes?

PR #1370, opened 2017-10-17
Clean webapp and upgrade compress plugin for linux builds

jharwig:
- [x] joeferner
- [x] sfeng88

Upgrade compress plugin to fix travis and add clean task

Testing Instructions: Run production build

Points of Regression: None


web/war/pom.xml
```diff
@@ -35,6 +35,20 @@
...
+                <configuration>
+                    <filesets>
+                        <fileset>
+                            <directory>src/main/webapp</directory>
+                            <includes>
+                                <include>libs/</include>
+                                <include>node_modules/</include>
+                                <include>imgc/</include>
+                                <include>jsc/</include>
+                                <include>css/</include>
```
mwizeman: The [makefile](https://github.com/visallo/visallo/blob/master/web/war/src/main/webapp/Makefile) also cleans a directory named `build`. Should that be in this list too? Do we still need the Makefile or is all of its functionality contained in mvn now?

jharwig: Added `build`. The makefile is handy for running without maven while your in the web app dir. It uses the m2 node/yarn, etc

jharwig: Also no other way to run `grunt` with watching enabled. Where the makefile's default target is development watching

PR #1372, opened 2017-10-17
Fix selection after enabling snap to grid

joeybrk372:
- [x] joeferner
- [x] sfeng88

no CHANGELOG


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/worker/snap.js
```diff
@@ -8,7 +8,7 @@ define([], function() {
     function snapPosition(p) {
         return {
             x: snapCoordinate(p.x, GRAPH_SNAP_TO_GRID),
-            y: snapCoordinate(p.y, GRAPH_SNAP_TO_GRID_Y) + (GRAPH_SNAP_TO_GRID_Y - GRAPH_NODE_HEIGHT) / 2
+            y: Math.trunc(snapCoordinate(p.y, GRAPH_SNAP_TO_GRID_Y) + (GRAPH_SNAP_TO_GRID_Y - GRAPH_NODE_HEIGHT) / 2)
```
jharwig: Why not move this to the `snapCoord` function>?

joeybrk372: After a closer look, the constants calculation in the second half of the line `(GRAPH_SNAP_TO_GRID_Y - GRAPH_NODE_HEIGHT) / 2)` was the offender for the fractional value so I moved it. `snapCoordinate` was already returning integer results.

PR #1374, opened 2017-10-18
rename concept to entity type

sfeng88:
- [x] joeferner
- [x] sfeng88

Testing Instructions:
- Make sure create new entity popover shows `Entity Type`
- Make sure histogram show `Entity Type`

Points of Regression: N/A

CHANGELOG
Changed: Rename concept to Entity Type


core/core/src/main/resources/MessageBundle.properties
```diff
@@ -691,15 +691,15 @@ search.sort.placeholder=Add Sort…
 search.filters.related_to=Related To
 search.filters.title_multiple={0} entities
 search.filters.all_concepts=Add Type…
-search.filters.all_concepts.focus=Type to filter concepts…
+search.filters.all_concepts.focus=Type to filter entity type…
```
mwizeman: entity types

sfeng88: done

PR #1376, opened 2017-10-18
Add hover that opens minimap over clusters

jharwig:
- [x] joeferner

Testing Instructions: Open map with clusters, hover over clusters to see elements inside (they do not cluster so if 2 are the same they could be on top of each other.)

Points of Regression: None

CHANGELOG
Added: Hover mini-map popup on clusters that show the pins inside

cluster


joeferner:
What happens if there are 100 items in the cluster?

jharwig:
They might overlap

PR #1379, opened 2017-10-20
Add initialize to product toolbar extension point

jharwig:
- [x] joeferner

Allows plugins to do some initialize work in the product before the user opens the toolbar item.

Testing Instructions: None

Points of Regression: None

NO CHANGELOG


jharwig:
Oops, you're right. I also added a teardown in case we want to attach events

This will ensure the initialize is only called once the injected props are passed and only once until a product change

PR #1380, opened 2017-10-20
Memoize Vertex Titles

mwizeman:
Memoize the vertex titles to prevent having to recompute them with every graph render

- [x] joeferner

CHANGELOG
Fixed: The vertex titles will now only be calculated when the vertex state changes and when a vertex is being hovered instead of on every render.


joeferner:
@mwizeman did you happen to get counts of the title formula calls before and after?

mwizeman:
@joeferner The following numbers are for a graph containing 6 vertices and no edges. The title formula for the vertices is simply `prop('http://visallo.org/broad#name') || ''`

Before this change, the prop method was called the following number of times for each value:
```
conceptType: 36
entityImageUrl: 12
entityImageVertexId: 12
http://visallo.org/broad#name: 12
```

After this change:
```
conceptType: 30
entityImageUrl: 12
entityImageVertexId: 12
http://visallo.org/broad#name: 6
```

The import change is the count for the `http://visallo.org/broad#name` property as it shows that the title formula was only rendered once instead of twice. For many more vertices or title formulas with several calls to prop, the change will be much more noticeable.

PR #1382, opened 2017-10-20
Hide compound properties from search filter selectors

joeybrk372:
- [x] joeferner

Testing Instructions: Search sort/filter by property fields should only show searchable, non-compound properties and dependent properties should be listed.

Points of Regression: Any property creation/selection instances

CHANGELOG
Added: You can pass a boolean `hideCompound` attribute to `components/propertySelect` to hide compound/parent properties from the list of options. (Defaults to false)

### DIFF
--- a/web/war/src/main/webapp/js/components/ontology/PropertySelector.jsx
+++ b/web/war/src/main/webapp/js/components/ontology/PropertySelector.jsx
@@ -25,6 +25,7 @@ define([
             filter: PropTypes.shape({
                 conceptId: PropTypes.string,
                 relationshipId: PropTypes.string,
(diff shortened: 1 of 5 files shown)


joeferner:
I will say this is little confusing. If I'm a user and I see Address on the details pane then I would expect to search for address in the search panel. Also what happens if you click the "i" on Address in the details pane and click search?

joeybrk372:
We currently don't support searching by compound properties. This is fixing a regression that caused dependent properties to not be shown and also hiding all compound properties because we can't rely on `searchable=false`.

If you click the 'i' icon from the inspector for a compound property, it adds filters for each dependent property.

PR #1383, opened 2017-10-20
Add Reindex LRP and CLI, Remove Reindex GPW

joeferner:
- [x] joeferner

CHANGELOG
Added: Reindex LRP and CLI
Removed: Reindex GPW


core/core/src/main/java/org/visallo/core/model/longRunningProcess/ReindexLongRunningProcessWorker.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    protected void processInternal(JSONObject longRunningProcessQueueItem) {
+        ReindexLongRunningProcessQueueItem queueItem = ClientApiConverter.toClientApi(
+                longRunningProcessQueueItem.toString(),
+                ReindexLongRunningProcessQueueItem.class
+        );
+        int batchSize = queueItem.getBatchSize();
+        Range range = new Range(queueItem.getStartId(), queueItem.getEndId());
+        if (queueItem.getElementType() == ElementType.VERTEX) {
+            LOGGER.debug("reindex all vertices");
```
mwizeman: Instead of "all", can you log the batch and range parameters?

core/core/src/main/java/org/visallo/core/model/longRunningProcess/ReindexLongRunningProcessWorker.java
```diff
@@ -0,0 +1,89 @@
...
+            reindexVertices(range, batchSize, authorizations);
+        } else if (queueItem.getElementType() == ElementType.EDGE) {
+            LOGGER.debug("reindex all edges");
```
mwizeman: Instead of "all", can you log the batch and range parameters?

tools/reindex/src/main/java/org/visallo/tools/Reindex.java
```diff
@@ -40,18 +51,93 @@ protected int run() throws Exception {
...
+        try {
+            return getSplitsFromAccumuloGraph(elementType);
+        } catch (Exception ex) {
+            LOGGER.warn("Could not get splits. This will result in one job to reindex all the vertices/edges.", ex);
+            return new ArrayList<>();
+        }
+    }
+
+    private List<String> getSplitsFromAccumuloGraph(ElementType elementType) {
+        AccumuloGraph accumuloGraph = (AccumuloGraph) getGraph();
```
mwizeman: When we talked yesterday, I wasn't considering the fact that this was in open source and we shouldn't be depending on Accumulo for this tool to work. Maybe we can add `getVerticesTableSplits` and `getEdgeTableSplits` to `org.vertexium.Graph`? The SQL and in memory graphs could just return a single range with the max/min vertex ids.

core/core/src/main/java/org/visallo/core/model/longRunningProcess/ReindexLongRunningProcessWorker.java
```diff
@@ -0,0 +1,89 @@
...
+            reindexEdges(range, batchSize, authorizations);
+        } else {
+            throw new VisalloException("Unhandled element type: " + queueItem.getElementType());
+        }
+    }
+
+    public void reindexVertices(Range range, int batchSize, Authorizations authorizations) {
+        Iterable<Vertex> vertices = graph.getVerticesInRange(range, authorizations);
```
mwizeman: Should we provide FetchHints here in case someone has messed around with the `getGraph().getDefaultFetchHints()` settings?

core/core/src/main/java/org/visallo/core/model/longRunningProcess/ReindexLongRunningProcessWorker.java
```diff
@@ -0,0 +1,89 @@
...
+        reindexElements(vertices, batchSize, authorizations);
+    }
+
+    public void reindexEdges(Range range, int batchSize, Authorizations authorizations) {
+        Iterable<Edge> edges = graph.getEdgesInRange(range, authorizations);
```
mwizeman: Should we provide FetchHints here in case someone has messed around with the `getGraph().getDefaultFetchHints()` settings?

tools/reindex/src/main/java/org/visallo/tools/Reindex.java
joeferner: That would work. Technically this doesn't depend on Accumulo, if you aren't using Accumulo this catch will hit https://github.com/visallo/visallo/pull/1383/files#diff-fc32c832ff07a11a725ddcf03834603dR108 and you'll get one big batch. I probably should catch the specific class not found exception instead of all exceptions.

core/core/src/main/java/org/visallo/core/model/longRunningProcess/ReindexLongRunningProcessWorker.java
joeferner: done

core/core/src/main/java/org/visallo/core/model/longRunningProcess/ReindexLongRunningProcessWorker.java
joeferner: done

core/core/src/main/java/org/visallo/core/model/longRunningProcess/ReindexLongRunningProcessWorker.java
joeferner: I went with `FetchHint.ALL` ?

core/core/src/main/java/org/visallo/core/model/longRunningProcess/ReindexLongRunningProcessWorker.java
joeferner: same `FetchHint.ALL`

tools/reindex/src/main/java/org/visallo/tools/Reindex.java
joeferner: I added better exception handling and default it to a range of letters for splits. Can you take a look again.

PR #1385, opened 2017-10-23
Improve Inspector Performance for Large Vertices

mwizeman:
When loading property values for the propRaw method, apply the optional key before transforming and sorting the values.

I created a vertex with 5010 properties and measured how long it took the inspector to open when selecting that vertex. Before these changes it was taking ~253 seconds. After these changes it now takes ~800 milliseconds.

I think there are still some improvements that can help make this perform even better, but this change gets us most of the way.

- [x] joeferner


mwizeman:
@jharwig Done indirectly with https://github.com/visallo/visallo/pull/1385/commits/c9040f19492cdd24bb4ce8f19b5e8e869cbad2f2.

When I was looking into how to test it, I noticed that the `prop` and `props` functions were filtering by key after calling `transformMatchingVertexProperties`. Now that key filtering is handled in `transformMatchingVertexProperties`, the don't need do that. After removing their filtering, I verified that the existing tests pass and will fail if I comment out the new filtering code.

PR #1386, opened 2017-10-23
Compound prop partial values

sfeng88:
- [x] joeferner

Testing Instructions: 
- Running the qa owl, add an address property. You should be able to add the property without having to fill in all of the dependent properties

Points of Regression: N/A

CHANGELOG
Added: Allow compound properties to be added without all dependent properties being filled in

### DIFF
--- a/web/war/src/main/webapp/js/fields/withPropertyField.js
+++ b/web/war/src/main/webapp/js/fields/withPropertyField.js
@@ -142,8 +142,10 @@ define([
                 } else {
                     handle(true);
                 }
-            } else if (!this._markedInvalid) {
+            } else if (result === false || (this._markedInvalid === undefined && self._previousValue)) {
                 handle(false);
+            } else {
+                handle(true);
             }
 
             function handle(isValid) {
--- a/web/war/src/main/webapp/js/fields/withPropertyField.js
+++ b/web/war/src/main/webapp/js/fields/withPropertyField.js
@@ -81,7 +81,7 @@ define([
 
                 self.$node.find('input:not([type=checkbox])').each(function() {
                     var $this = $(this);
-                    if ($this.data('optional') !== true) {
+                    if ($this.data('optional') !== true && self.attr.composite !== true) {
                         $this.attr('required', true)
                     }
                 });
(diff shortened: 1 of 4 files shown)


web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexSetProperty.java
```diff
@@ -169,52 +172,68 @@ public ClientApiVertex handle(
...
     ) {
-        if (valueStr == null && valuesStr != null && valuesStr.length == 1) {
-            valueStr = valuesStr[0];
-        }
-        if (valuesStr == null && valueStr != null) {
-            valuesStr = new String[1];
-            valuesStr[0] = valueStr;
-        }
-
         Object value;
```
joeferner: Can you add a little validation in here. Error is `valueStr` and `valuesStr` is set? Error is both are missing? Error when one is required? Something like that.

Or, possibly better. Split this function out into dependent (taking valuesStr) vs non-dependent (taking valueStr) methods and validate that way.

web/war/src/main/webapp/js/fields/withPropertyField.js
```diff
@@ -140,13 +140,15 @@ define([
...
-            } else if (!this._markedInvalid) {
-                handle(false);
+            } else if (result === false || (this._markedInvalid === undefined && self._previousValue)) {
+                handle(false, self.attr.composite);
+            } else {
+                handle(true, self.attr.composite);
             }
 
-            function handle(isValid) {
+            function handle(isValid, handleCompoundField) {
```
joeybrk372: I would just check `self.attr.composite` inside this function, rather than passing it as an additional argument since we are just reading the value each time.

web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexSetProperty.java
sfeng88: done

joeybrk372:
@joeferner is this pr good to merge?

PR #1390, opened 2017-10-24
Update generate-archetype docs to correct path to cli dir and point browser to correct port

nidasid:
- [x] joeferner

Testing Instructions: Ensure the cli directory is located in `./plugins/cli` assuming you are in the root of the archetype project and navigating to `https://localhost:8443` brings you to Visallo after following instructions on http://docs.visallo.org/versions/master/tutorials/starting.html

Points of Regression: n/a

CHANGELOG
Documentation: Correct path to cli directory and point browser to port 8443


docs/tutorials/starting.md
```diff
@@ -28,9 +28,9 @@ Great!  You have your project set up for development.  At this point, cd into th
 
 Now run ```mvn package```.  We need to do this so that maven will download all of our dependencies.  This could take a while since it needs to download a lot, so grab some coffee or water while it finishes.  Fortunately for us, it will be much faster once all of the dependencies are downloaded.
```
sfeng88: Can we also re-word this line to something like `Run ```mvn package``` to download Visallo dependencies.`?

nidasid: @sfeng88 done

nidasid:
@sfeng88 done

PR #1391, opened 2017-10-24
filter out not addable properties when adding a property

sfeng88:
- [x] joeferner

Testing Instructions:
- Make a property as non addable
- In the inspector, make sure the property doesn't show up in the list
- In the search filter, make sure the property does show up in the filter list

Points of Regression:

NO CHANGELOG


web/war/src/main/webapp/js/components/ontology/PropertySelector.jsx
```diff
@@ -81,6 +81,9 @@ define([
                 if (test && filter && filter.hideCompound) {
                     test = test && !p.dependentPropertyIris;
                 }
+                if (test && filter && filter.addable) {
```
joeybrk372: `addable` should go in `FilterProps` array rather than having a separate if statement

sfeng88: done

PR #1403, opened 2017-10-26
Use @visallo/cytoscape fork

joeybrk372:
- [x] joeferner

Fixed edge rendering issues with  `curve-style=bezier` and `shape=roundRectangle`

CHANGELOG
Fixed: Entities with 2 or more relationships between them would disappear when arranged in certain positions on the graph.


joeferner:
@joeybrk372 Is this temporary? Did we submit a PR?

joeybrk372:
@joeferner This is a temporary substitution until that rendering fix makes it to a released version. Cytoscape currently has a commit for this fix.

PR #1405, opened 2017-10-27
Upgrade Visallo to Elasticsearch 5.5 & log4j2

mwizeman:
As a result of upgrading to Elasticsearch 5.5, this PR also upgrades Visallo to log4j2.

- [x] joeferner

Testing Instructions:
1.) Migrate an existing Visallo installation from Elasticsearch 1.7 to Elasticsearch 5.5
2.) Verify that all expected data is present
3.) Verify search functionality

CHANGELOG
Changed: Upgraded Elasticsearch from version 1.7 to version 5.5
Changed: Upgraded log4j to log4j2
Changed: Term aggregations are no longer supported for properties that are indexed as FULL_TEXT only.


core/core/src/main/java/org/visallo/core/config/ConfigurationLoader.java
```diff
@@ -122,25 +123,25 @@ protected void doConfigureLog4j() {
...
-                    DOMConfigurator.configure(log4jResource);
+                    Configurator.initialize((String)null, null, log4jResource.toURI());
                     log4jLocation = log4jResource.toExternalForm();
                 } else {
-                    throw new VisalloResourceNotFoundException("Could not find log4j.xml on the classpath");
+                    throw new VisalloResourceNotFoundException("Could not find log4j2.xml on the classpath");
                 }
-            } catch (RuntimeException e) {
+            } catch (RuntimeException | URISyntaxException e) {
                 Throwables.propagate(e);
```
joeferner: I'm not sure why this is here but it's deprecated in newer versions of Guava. Can we just leave out this try/catch all together or throw a `VisalloException`

core/plugins/model-vertexium-elasticsearch/src/main/java/org/visallo/vertexium/es/IriIndexSelectionStrategyBase.java
```diff
@@ -2,10 +2,10 @@
 
 import org.elasticsearch.action.admin.indices.stats.IndexStats;
```
joeferner: Since these two classes are deprecated and it looks like that's all that is in these classes can we just remove the whole module altogether. I don't use it.

root/pom.xml
```diff
@@ -63,20 +63,21 @@
         <commons.lang3.version>3.4</commons.lang3.version>
         <commons.httpclient.version>3.1</commons.httpclient.version>
         <curator.version>2.11.0</curator.version>
-        <elasticsearch.version>1.7.5</elasticsearch.version>
+        <elasticsearch.version>5.5.3</elasticsearch.version>
```
joeferner: If I bring in new code and change this back to 1.7.5 until I'm ready what are the chances that it will work?

core/core/src/main/java/org/visallo/core/config/ConfigurationLoader.java
mwizeman: Fixed in https://github.com/visallo/visallo/pull/1405/commits/993941d8a16d1eca53be030fd13a9d66a6ada5b6

core/plugins/model-vertexium-elasticsearch/src/main/java/org/visallo/vertexium/es/IriIndexSelectionStrategyBase.java
mwizeman: Removed with https://github.com/visallo/visallo/pull/1405/commits/993941d8a16d1eca53be030fd13a9d66a6ada5b6. 

I considered removing them too, but left them in since they weren't deprecated in 3.1.4. Since there are already a few breaking changes in this release, I went ahead and removed them.

root/pom.xml
mwizeman: We're going to try this in our test environment and let you know. It looks like it should be pretty simple, but I'd like to try it.

mwizeman: @joeferner After a small tweak to this pom, the test environment came up just fine with ES 1.7.5 instead of 5.5.3.

PR #1407, opened 2017-10-29
expose props function to formula evaluator

sfeng88:
As a user, I would like to ability to pick which multi-value property value to display for the title formula by matching a pattern based on the property values.

- [x] joeferner

Testing Instructions: 
- Add a person entity with 2 title property value of `test` and `test test`
- Go to the admin panel and modify the title formula to be
```
var iri = 'http://visallo.org#title',
    properties = props(iri);
if (properties.length === 0) {
  return '';
}
var test = _.filter(properties, function(p) {return new RegExp("^[a-zA-Z]+$").test(p.value);})
if (test.length > 0) {
    return test[0].value;
} 
return prop(iri) || ‘’;
```  
- the vertex should have the title of `test` and not `test test` on the graph 

Points of Regression: N/A

CHANGELOG
Added: props function to return a list of all matching property values to use in formula evaluator

### DIFF
--- a/docs/getting-started/ontology.md
+++ b/docs/getting-started/ontology.md
@@ -138,6 +138,7 @@ repository.ontology.owl.default.dir=$VISALLO_DIR/config/ontology-sample
         * `pluralDisplayName` The plural display name of type
         * `properties` The property iris defined on this type
     * `prop`: Function that accepts a property IRI and returns the display value.
+    * `props`: Function that accepts a property IRI and returns a list of all matching properties.
     * `propRaw`: Function that accepts a property IRI and returns the raw value.
 
     ```xml
--- a/web/war/src/main/webapp/js/util/vertex/formatters.js
+++ b/web/war/src/main/webapp/js/util/vertex/formatters.js
@@ -970,12 +970,6 @@ define([
             props: function(vertex, name, optionalKey) {
                 checkVertexAndPropertyNameArguments(vertex, name);
 
-                var hasKey = !_.isUndefined(optionalKey);
-
-                if (arguments.length === 3 && !hasKey) {
-                    throw new Error('Undefined key is not allowed. Remove parameter to return all named properties');
-                }
-
                 name = V.propName(name);
 
                 var ontologyProperty = getProperty(name),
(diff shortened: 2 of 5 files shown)


web/war/src/main/webapp/js/util/vertex/formatters.js
```diff
@@ -970,12 +970,6 @@ define([
             props: function(vertex, name, optionalKey) {
                 checkVertexAndPropertyNameArguments(vertex, name);
 
-                var hasKey = !_.isUndefined(optionalKey);
```
sfeng88: Do we need this check here? We check to see if optionalKey is undefined in [transformMatchingVertexProperties](https://github.com/visallo/visallo/blob/master/web/war/src/main/webapp/js/util/vertex/formatters.js#L1314-L1333) again and filter appropriately if it set.

docs/getting-started/ontology.md
```diff
@@ -138,6 +138,7 @@ repository.ontology.owl.default.dir=$VISALLO_DIR/config/ontology-sample
         * `pluralDisplayName` The plural display name of type
         * `properties` The property iris defined on this type
     * `prop`: Function that accepts a property IRI and returns the display value.
+    * `props`: Function that accepts a property IRI and a list of all matching properties.
```
mwizeman: should read "and **returns** a list of all"

web/war/src/main/webapp/js/util/vertex/formatters.js
mwizeman: It doesn't look like we need it there, but may have been trying to provide some information about misusing the method? Not sure... strange that we chose to do that there but not elsewhere.

web/war/src/main/webapp/test/unit/spec/util/vertex/vertexFormattersTest.js
```diff
@@ -400,11 +400,6 @@ define([
 
                 expect(property).to.be.an('array').that.has.property('length').that.equals(0)
             })
-            it('should throw error if key is passed but is undefined', function() {
-                expect(function() {
-                    V.props(vertexFactory(), PROPERTY_NAME_FIRST, undefined);
-                }).to.throw('Undefined key')
-            })
             it('should not throw error if key is passed is empty string', function() {
```
jharwig: Should add a test for forumulas that use `props`

sfeng88:
@jharwig I added a test for using props in the title formula. can you please review again?

PR #1420, opened 2017-11-07
fix users not seeing real time additions of new work product in share…

sfeng88:
- [x] joeferner

Testing Instructions:
```
Test 1:
Load a product then view the network tab. There should only be one request where `product?productId=.

As User A create a case and add a graph product with a vertex.
Share the case to User B
As User A load a different case.
As User B, in a separate session, load the shared case and move the vertex.
As User A check the network panel shows a product?productId=<productId> request with an extendedData: false header
```
```
Test 2:
Have users test01 and test02
Share test01's case with test02 with edit access
As test01 create a graph and add data
As test02 on the same case, add another graph and add data
View available work products as test01, you should see the new graph work product
```

Points of Regression: Switching cases/selecting work products. Product list not updating correctly

NO CHANGELOG


web/war/src/main/webapp/js/data/web-worker/store/product/actions-impl.js
```diff
@@ -57,16 +57,11 @@ define([
 
         changedOnServer: (productId) => (dispatch, getState) => {
             const state = getState();
-            const productWorkspaceId = _.findKey(state.product.workspaces, workspace => productId in workspace.products);
-
-            if (productWorkspaceId) {
```
joeybrk372: This check was preventing clients that receive `workProductChange` messages from requesting the product when they hadn't loaded the case it belongs to yet. Could we add `workspaceId` to the `workProductChange` message and then check for `state.product.workspaces[workspaceId]` here to avoid those requests?

sfeng88: done

PR #1425, opened 2017-11-09
fix adding geolocation property throwing server error

sfeng88:
- [x] joeferner

Testing Instructions:
Add a geolocation property

Points of Regression: N/A

NO CHANGELOG


web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexSetProperty.java
```diff
@@ -193,11 +193,11 @@ public ClientApiVertex handle(
                         user,
                         authorizations);
             } else {
-                if (valueStr == null) {
+                if (valueStr == null && valuesStr == null) {
```
joeferner: Should this be `||`?

sfeng88: we should throw an error if they are both blank. either `valueStr` or `valuesStr` should be filled in

PR #1433, opened 2017-11-16
Map layers

joeybrk372:
- [x] joeferner

CHANGELOG
Added: Work product toolbar item to configure the display and order of layers on map products.
Added: Map work products will now display Geojson and KML files. These files will appear as separate layers on the map.


joeferner:
Other than the conflict is this ready to merge?

PR #1435, opened 2017-11-17
Change selection style of nodes/edges in graph

jharwig:
- [x] joeferner

Allows graph style changes to be set to similar colors as selection
color but still show selection.

Changes the focus style to dim other entities, and changes the find path highligted edges to use similar styles.

Testing Instructions: Change selection in graph, try hovering diff badge and histograms

Points of Regression: None

CHANGELOG
Changed: Graph selection style now uses overlays instead of changing background color.


joeybrk372:
LGTM after fixing tests:
```
[INFO] FAILED TESTS:
[INFO]   productSelectors
[INFO]     ✖ should get the focused elementIds in a product
```

PR #1439, opened 2017-11-20
Don't Use CompositeQuery when performing a "Search Related"

mwizeman:
CompositeQuery doesn't properly handle sorting and computing the total hits when performing a "Search Related" query. This update switched to using the new `hasId` query method in Vertexium to retrieve all of the matching vertices in a single query.

- [x] joeferner

Testing Instructions:
1. Right click on a vertex and select "Search -> Related".
1. Ensure that all vertices related to the clicked vertex are returned.
1. Add a `Relationship Type` filter and ensure that only the matching vertices are returned
1. Add another `Relationship Type` filter that does not match any vertices and ensure that the search results are the same as the previous step.
1. Remove the original `Relationship Type` filter and ensure that no results are found.
1. Select multiple vertices, right click, select "Search -> Related", and repeat the steps above.

CHANGELOG
Fixed: Sorting search results when performing a "Search Related" query.


core/core/src/main/java/org/visallo/core/model/search/VertexiumObjectSearchRunnerBase.java
```diff
@@ -239,44 +239,47 @@ protected void applySortToQuery(QueryAndData queryAndData, SearchOptions searchO
...
+        return getElementTypeFilters("conceptTypes", "conceptType", searchOptions);
+    }
+
+    private Collection<OntologyRepository.ElementTypeFilter> getElementTypeFilters(String parameterName, String legacyParameterName, SearchOptions searchOptions) {
+        String typesStr = searchOptions.getOptionalParameter(parameterName, String.class);
+        if (typesStr != null) {
+            JSONArray types = new JSONArray(typesStr);
+            List<OntologyRepository.ElementTypeFilter> filters = new ArrayList<>(types.length());
+            for (int i = 0; i < types.length(); i++) {
+                JSONObject type = (JSONObject) types.get(i);
```
joeferner: `getJSONObject` instead?

core/core/src/main/java/org/visallo/core/model/search/VertexiumObjectSearchRunnerBase.java
```diff
@@ -239,44 +239,47 @@ protected void applySortToQuery(QueryAndData queryAndData, SearchOptions searchO
...
+                filters.add(filter);
+            }
+            return filters;
         }
-        return filters;
+
+        // Try the legacy parameter
+        String elementType = searchOptions.getOptionalParameter(legacyParameterName, String.class);
+        if (elementType != null) {
+            final Boolean includeChildNodes = searchOptions.getOptionalParameter("includeChildNodes", Boolean.class);
```
joeferner: remove `final`

core/core/src/main/java/org/visallo/core/model/search/VertexiumObjectSearchRunnerBase.java
mwizeman: @joeferner Fixed with https://github.com/visallo/visallo/pull/1439/commits/38c94145c077748f8217073f735a9636019d6e00

core/core/src/main/java/org/visallo/core/model/search/VertexiumObjectSearchRunnerBase.java
mwizeman: @joeferner Fixed with https://github.com/visallo/visallo/pull/1439/commits/38c94145c077748f8217073f735a9636019d6e00

PR #1448, opened 2017-11-28
Any geolocation search support

joeybrk372:
- [x] joeferner

Added support for searching by data type rather than property name, currently geolocation is the only searchable data type filter. Supports `has`, `hasNot`, `intersects`, `disjoint`, `within`, `contains` predicates.

Changed the search to pass filter options to property select in filter items list rather than a (deprecated) list of filtered properties, so we can update the list without removing the additional dataType options provided.

CHANGELOG
Added: Searching by any property with a type of geolocation.
Changed: You can pass additional filters as object data for the `filter` event on a `PropertyList` component.
Added: Added additional filter option `domainType` to `PropertyList` component. You can specify `edge` or `vertex`.


core/core/src/main/java/org/visallo/core/model/search/VertexiumObjectSearchRunnerBase.java
```diff
@@ -319,7 +353,7 @@ private void updateQueryWithFilter(Query graphQuery, JSONObject obj, User user,
                 } else if (PropertyType.BOOLEAN.equals(propertyDataType)) {
                     graphQuery.has(propertyName, Compare.EQUAL, value0);
                 } else if (PropertyType.GEO_LOCATION.equals(propertyDataType)) {
-                    graphQuery.has(propertyName, GeoCompare.WITHIN, value0);
+                    graphQuery.has(propertyName, GeoCompare.INTERSECTS, value0);
```
mwizeman: I need to update this to use the predicate passed from the UI.

core/core/src/main/resources/MessageBundle.properties
```diff
@@ -717,6 +721,9 @@ search.filters.predicates.date.<=Before
 search.filters.predicates.date.>=After
 search.filters.predicates.range=Between
 search.filters.predicates.within=Within
+search.filters.predicates.intersects=Intersects
+search.filters.predicates.disjoint=Disjoint
```
jharwig: _Does not Intersect_ instead of _disjoint_

PR #1450, opened 2017-11-28
Fix find path display when source/target nodes are inside of collapsed nodes

joeybrk372:
- [x] joeferner

CHANGELOG
Fixed: Found paths were not displayed when the source/target entity were inside of a collapsed node.


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -349,19 +349,37 @@ define([
...
+                }
+
+                return mappedIds[id]
+            }
 
             if (data.paths.length > MaxPathsToFocus) {
                 data.paths = data.paths.slice(0, MaxPathsToFocus);
-                $(document).trigger('displayInformation', { message: 'Too many paths to show, will display the first ' + MaxPathsToFocus })
+                $(document).trigger('displayInformation', {
+                    message: 'Too many paths to show, will display the first ' + MaxPathsToFocus
```
jharwig: I know it wasn't before, but can you `i18n`

PR #1451, opened 2017-11-28
Upgrade to React 16.2.0

jharwig:
- [x] joeferner

Most shouldn't have issues as we already did most of the work in `15.6.x`. Table needed the most changes as it's one of the oldest and used an old version of virtualized. We now include that library, so we just have the table use that one and update some deprecated usages. (Should save some downloads.)

Testing Instructions: Umm, all react components? Table, Graph, Map, Products, User Admin, Text resolving popover, user account settings, diff panel, dynamic ontology typeaheads

Points of Regression: None

CHANGELOG
Changed: Upgraded to React 16, plugins that register jsx components need to follow the upgrade path from 15.x to 16.


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/package.json
```diff
@@ -27,15 +27,17 @@
...
     "eslint-plugin-react": "^7.3.0",
     "fast-json-patch": "1.0.0",
     "jest": "^18.1.0",
-    "react": "15.6.1",
-    "react-dom": "15.6.1",
-    "react-test-renderer": "15.6.1",
+    "prop-types": "^15.6.0",
+    "react": "16.2.0",
+    "react-dom": "16.1.1",
+    "react-test-renderer": "16.1.1",
```
joeybrk372: `react-dom` & `react-test-renderer` -> 16.2.0

web/plugins/map-product/src/main/resources/org/visallo/web/product/map/package.json
```diff
@@ -22,8 +22,8 @@
     "eslint": "^4.6.1",
     "eslint-loader": "^1.9.0",
     "eslint-plugin-react": "^7.3.0",
-    "react": "15.6.1",
-    "react-dom": "15.6.1",
+    "react": "16.2.0",
+    "react-dom": "16.1.1",
```
joeybrk372: -> 16.2.0

web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/package.json
jharwig: oops, fixed

web/plugins/map-product/src/main/resources/org/visallo/web/product/map/package.json
jharwig: Removed these as we shouldn't specify them anyways since they are external. Graph runs unit tests so does need them

PR #1461, opened 2017-12-01
Add sort priority to properties to change the order they show up in the inspector

joeferner:
- [x] joeferner

Testing Instructions:
- Add an entity with two different properties
- Set the sort priority of the property with a sort order at the end to a number > 0. Verify the property now appears above the other property.
- Set the sort priority of the property now at the bottom to a value larger than the previous step. Verify this property now shows up higher.
- Finally set both properties back to 0 and verify the original sort order is correct

CHANGELOG
Added: sort priority to properties to change the order they show up in the inspector


sfeng88:
@joeferner looks like there are some backend test failures

joeferner:
pushed a fix, should be working now

PR #1470, opened 2017-12-05
Add dynamic ontology privileges to default user privileges(master)

nidasid:
- [x] joeferner

Testing Instructions: Verify that a new user logging in has the dynamic ontology privileges

Points of Regression: n/a


sfeng88:
@nidasid can you delete the changelog in this PR?

PR #1473, opened 2017-12-06
Teardown all child component instances in inspector

joeybrk372:
- [x] joeferner

The inspector wasn't removing all of its components when removing children. However some components rely on their teardown advice to clean up side effects such as the element list setting element focus on hover.

no CHANGELOG


web/war/src/main/webapp/js/detail/detail.js
```diff
@@ -99,7 +99,7 @@ define([
         };
 
         this.teardownComponents = function() {
-            this.select('detailTypeContentSelector').teardownAllComponents().empty();
+            this.select('detailTypeContentSelector').find('*').teardownAllComponents().empty();
```
jharwig: Should it also teardown all components on the `detailTypeContentSelector` still?

jharwig:
Can you make a 4.0 PR too? and merge both

PR #1481, opened 2017-12-06
Add histogram focusing to map products (4.0)

jharwig:
- [x] joeferner

(Hover of multiple selection inspector histograms or diff panel)

focus

Testing Instructions: Hover of elements and map should reveal those (and dim others.) Should work with clusters and pins. Clusters will show the actual number (1/12) that are focused in side and the total.

Points of Regression: None

CHANGELOG
Added: Map products now will highlight element pins when hovering with mouse over elements in histograms, element lists, publish panel badge, etc.


web/plugins/map-product/src/main/resources/org/visallo/web/product/map/util/cache.js
```diff
@@ -0,0 +1,229 @@
...
+                        context.shadowOffsetX = 0;
+                        context.shadowOffsetY = 0;
+                        context.fill();
+
+                        context.restore();
+                    },
+                    zIndex: 0
+                }))
+            }
+            //if (selected) {
```
joeybrk372: should this commented code be removed?

web/plugins/map-product/src/main/resources/org/visallo/web/product/map/OpenLayers.jsx
```diff
@@ -9,7 +9,6 @@ define([
     createReactClass,
     PropTypes,
     ol,
-    jsonpatch,
```
joeybrk372: need to remove from above array as well

PR #1483, opened 2017-12-07
Histogram should handle PVs with many values better (4.0)

kunklejr:
We don't currently limit how many possible values are shown in the histogram. This PR will make Visallo handle PVs like all other datatypes.

screen shot 2017-12-07 at 10 12 30 am

- [ ] joeferner

Testing Instructions:
Select more than 5 entities on the graph, each of which should have a possible value property set to a different value. The histogram should not show more than five bars for that property.

Points of Regression:

CHANGELOG
Fixed: Fixed handling of possible values in the histogram when there are many possible value types


sfeng88:
@kunklejr can you remove the blank lines in the changelog section please?

kunklejr:
done @sfeng88

PR #1492, opened 2017-12-08
Limit geo predicates (4.0)

joeybrk372:
- [x] joeferner

Limit geo predicates to within and not within for now, since we only do point searching

no CHANGELOG


web/war/src/main/webapp/js/search/filters/filterItem.js
```diff
@@ -20,10 +20,10 @@ define([
     };
 
     const GEO_PREDICATES = {
-        INTERSECTS: 'intersects',
+//        INTERSECTS: 'intersects',
```
joeferner: Remove commented out code

web/war/src/main/webapp/js/search/filters/filterItem.js
```diff
@@ -20,10 +20,10 @@ define([
...
         DISJOINT: 'disjoint',
         WITHIN: 'within',
-        CONTAINS: 'contains'
+//        CONTAINS: 'contains'
```
joeferner: here too

PR #1495, opened 2017-12-08
Fix graph focusing when elements have style extensions that change opacity (4.0)

jharwig:
- [x] joeferner

...to a low number

1. When focusing an element that has opacity < 1, Change to 1
2. When dimming an element with opacity < 0.25, don't increase the
opacity

Testing Instructions: 
1. Focus on an element that has opacity => Element opacity should reset to 1
2. Focus on another element, the first element should stay the same opacity (assuming it's set to less than the default dimming of 0.25%)

Points of Regression: None

CHANGELOG
Fixed: Graph focusing on elements with opacity set


jharwig:
https://github.com/visallo/visallo-lts/issues/3404
https://github.com/visallo/visallo-lts/issues/3403

PR #1497, opened 2017-12-08
Eliminate all usage of HttpSession

kunklejr:
- [x] joeferner

Testing Instructions:
Ensure that the app works as expected with multiple users.

CHANGELOG
Changed: Removed all usage of HttpSession, making the front-end completely stateless. This allows much easier scaling of web server clusters.


web/web-base/src/main/java/org/visallo/web/auth/AuthTokenFilter.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
+        doFilter((HttpServletRequest) servletRequest, (HttpServletResponse) servletResponse, filterChain);
+    }
+
+    public void doFilter(HttpServletRequest request, HttpServletResponse response, FilterChain chain) throws IOException, ServletException {
+        //TODO: request = new NoSessionHttpRequestWrapper(request);
```
joeferner: remove or fix

web/web-base/src/main/java/org/visallo/web/auth/AuthTokenHttpResponse.java
```diff
@@ -0,0 +1,103 @@
...
+        addCookie(tokenCookie);
+    }
+
+    private Date calculateTokenExpiration() {
+        return new Date(System.currentTimeMillis() + (tokenValidityDurationInMinutes * 60000));
+    }
+
+    private boolean isTokenNearingExpiration(AuthToken token) {
+        // nearing expiration if remaining time is less than half the token validity duration
+        return (token.getExpiration().getTime() - System.currentTimeMillis()) < (tokenValidityDurationInMinutes * 30000);
```
joeferner: Can you write `tokenValidityDurationInMinutes * 30000` -> `(tokenValidityDurationInMinutes * 60 * 1000) / 2` to be more clear on the math.

web/web-base/src/main/java/org/visallo/web/auth/AuthTokenWebSocketInterceptor.java
```diff
@@ -0,0 +1,85 @@
...
+    public void configure(AtmosphereConfig config) {
+        String keyPassword = config.getInitParameter(AUTH_TOKEN_PASSWORD);
+        checkNotNull(keyPassword, "AtmosphereConfig init parameter '" + AUTH_TOKEN_PASSWORD + "' was not set.");
+        String keySalt = config.getInitParameter(AUTH_TOKEN_SALT);
+        checkNotNull(keySalt, "AtmosphereConfig init parameter '" + AUTH_TOKEN_SALT + "' was not set.");
+
+        try {
+            tokenSigningKey = AuthToken.generateKey(keyPassword, keySalt);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
joeferner: Change to `VisalloException`

web/web-base/src/main/java/org/visallo/web/auth/AuthTokenWebSocketInterceptor.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public void destroy() {
+        // noop
+    }
+
+    private AuthToken getAuthToken(AtmosphereRequest request) throws AuthTokenException {
+        String cookieString = request.getHeader("cookie");
+
+        if (cookieString != null) {
+            int tokenCookieIndex = cookieString.indexOf(AuthTokenFilter.TOKEN_COOKIE_NAME);
```
joeferner: Can this be replaced with `HttpCookie.parse` (see https://docs.oracle.com/javase/7/docs/api/java/net/HttpCookie.html#parse(java.lang.String))

joeferner:
Does this need the latest webster

web/web-base/src/main/java/org/visallo/web/auth/AuthTokenFilter.java
kunklejr: This is a reminder to fix after we merge the latest version of Webster.

web/web-base/src/main/java/org/visallo/web/auth/AuthTokenWebSocketInterceptor.java
kunklejr: Unfortunately, no. I tried that first, but for whatever reason, `HttpCookie.parse` only returns the first cookie in the list, which we can't guarantee will always be the case.

web/web-base/src/main/java/org/visallo/web/auth/AuthTokenHttpResponse.java
kunklejr: I started that way, but decided to save a few calculations since it gets called a lot. I can change it though if it's too unclear.

kunklejr:
It does need the latest webster to be complete, but I didn't want to introduce such a big change with all the merging we're likely to have after fixing things for 4.0.

web/web-base/src/main/java/org/visallo/web/auth/AuthTokenHttpResponse.java
joeferner: The compiler will optimize it out but the code should read nicer

mwizeman:
@kunklejr Shouldn't there be a matching 4.0 PR or do you plan to cherry pick after this is merged?

kunklejr:
@mwizeman - Not planning this for 4.0. The change is a little bigger than I want to introduce at this stage. It also requires some other changes that aren't done yet before it's complete.

PR #1507, opened 2017-12-11
Fix collapsed nodes not being updated/removed when children were removed but not uncollapsed(4.0)

joeybrk372:
- [x] joeferner

Testing Instructions:
- Create a collapsed node
- Select some/all of it's children and remove them from the graph
- Drag those elements back into the graph / undo the removal

CHANGELOG
Fixed: Removing entities that were part of a collapsed node from a graph and then re-adding them to that product would not place them back inside the collapsed node.


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/webpack.config.js
```diff
@@ -94,7 +94,7 @@ var baseConfig = {
   devtool: 'source-map',
   plugins: [
     new webpack.optimize.UglifyJsPlugin({
-        mangle: true,
+        mangle: false,
```
joeybrk372: We've flipped this option differently in different web plugins, but it should be false. Otherwise it makes debugging a pain and we don't have a good story for detecting dev vs prod mode in these maven -> webpack builds.

jharwig: Okay, that's fine for this PR. But I'm going to make another that mangles for production...

You'll just have to set a `NODE_ENV=development` environment var.

jharwig: https://github.com/visallo/visallo/pull/1513

jharwig: Can you revert this change to avoid the conflict?

PR #1509, opened 2017-12-11
Add right click to geoshape file features(4.0)

joeybrk372:
- [x] joeferner

Also fixed: Hide cluster hover overlay on right click, otherwise it blocks the context menu

no CHANGELOG


web/plugins/map-product/src/main/resources/org/visallo/web/product/map/webpack.config.js
```diff
@@ -73,7 +73,7 @@ var baseConfig = {
   devtool: 'source-map',
   plugins: [
     new webpack.optimize.UglifyJsPlugin({
-        mangle: true,
+        mangle: false,
```
jharwig: revert this for inflight PR

PR #1519, opened 2017-12-13
fix edges erroring when opening in full screen (4.0)

sfeng88:
- [x] joeferner

Testing Instructions:
- open edge in fullscreen
- open elements (vertices + edges) in fullscreen

Points of Regression: N/A

CHANGELOG
Fixed: Edge not opening in full screen when user is authorized to view the data


web/war/src/main/webapp/js/appFullscreenDetails.js
```diff
@@ -213,16 +213,19 @@ define([
...
-                    );
+                    if (v.type === 'vertex') {
+                        descriptors.push(
+                            F.vertex.displayType(v) === 'document' ? '1' : '0'
+                        );
+                        // Sort by title
+                        descriptors.push(F.vertex.title(v).toLowerCase());
+                    } else {
+                        // Sort by title
+                        descriptors.push(F.edge.title(v).toLowerCase());
```
joeybrk372: vertex/edge title are the same function so you don't need to split them out here. can just run `F.vertex.title`

sfeng88: fixed on both PRs

jharwig: we should probably create an `element` alias...

PR #1523, opened 2017-12-13
Graph layout spacing (4.0)

jharwig:
- [x] joeferner

* Fix dagre configuration to include labels
* Animate all layouts
* Fix grid layout differences based on pixelRatio
* Change title truncation to use built-in cytoscape truncating instead
of 3 words. Will now wrap to multiple lines instead of long one line.
  * Tested on collapsed nodes also
* Tested all layouts on pixelRatio=1 & 2 devices give same or very
similar results
* Adjust snap to grid numbers to match grid layout

Testing Instructions: Test graph layouts on browser on hi-dpi monitors and low. Test title truncation/hovering shows wrapped title

Points of Regression: Graph may show less of a title, it's based on width, where before was 3 words.

NO CHANGELOG because the layouts spacing was off in between releases

### DIFF
--- a/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Cytoscape.jsx
+++ b/web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Cytoscape.jsx
@@ -36,15 +36,20 @@ define([
     });
     const LAYOUT_OPTIONS = {
         // Customize layout options
-        random: { padding: 10 },
+        circle: {
+            nodeDimensionsIncludeLabels: true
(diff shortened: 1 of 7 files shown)


joeferner:
@jharwig when you say it will wrap to multiple lines what do you mean? I have some decorations which add text below the vertex title. Will those have issues?

jharwig:
It’s only on mouseover of the node. Although I didn’t test bottom decorations.

sfeng88:
@joeferner are you ok with us merging this?

PR #1532, opened 2017-12-14
Add metadata to Video Transcript & couple of helper methods to ACLProvider (4.0)

sfeng88:
- [x] joeferner

Testing Instructions:
- On Video Transcripts, the "i" popover should show metadata

Points of Regression: N/A

CHANGELOG
Fixed: Modified By, Visibility, Modified Date not displaying in property popover for Video Transcripts


core/core/src/main/java/org/visallo/core/security/ACLProvider.java
```diff
@@ -299,6 +299,45 @@ protected final boolean isComment(String propertyName) {
...
+        }
+        return true;
+    }
+
+    protected final boolean isArtifact(ClientApiElement clientApiElement, OntologyElement ontologyElement) {
+        if (clientApiElement == null) {
+            return false;
+        }
+        if (clientApiElement instanceof ClientApiVertex) {
+            Concept c = (Concept) ontologyElement;
```
mwizeman: Can you pull the body of this `if` into a private method that is shared by both versions of isArtifact?

PR #1534, opened 2017-12-14
Fix aggregation bucket search missing filters (4.0)

joeybrk372:
- [ ] joeferner

When searching for a bucket of a saved search aggregation we provide the search id and bucket filters. Merge these two together instead of overwriting the saved search filter with bucket filter.

Also fixed issue with edge items missing height when inside a saved search card.

CHANGELOG
Fixed: When clicking on a search report with aggregations such as a bar/pie chart will now apply all the correct filters including the range for the slice you selected.
Fixed: After click on a piece of a search report, opening in search will show all the correct filters including the range for the slice you selected.
Fixed: Opening a relationship type search from a report such as a bar/pie chart will correctly flip to a relationship search.


joeybrk372:
ready for review now. I changed this pr to gather the search filters on the front-end rather than back-end because we also needed to pass those to open in search pane

PR #1578, opened 2017-12-22
Clear graph caches on ontology updates (4.0)

joeybrk372:
- [ ] joeferner

CHANGELOG
Fixed: When changing the label of a relationship via the admin panel with a graph open, the relationships will now update immediately.


web/plugins/graph-product/src/main/resources/org/visallo/web/product/graph/Graph.jsx
```diff
@@ -227,10 +227,27 @@ define([
             if (nextProps.registry !== this.props.registry) {
                 memoizeClear();
             }
-            if (nextProps.concepts !== this.props.concepts ||
-                nextProps.relationships !== this.props.relationships) {
-                memoizeClear('vertexToCyNode');
+
+            const clearForOntologyUpdates = [];
+            if (nextProps.concepts !== this.props.concepts) {
```
jharwig: I hate to add more things that cause lots of work, but property changes probably need to clear both nodes (and edges maybe?) because they might have title formula changes

jharwig: probably not edges since we only display the edgeLabel and not any properties though

PR #1593, opened 2017-12-26
Prevent accidental use of HttpSession

kunklejr:
This filter will cause an `UnsupportedOperationException` to be thrown on any attempt to use an `HttpSession` object.

- [x] joeferner

Points of Regression: Any third-party plugins using the `HttpSession` will now break Visallo. They must be rewritten to be stateless.

CHANGELOG
Removed: The use of `javax.servlet.http.HttpSession` is no longer allowed. Components relying on a session must be rewritten to either be stateless or manage their state somewhere else.


web/web-base-test/src/test/java/org/visallo/web/SessionProhibitionFilterTest.java
```diff
@@ -0,0 +1,105 @@
...
+        filter.doFilter(request, response, chain(nextHandler));
+    }
+
+    public FilterChain chain(RequestResponseHandler handler) {
+        HandlerChain handlerChain = new HandlerChain(new RequestResponseHandler[] { handler });
+        return (request, response) -> {
+            try {
+                handler.handle((HttpServletRequest)request, (HttpServletResponse)response, handlerChain);
+            } catch (UnsupportedOperationException e) {
+              throw e;
```
joeferner: code formatting

joeferner:
@kunklejr What is left to support seamless fail-over to other web servers? Are we planning on fixing the front-end to not refresh the page when a web socket disconnects (how will we handle possible missed messages)?

kunklejr:
@joeferner There's one PR left (https://github.com/visallo/visallo-lts/pull/3517) and then I need @jharwig 's help to deal with automatic logouts after a period of inactivity. The plan is to notice inactivity from the front-end and effectively click the logout button in all open windows.

PR #1597, opened 2018-01-02
Have the client handle session expiration

jharwig:
- [x] joeferner

screen shot 2018-01-02 at 10 16 38 am

Allows us to timeout the session based on user activity instead of client network activity. New Configuration for setting timeout `web.ui.auth.token.expiration_minutes=60`. Must be >= 1

* Events that extend timeout: (mouse|key)(down|up)
* When a timeout occurs a new overlay (similar to offline) is displayed with a button to continue using Visallo.
* Client monitors actual token expiration using response header, and sends heartbeat requests as needed so authentication token doesn't expire before user activity should. Otherwise clicking "Continue Session" would sometimes be a 403.
* Expiration Date is saved in local storage, so new windows/tabs can extend the date and not expire all windows if any one is idle.

Testing Instructions: 
Ensure logout screen appears when no mouse/key events after expiration minutes. The warning screen will show 30 seconds before. Try with multiple windows/tabs (same session) not different browser, incognito, those will be new sessions that are independent.

Multiple windows will show the warning at about the same time and clicking continue in any one should close the warning in all and extend the time. Also just loading a new tab of the same session will extend all others (just loading is considered some activity).

Offline overlay should override the session overlay. If the system is offline the session warning won't appear and if the warning is showing and the system goes down the offline overlay will replace the session one.

Points of Regression: Login, Offline Overlay, Logout

CHANGELOG
Changed: Session expiration configuration changed to `web.ui.auth.token.expiration_minutes`.
Fixed: More accurate session expiration that uses user activity instead of requests to server. With new session expiration overlay that allows users to extend their session.


kunklejr:
@jharwig can you please look into the logout issue? Logging out in one window isn't causing other windows in the same session to logout.

jharwig:
Yup, I’ll fix that

jharwig:
@kunklejr done

kunklejr:
@jharwig noticed two things. 

1. Explicitly logging out in one window now logs users in other windows out, but the other windows have an error message that says "Session Expired".
2. Clicking the Continue button in IE does not remove the overlay in other windows and does not stop the other windows from expiring and logging the user out. There may be other IE issues too, but I didn't have enough time to test thoroughly. I will the next time around.

jharwig:
@kunklejr pushed fix to first issue, checking IE now...

jharwig:
@kunklejr Fixed IE. After a storageChange event, we would call getItem and get the old value. Have to wait until after the storage event to get the most recent value.

PR #1604, opened 2018-01-03
Fix Saved Search Aggregation Configuration

jharwig:
- [x] joeferner

The property selection was broken allowing for selection of properties whose datatype doesn't support the aggregation. Switch to whitelist of datatypes for each aggregation type, and fix the filters passed to the updated flight shim to new react selection component.

Testing Instructions: Shouldn't be able to create "Count" or "Histogram" aggregations for geolocations. Geolocations should be selectable for "Geo-Coordinate Cluster"

Points of Regression: Saved Search Card Aggregation Configuration

CHANGELOG
Fixed: Saved search aggregation configuration would allow selection of incompatible properties.


sfeng88:
@jharwig can you cherry-pick this to release-4.0 once its been merged please?

PR #1616, opened 2018-01-08
Fix parsing exception for string list properties

joeybrk372:
- [x] joeferner

Testing Instructions:

Points of Regression: Adding/removing/collapsing of entities on the graph

CHANGELOG
Fixed: When performing an undo of collapsed entities on a graph product, the product would become blank permanently.


core/core/src/main/java/org/visallo/core/model/properties/types/StringListSingleValueVisalloProperty.java
```diff
@@ -16,10 +16,19 @@ public String wrap(List<String> value) {
...
     public List<String> unwrap(Object value) {
         if (value == null) {
             return null;
         }
-        return JSONUtil.toStringList(JSONUtil.parseArray(value.toString()));
+
+        List<String> valueList;
+        if (value instanceof String) {
+            valueList = JSONUtil.toStringList(JSONUtil.parseArray(value.toString()));
+        } else {
```
joeferner: can you change this to `else if (value instanceof List) { ... } else { throw new VisalloException("Could not unwrap type: " + value.getClass().getName()); }

sfeng88:
@joeferner can you re-review this PR when you get a chance please?

sfeng88:
@jharwig looks like there are some test failures in `visallo-web-test`

sfeng88:
@joeferner is this ready to be tested?

joeferner:
@sfeng88 good to go

sfeng88:
LGTM

PR #1619, opened 2018-01-08
Add `Content-Security-Policy` header

jharwig:
- [x] joeferner

Adds header that defines what the page can do as far as loading, evaluating scripts, images, connections, etc. This PR only implements the CSP2 spec which works in Firefox, Chrome. IE will support in next version. Decided not to implement the version 1 spec as it's not recommended to have both headers.

Only adds the header to index page which is the main app and fullscreen.

Adds a report endpoint that just logs an error in log with the violation and rule broken.

https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

Testing Instructions: App should work as normal.

Points of Regression: All web requests

CHANGELOG
Security: Added default `Content-Security-Policy` header and configuration `web.response.header.Content-Security-Policy` to mitigate XSS from plugins and 3rd party dependencies.


joeferner:
What is the for?

jharwig:
No external requirement, just improve front-end security to match latest practices. We don't vet all 3rd party code that we include with npm. So this enforces what we expect the app to do. Like:
* We don't post forms to other domains. Prevents a way of exfiltrating data should we have an XSS vulnerability.
* We don't load scripts from other domain, we allow eval as that's how title formulas work (I just realize we do sometimes use google analytics, so will have to add that to an enterprise config.)
* Prevents clickjacking using frames by not allowing embedding.

If any of these occur, log a violation that we can investigate. Any of these can be changed with configuration depending on requirements if need be.

It's just another tool to minimize bad actors, XSS, CSRF, X-Frame-Options, etc.

jharwig:
https://csp.withgoogle.com/docs/why-csp.html

joeybrk372:
**Unable to load worker scripts in Chrome**
`data-worker.js:316 Refused to load the script  ... it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' 'unsafe-eval'".`

Firefox and IE are working as expected

jharwig:
@joeybrk372 What environment were you testing in?

joeybrk372:
@jharwig I tested on an enterprise test cluster with an incognito window

sfeng88:
@jharwig looks like there are some test failures in `visallo-web-test`

joeybrk372:
LGTM

PR #1620, opened 2018-01-09
Filter unauthorized vertices when requesting highlighted text

joeybrk372:
- [x] joeferner

Points of Regression: setting vertex visibility, resolve form popover

CHANGELOG
Fixed: Text that had been resolved to an element which you are unauthorized to see would still be highlighted although you couldn't perform any actions in the popover. The text is no longer highlighted.


web/web-base/src/main/java/org/visallo/web/routes/vertex/VertexHighlightedText.java
```diff
@@ -85,6 +83,12 @@ public void handle(
                 response.respondWithHtml("");
             } else {
                 Iterable<Vertex> termMentions = termMentionRepository.findByOutVertexAndProperty(artifactVertex.getId(), propertyKey, propertyName, authorizationsWithTermMention);
+                termMentions = StreamUtils.stream(termMentions)
+                        .filter(termMention -> {
+                            String resolvedToVertexId = VisalloProperties.TERM_MENTION_FOR_ELEMENT_ID.getPropertyValue(termMention);
+                            return resolvedToVertexId == null || graph.doesVertexExist(resolvedToVertexId, authorizations);
```
jharwig: Use `doVerticesExist` once which returns a `Map<String id, boolean exists>` instead as it limits the lookups

sfeng88:
After testing, found the following in Chrome:

**Term mentions do not recover**
1. Have 2 users with the following auths
```
User A: Auth A, AuthB
User B: Auth A, Auth C
```
1. Share a case between User A and User B
1. As User A, import `test02.pdf`
1. As User A, resolve `Jeremy Stribling` with no visibility. Note that `Jeremy Stribling` has a dotted underline.
1. As User B, view the shared case: `Jeremy Stribling` is "resolved"
1. As User A, change the visibility of `Jeremy Stribling` to `B`
1. For User B, `Jeremy Stribling` is no longer bolded, but the dotted underline didn't recover.

**Changing Visibility From something the user can't see to something they can see**
1. Have 2 users with the following auths
```
User A: Auth A, AuthB
User B: Auth A, Auth C
```
1. Share a case between User A and User B
1. As User A, import `test02.pdf`
1. As User A, resolve `Daniel Aguayo` with a visibility of `B`
1. As User B, view the shared case: `Daniel Aguayo` is "not-resolved"
1. As User A, change the visibility of `Daniel Aguayo` to `A`
1. For User B, `Daniel Aguayo` is not bolded, but does show up in the diff panel. Expected `Daniel Aguayo` to be resolved.

joeybrk372:
Addressed feedback for B: Term mention vertices are saved with public visibility and resolved edges will follow the visibility of the resolved to vertex. Leaving A for a future milestone because it requires a nontrivial rewrite

core/core/src/main/java/org/visallo/core/model/termMention/TermMentionRepository.java
```diff
@@ -228,6 +243,39 @@ public Vertex findById(String termMentionId, Authorizations authorizations) {
...
+            VisibilityJson visibilityJson = SandboxStatusUtil.getSandboxStatus(edge, workspaceId) != SandboxStatus.PUBLIC
+                ? VisibilityJson.updateVisibilitySourceAndAddWorkspaceId(null, newVisibilitySource, workspaceId)
+                : VisibilityJson.updateVisibilitySourceAndAddWorkspaceId(null, newVisibilitySource, null);
+            Visibility visibility = visibilityTranslator.toVisibilityNoSuperUser(visibilityJson);
+            Visibility defaultVisibility = visibilityTranslator.getDefaultVisibility();
+
+            m.alterElementVisibility(visibility);
+            VisalloProperties.VISIBILITY_JSON.setProperty(m, visibilityJson, defaultVisibility);
+
+            m.save(authorizations);
```
mwizeman: If you switch this `forEach` to `map` and return the result of this `save`, you could do a single flush at the end. After the single flush, you could loop over the edges and push them all. Flushing the graph is pretty expensive.

sfeng88:
LGTM

PR #1621, opened 2018-01-10
Fix propForm being disabled when trying to change element visibility to public

joeybrk372:
- [x] joeferner

Points of Regression: Any property editing/adding in propForm

CHANGELOG
Fixed: The submit button was disabled when trying to change the visibility to public on an element that was protected by a visibility.


sfeng88:
LGTM

PR #1622, opened 2018-01-10
Append acl's to referenced elements in search results

joeybrk372:
- [ ] joeferner

Otherwise UI components that depend on element acl's break because those referenced element results get put into the store.

Points of Regression: components that use `util/acl` (inspector toolbar, property info popover, property form)

CHANGELOG
Fixed: After running a relationship search, if you inspected one of the entities connected to the relationship results, you could not view property infos and the toolbar did not show any options.


sfeng88:
@joeferner can you review this PR when you get a chance?

PR #1623, opened 2018-01-10
Fix map product preview not showing.

jharwig:
- [x] joeferner

Also Removes viewport restoring when switching clients (was only client.) This matches graph behavior as that was removed in 4.0, and was bugging. Added an enhancement to add
back with persisting across session the viewport. The product will simply fit when opened.

Viewport still saves to store for the map search popover.

Points of Regression: Map previews

CHANGELOG
Fixed: Map product previews not showing


jharwig:
Right, forgot about the map search. I'll add the saving to store back. I just won't use the viewport anymore in the product.

sfeng88:
Found the following during testing (seeing the same behavior in all three browser):

** Map Preview not updating **
1. On a new map work product, zoom into the map.
I expected the map work preview to update. (https://cl.ly/420X1K3g1S2l)

** Map Preview doesn't show entities **
I think the map preview use to show entities. Not seeing that behavior anymore.
screen shot 2018-01-15 at 12 52 16 pm

jharwig:
That behavior is expected. We take a screenshot of only what's in view (because that's all that's possible) and we do it when:
1) there isn't already a preview
2) anything has changed

jharwig:
Also for every product switch we fit. We used to save the viewport, but only for that session (not persisted to server).

That was removed because it didn't worked, hurt performance, and made code more complex.

sfeng88:
Everything else looks good except for missing map pins on the preview

screen shot 2018-01-22 at 2 11 18 pm

jharwig:
I'm going to try something new here so we can get an always fitted preview. Previously we had to just take the viewport currently displayed, but If I create a preview map instance with the same layers I can fit the preview map without changing the displayed one.

jharwig:
@joeybrk372 this is ready for re-review in the commit https://github.com/visallo/visallo/pull/1623/commits/f760ef1a40a28e8226741a899e6ed12b67c3794e

joeybrk372:
@jharwig please fix merge conflicts and then put back into `ready for qa`

joeybrk372:
LGTM

PR #1624, opened 2018-01-10
Update propForm when choosing entity to resolve a property on

joeybrk372:
- [x] joeferner

Teardown property selection field in propForm before setting up a new one. Fixes issue when selecting a new vertex the property selection field is not given the new concept type to limit by.

Points of Regression: propForm (adding/editing properties of an element in the inspector), resolving entities

CHANGELOG
Fixed: When resolving text to a property on an entity, the correct entity type and available properties for that entity type are shown.


sfeng88:
LGTM

PR #1626, opened 2018-01-11
Fix product error after case deletion

joeybrk372:
- [x] joeferner

If there is not a current workspace show the ProductDetailEmpty component, fixes issue where an already selected product can error when trying to access the workspace object in between the deletion of a case and selection/creation of a new one

Points of Regression: selection and loading of products, empty product screen

CHANGELOG
Fixed: If you were on a graph work product and only had a single case, deleting the case would cause the product fullscreen view to remain blank.

### DIFF
--- a/web/war/src/main/webapp/js/product/ProductDetail.jsx
+++ b/web/war/src/main/webapp/js/product/ProductDetail.jsx
@@ -13,7 +13,8 @@ define([
             }).isRequired,
             extension: PropTypes.shape({
                 componentPath: PropTypes.string.isRequired
-            }).isRequired
+            }).isRequired,
+            workspace: PropTypes.object
         },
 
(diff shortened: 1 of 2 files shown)


sfeng88:
**Multi default case created**
This might be a different issue, but when deleting the only case I get multiple default cases created per window I am logged in to. For example, I'm logged in to Chrome and FF52 as the same user. In Chrome I delete the only case, I get two new default cases. If you're logged into three different browsers, you get 3 new default cases. https://cl.ly/2w420D1g3n2o & https://cl.ly/0M1Y3T382S1b (only logged into FF52)

@kunklejr thoughts? Maybe this has to do with the new session changes?

jharwig:
Nope. Not session related. The creation is done on all clients with no locking. We need to add some locking and do on server

PR #1627, opened 2018-01-11
Revert session user counts in active user list, but add it to user editor

jharwig:
- [x] joeferner

screen shot 2018-01-11 at 9 49 35 am

Points of Regression: Admin user editor and active list

CHANGELOG
Added: Session counts to admin user editor


sfeng88:
LGTM

PR #1629, opened 2018-01-11
Align word count truncation between graph and inspector

jharwig:
- [x] joeferner

Graph title truncation word count matches element inspector title truncation.

Points of Regression: Graph element titles and inspector

NO CHANGELOG: too minor

### DIFF
--- a/web/war/src/main/webapp/js/detail/item/layoutComponents/vertex.js
+++ b/web/war/src/main/webapp/js/detail/item/layoutComponents/vertex.js
@@ -5,6 +5,8 @@ define([
 ], function(Toolbar, F, ontology) {
     'use strict';
 
+    const MaxTitleWords = 12;
+
     var conceptDisplay = _.compose(_.property('displayName'), F.vertex.concept),
         vertexDisplay = F.vertex.title;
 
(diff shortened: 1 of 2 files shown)


web/war/src/main/webapp/js/detail/item/layoutComponents/vertex.js
```diff
@@ -5,6 +5,8 @@ define([
 ], function(Toolbar, F, ontology) {
     'use strict';
 
+    const MaxTitleWords = 12;
```
joeferner: Should we have a comment here as well saying if you update it here update it in the other place. Or better yet stuff it in config it something.

jharwig: I thought it maybe should be a constant, but not really important enough to be a `WebConfiguration` constant. Also one is the main webapp and one is a plugin...

sfeng88:
Follow the following:

**Title should be truncated in histogram view**
Graph now matches the title in the element inspector. I expected the histogram view to match the truncation on the element inspector as well.

View on Histogram View:
screen shot 2018-01-15 at 4 13 58 pm


View on Element Inspector:
screen shot 2018-01-15 at 4 14 06 pm

jharwig:
The histogram view doesn’t have a max word truncation. Just the available space in a single line. So it should match

jharwig:
Same with element list below...

sfeng88:
Make sense for the element list below, but for the histogram we should truncate some. In the above picture, because the title is so long you can't see the count on the right.

jharwig:
That's a different limitation of the svg styling. We don't want to add a fixed cutoff there

PR #1631, opened 2018-01-12
Fix partial cluster selection on the map

joeybrk372:
- [x] joeferner

Points of Regression: elements on the map selection/focus

CHANGELOG
Fixed: Clicking or shift+clicking a cluster that is partially selected will select all of the cluster's contents.
Fixed: On the map, shift+clicking another cluster/pin while a cluster is partially selected would select all of the cluster's contents.
Changed: Shift+clicking clusters on the map will add them to the selection. Cmd+click or ctrl+click, on Mac or Windows respectively, will toggle the cluster's selection state. This is now consistent with graph behavior.


sfeng88:
LGTM

PR #1632, opened 2018-01-12
Fix diff action selection when users don't have ONTOLOGY_PUBLISH

joeybrk372:
- [x] joeferner

Points of Regression: none

CHANGELOG
Changed: Editors and publishers without ontology publish privileges can still undo any diff that requires a schema change.
Fixed:  Publish All button selected property diffs that required schema changes even if you didn't have the ontology publish privilege, if the entity type was already published.
Fixed: Publish/Undo All buttons not accounting for diffs that required schema changes when deciding if all changes are in the same state.


sfeng88:
LGTM

PR #1638, opened 2018-01-19
Fix race condition when creating a new ontology property in propForm

joeybrk372:
- [ ] joeferner

Points of Regression: Any component requesting `ontology/properties` or `ontology/relationship`

CHANGELOG
Fixed: Property form fields do not appear in IE and inconsistently in Firefox after creating a new property type to add to an element.


web/war/src/main/webapp/js/data/withDataRequestHandler.js
```diff
@@ -131,7 +129,35 @@ define([], function() {
                     )) {
                         const ontologyPath = 'ontology/ontology';
                         const existing = FAST_PASSED[ontologyPath] && FAST_PASSED[ontologyPath].promise;
-                        Promise.resolve(existing || this.refreshOntology()).then(r => {
+                        let ontologyPromise;
+                        if (existing) {
+                            ontologyPromise = new Promise((fulfill) => {
+                                this.on('ontologyUpdated', (event, data) => {
```
jharwig: We need to be careful with `this.off` with just an event name. Especially because this is a mixin for the Data component. If any other mixin, or the data component itself subscribes to that event you remove it here.

It should be rewritten to include the handler like `this.on('ontologyUpdated', handler)`, and `this.off('ontologyUpdated', handler)`

PR #1639, opened 2018-01-19
Fix colors of dashboard maps because IE doesn't support hex alphas

jharwig:
- [x] joeferner

Points of Regression: Map dashboard cards

NO CHANGELOG


sfeng88:
LGTM

PR #1641, opened 2018-01-19
Ensure auth token is sent on redirects

kunklejr:
- [ ] joeferner

We need to ensure that the auth token is set on redirects.


sfeng88:
Tested and working fine. @joeferner please merge once you've review the PR.

PR #1644, opened 2018-01-22
Remove UserStatus concept

kunklejr:
- [x] joeferner

Remove `UserStatus` since it results in incorrect online/offline user counts.

Points of Regression:
- Active User list
- Type-ahead search when sharing workspaces

CHANGELOG
Fixed: Online user counts are now correct in the admin UI.


core/core/src/main/java/org/visallo/core/user/SystemUser.java
```diff
@@ -2,15 +2,14 @@
...
-import org.visallo.web.clientapi.model.UserStatus;
 import org.visallo.web.clientapi.model.UserType;
 
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
 
 public class SystemUser implements User {
-    private static final long serialVersionUID = 1L;
+    private static final long serialVersionUID = 2L;
```
joeferner: I'm not sure if we ever serialize user but in any case I don't think serial version id needs to change since the fields haven't changed.

core/core/src/main/java/org/visallo/core/user/User.java
```diff
@@ -1,15 +1,14 @@
...
-import org.visallo.web.clientapi.model.UserStatus;
 import org.visallo.web.clientapi.model.UserType;
 
 import java.io.Serializable;
 import java.util.Date;
 import java.util.Map;
 
 public interface User extends Serializable {
-    long serialVersionUID = 2L;
+    long serialVersionUID = 3L;
```
joeferner: same here

core/core/src/main/java/org/visallo/core/user/SystemUser.java
kunklejr: You're right. I removed methods related to UserStatus but didn't pay enough attention to notice that didn't change any fields. Nice catch!

mwizeman:
LGTM

PR #1645, opened 2018-01-22
Logout could throw null pointer when no user or invalid user.

jharwig:
- [x] joeferner

Points of Regression: Logout


mwizeman:
LGTM

PR #1646, opened 2018-01-22
Hide visibility field when resolving to existing

jharwig:
- [x] joeferner

We didn't modify the visibility before so don't even show it.

Points of Regression: Resolve terms

CHANGELOG
Fixed: Don't show visibility field when resolving to existing entities (from text) as it was ignored


jharwig:
I don't think we should keep it around. To make it worth keeping around they have to create new, add a visibility, then go back and resolve to existing, then back to create new (possibly with a different title.) I just don't think it's useful or common, and not worth keeping the component mounted.

joeybrk372:
Sorry, I should have elaborated more on my comment. It seems like if you edited the visibility it would update the component state with the value, then if you went back and chose to resolve as existing and resolved that it would (incorrectly) send the modified visibility. Maybe in `onElementSelected` we should also clear `this.state.visibility`?

jharwig:
Oh, right. It does send it, but we never used it. But I can clear it too in case the route changes

joeybrk372:
LGTM

PR #1647, opened 2018-01-22
Add locking to workspace create

jharwig:
- [x] joeferner

...so multiple workspaces can't be created with same name

Points of Regression: Workspace deletion, creation

CHANGELOG
Fixed: Bug that caused multiple cases with same name to be created when deleting all workspaces and multiple sessions connected


jharwig:
Can either @joeferner or @kunklejr review this?

joeferner:
@jharwig one thing, can you add a comment or more info in the commit message to explain why this code exists.

jharwig:
@joeferner done

sfeng88:
LGTM

PR #1648, opened 2018-01-22
Keep track of the last active product per case. Also include option to workspace list route to include products.

joeferner:
- [x] joeferner

CHANGELOG
Added: Keep track of the last active product per case
Added: Include option to workspace list route to include products.


jharwig:
@joeferner I committed the fix to actually make the UI load the active product

joeybrk372:
LGTM

PR #1649, opened 2018-01-22
Upgrade jquery because of vulnerability

jharwig:
- [x] joeferner

* Gridstack needed upgrading after going to 3.0
  * Some css needed to be updated for this
* Change `andSelf` references to `addBack`
* Add backwards compatible functions with console warnings in case
plugins are using those jquery methods

Points of Regression: Dashboard, Search

CHANGELOG
Security: Upgrade to jQuery 3.x as earlier versions had some vulnerabilities.


joeybrk372:
*  date picker is empty when clicking on date property value form fields (e.g. search filter, add property form)

sfeng88:
LGTM

PR #1650, opened 2018-01-22
Don't fit the map until after the initial render

joeybrk372:
- [x] joeferner

no CHANGELOG


mwizeman:
The error is gone but zooming behavior is still not quite right. Followed the instructions in https://github.com/visallo/visallo-lts/issues/3597.

When switching from the graph to the map, the map ended up zoomed in really far. Pressing the "Fit" button zoomed out a little, but still not to the proper distance. Pressing the "Fit" button a third time zoomed to a good level for my text annotaion. Pressing the button additional times had no effect.

jharwig:
Can we close this PR, as I've had to make changes to fit to work with map previews. I'm also addressing the fit already #1623

joeybrk372:
Are those changes not pushed yet? Looking at #1623 currently wouldn't solve this issue because fit could still be called before the map renders. The problem is layers' custom getExtent functions calling map methods that depend on the rendered state

jharwig:
Correct, I haven't pushed

sfeng88:
@joeybrk372 can we delete this branch?

PR #1651, opened 2018-01-23
Initialize extension layers when switching between different maps

joeybrk372:
- [x] joeferner

Initialize extension layers when the map product is changed. This was ok when first loading a map product or switching from a different product kind to a map product, but failed when pointing the existing component to a different map product.

merge after: https://github.com/visallo/visallo/pull/1650

Points of Regression: Map layer UI

no CHANGELOG


joeybrk372:
removed `mapReady` code

sfeng88:
LGTM

PR #1654, opened 2018-01-26
Handle json in string property values

joeybrk372:
- [x] joeferner

* Fix clientApiConverter returning string property values that were valid json strings as json objects
* Fix string input fields not escaping json strings

Points of Regression: visibility viewer, visibility editor, display of visibility in properties list

CHANGELOG
Fixed: Setting a string property value to valid json would break the diff panel and properties list in the element inspector.


jharwig:
LGTM

PR #1656, opened 2018-01-29
Fix saved search table scrolling to the bottom after clicking on new tab/sort

joeybrk372:
- [x] joeferner

Previously, the table would reset it's scroll only on the next render but sometimes there could be multiple before it became scrollable again due to data fetching. Now it listens for the first user scroll input after resetting to the first row.

Points of Regression: scrolling the table normally

CHANGELOG
Fixed: After clicking on a column to change the sort direction of a table, or on a new entity/relationship type tab, the table would automatically scroll down instead of scrolling to the first row.


web/plugins/table/src/main/resources/org/visallo/web/table/js/table/Table.jsx
```diff
@@ -38,6 +38,23 @@ define([
...
+        },
+
+        componentWillReceiveProps(nextProps) {
+            if (nextProps.scrollToTop && !this.state.scrollToTop) {
+                this.setState({ scrollToTop: true });
+            }
+        },
+
+        shouldComponentUpdate(nextProps, nextState) {
+            if (this.state.scrollToTop && !nextState.scrollToTop) {
```
jharwig: What if there are other prop changes besides scrollToTop? They won't be rendered

jharwig:
LGTM

PR #1657, opened 2018-01-30
fix not being able to publish comments on a published entity

sfeng88:
- [x] joeferner

Points of Regression: publishing properties

CHANGELOG
Fixed: Not being able to publish comments on public entities


joeybrk372:
* error trying to publish a comment and entity at the same time, entity+comment are still published: 


* publishing entity publishes comments even if they are not marked to publish

sfeng88:
fixed and added two tests to cover the 2 scenarios

joeybrk372:
LGTM

PR #1658, opened 2018-01-30
Create new react tooltip component

jharwig:
- [x] joeferner

* Uses popper library for positioning
* Hooks into visallo panel sizing
* Use tooltip in justification editor
* Style is implemented by setting the existing bootstrap classes and similar DOM structure, but we could replace that implementation easily.
* Updated the tooltip text in the add/edit property form to differentiate value versus justification fields

tooltip

Points of Regression: Add/edit property form tooltips on value and justification

NO CHANGELOG: Not sure changing the text of tooltip is important enough. Did fix justification field not showing a tooltip


web/war/src/main/webapp/js/components/justification/JustificationEditor.jsx
```diff
@@ -136,21 +143,30 @@ define([
...
-                    className="editing"
-                    onChange={this.onChange}
-                    onPaste={this.onPaste}
-                    placeholder={validation === 'OPTIONAL' ?
-                        i18n('justification.field.placeholder.optional') :
-                        i18n('justification.field.placeholder.required')
-                    }
-                    type="text"
-                    value={justificationText || ''} />
+                <React.Fragment>
```
jharwig: We should update to babel 7 after this release so we can use the `<>` shortcut fragment syntax

mwizeman:
@jharwig I noticed two things when testing this PR:
1. When the tooltip is showing and you resize the panel, the value tooltip and justification tooltip act a little different. The value tooltip fades away as you resize and then fades back in when you stop resizing. The justification tooltip stays visible and moves with the panel while it is resized
2. Compound values (like an address) don't show the "Set Value with Reference" tooltip like other values do.

Neither of these is technically an error, but I wanted to check with you to see if you wanted to make them consistent before merging.

jharwig:
@mwizeman 
1) That's expected, using a different library
2) Let's hold off on the tooltip for compound as I don't think that we support setting the justification on those

PR #1659, opened 2018-01-30
Reload messagebundle files when in devMode

jharwig:
- [x] joeferner

Allows changing of i18n strings without server restart

Points of Regression: Messagebundle strings

CHANGELOG
Changed: MessageBundle resource bundles are reloaded without server restart when configuration: `devMode=true`.


core/core/src/main/java/org/visallo/core/config/VisalloResourceBundleManager.java
```diff
@@ -74,13 +76,28 @@ private Properties getLocaleProperties(Locale locale) {
     }
 
     private ResourceBundle getRootBundle(Locale locale) {
-        return ResourceBundle.getBundle(RESOURCE_BUNDLE_BASE_NAME, locale, new UTF8PropertiesControl());
+        if (devMode == null) {
```
joeferner: Can we move this to the constructor and have the constructor take a `Configuration` and mark the constructor as `@Inject`. I would like to avoid using the `InjectHelper` wherever possible unless it is absolutely needed (circular dependencies).

jharwig: Yeah, I can. I just saw that a lot of usages were `new VisalloResourceBundleManager()`, so those would have to change

jharwig: I could make it a singleton and refactor all those usages

joeferner: Singletons (and in this case InjectHelper) are not fun to test with.

jharwig:
@joeferner done

joeybrk372:
LGTM

PR #1661, opened 2018-01-31
Fix CSP report endpoint, and log help when enabling liveupdate

jharwig:
- [x] joeferner

* Fix json error in violation endpoint that would log exception instead of error log
* Log what configuration is needed if live reload is enabled and fails.

Points of Regression: CSP Violations, Live Reload

NO CHANGELOG new in 4.1


web/war/src/main/webapp/js/cli.js
```diff
@@ -10,8 +10,14 @@ define([], function() {
                 if ('localStorage' in window) {
                     if (enable === true || typeof enable === 'undefined') {
                         console.debug('Enabling LiveReload...')
-                        require([`//${location.host.replace(/:.*$/, '')}:35729/livereload.js`], function() {
+                        const host = location.host.replace(/:.*$/, '');
+                        require([`//${host}}:35729/livereload.js`], function() {
                             console.debug('LiveReload successfully enabled');
+                        }, function() {
+                            console.warn(`Need Content-Security-Policy Updates for live update?
```
joeybrk372: remove "?"

jharwig: Why? An error connecting to live reload could be other issues (like not running grunt). This is probably just the most common

joeybrk372: I expanded it in my head when reading to "You need Content-Security-Policy Updates for live update: ...". Looking again though I think it would be better to be more verbose with the warning: "Failed to enable LiveReload, check that grunt is running and you may also need to update your content security policy: ..."

jharwig: okay, that sounds good

jharwig:
Tester: Should see a warning in console after `enableLiveReload(true)`

And should see a CSP violation log on the server that we loaded a script from *:35729

web/war/src/main/webapp/js/cli.js
```diff
@@ -10,8 +10,14 @@ define([], function() {
                 if ('localStorage' in window) {
                     if (enable === true || typeof enable === 'undefined') {
                         console.debug('Enabling LiveReload...')
-                        require([`//${location.host.replace(/:.*$/, '')}:35729/livereload.js`], function() {
+                        const host = location.host.replace(/:.*$/, '');
+                        require([`//${host}}:35729/livereload.js`], function() {
```
mwizeman: I just tried this and I think you have an extra `}` after the hostname

jharwig:
@mwizeman oops, fixed

mwizeman:
LGTM

PR #1662, opened 2018-02-01
Fix graph undo because position updates aren't dispatched

jharwig:
- [x] joeferner

Points of Regression: Graph undo/redo

NO CHANGELOG broken in between releases


joeybrk372:
LGTM

PR #1664, opened 2018-02-01
Disable add related in read-only case

jharwig:
- [x] joeferner

Both keyboard command alt+r and context menu

Points of Regression: Graph "Add Related" in readonly case

CHANGELOG
Fixed: Disable Add Related items in a read-only case


mwizeman:
LGTM

PR #1666, opened 2018-02-01
Fix heatmap layer selecting when other layers hidden

jharwig:
- [x] joeferner

* Heatmap layer needed a way to set it was hidden by default
* Also adds some more clickable area around the visibility checkbox in layer
list

Points of Regression: Map layer list

NO CHANGELOG heat map added in 4.1


joeybrk372:
LGTM

PR #1667, opened 2018-02-02
Add workspace delete confirmation

jharwig:
- [x] joeferner

Disables the "Are you sure?" button for 1 second after Delete click

confirm

Points of Regression: Case delete

CHANGELOG
Added: Confirmation before case deletion


core/core/src/main/resources/MessageBundle.properties
```diff
@@ -843,6 +843,9 @@ workspaces.form.header=Edit @{alias.case}
 workspaces.form.sharing=Sharing
 workspaces.form.sharing.placeholder=Share with user…
 workspaces.form.button.delete=Delete
+workspaces.form.button.delete.warning=Deleting a @{alias.case.lowerCase} will remove the case and all work products.
```
kunklejr: You should probably use your case alias for the second instance of the term too.

joeybrk372:
LGTM

PR #1668, opened 2018-02-03
Don't create LRP vertices if LRP is started from system user

joeferner:
- [x] joeferner

Points of Regression:
- Create LRPs with both a user and a system user

CHANGELOG
Changed: Don't create LRP vertices if LRP is started from system user


sfeng88:
@joeferner LGTM, but do we need to do any cleanup for all of the LRP vertices that were created when LRPs were enqueued by system users from past releases?

joeferner:
I don't think so, they aren't really hurting anything.

sfeng88:
sounds good.

PR #1671, opened 2018-02-05
fix i18n message not displaying correctly when creating new relationships

sfeng88:
- [x] joeferner

Points of Regression: N/A

NO CHANGELOG


joeybrk372:
LGTM

PR #1672, opened 2018-02-05
add a wrapper for concept type and return THING if concept type is not set

sfeng88:
- [x] joeferner

Points of Regression: graph check tool, diff panel

CHANGELOG
Changed: When a concept type is null, return thing

### DIFF
--- a/core/core/src/main/java/org/visallo/core/model/properties/types/ConceptTypeSingleValueVisalloProperty.java
+++ b/core/core/src/main/java/org/visallo/core/model/properties/types/ConceptTypeSingleValueVisalloProperty.java
@@ -1,0 +1,28 @@
+package org.visallo.core.model.properties.types;
+
+import org.vertexium.Element;
+import org.visallo.core.model.properties.VisalloProperties;
+import org.visallo.web.clientapi.model.ClientApiElement;
+
+public class ConceptTypeSingleValueVisalloProperty extends StringSingleValueVisalloProperty {
+    public ConceptTypeSingleValueVisalloProperty(String key) {
+        super(key);
+    }
+
+    @Override
+    public String getPropertyValue(Element element) {
+        String propertyValue = super.getPropertyValue(element);
+        return propertyValue == null ? VisalloProperties.CONCEPT_TYPE_THING : propertyValue;
+    }
+
+    @Override
+    public String getPropertyValue(ClientApiElement element) {
+        String propertyValue = super.getPropertyValue(element);
+        return propertyValue == null ? VisalloProperties.CONCEPT_TYPE_THING : propertyValue;
+    }
+
+    public boolean hasConceptType(Element element) {
+        String propertyValue = super.getPropertyValue(element);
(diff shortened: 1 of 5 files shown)


core/core/src/main/java/org/visallo/core/model/properties/types/ConceptTypeSingleValueVisalloProperty.java
```diff
@@ -0,0 +1,28 @@
...
+import org.visallo.core.model.properties.VisalloProperties;
+import org.visallo.web.clientapi.model.ClientApiElement;
+
+public class ConceptTypeSingleValueVisalloProperty extends StringSingleValueVisalloProperty {
+    public ConceptTypeSingleValueVisalloProperty(String key) {
+        super(key);
+    }
+
+    @Override
+    public String getPropertyValue(Element element) {
```
joeferner: Can't you just override `unrwap` (instead of these two `getPropertyValue` methods) and if the object is null, return `VisalloProperties.CONCEPT_TYPE_THING`

core/core/src/main/java/org/visallo/core/model/properties/types/ConceptTypeSingleValueVisalloProperty.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    public String getPropertyValue(ClientApiElement element) {
+        String propertyValue = super.getPropertyValue(element);
+        return propertyValue == null ? VisalloProperties.CONCEPT_TYPE_THING : propertyValue;
+    }
+
+    public boolean hasConceptType(Element element) {
+        String propertyValue = super.getPropertyValue(element);
```
joeferner: If you make the change above you'll need to just use `element.getProperty() != null` here.

joeferner:
This scares me a little since it will mask errors of truly missing concept types. Can you explain in the commit message why we are doing this again so that future us will remember by looking at the commit logs.

core/core/src/main/java/org/visallo/core/model/properties/types/ConceptTypeSingleValueVisalloProperty.java
sfeng88: Couldn't just override the `unwrap` method because it doesn't actually get called if the value is null (https://github.com/visallo/visallo/blob/master/core/core/src/main/java/org/visallo/core/model/properties/types/SingleValueVisalloProperty.java#L71)

Two options:
1) Override the `getPropertyValue` methods
2) Change the `getPropertyValue` method in `SingleValueVisalloProperty` to always call the `unwrap` method even if the value is `null`

joeferner: gotcha, I see now.

sfeng88:
@joeferner I added a more detailed description to the commit message to describe why we did this.

PR #1674, opened 2018-02-05
Fix edge property not sorting be name

jharwig:
- [x] joeferner

Points of Regression: Edge property inspector

CHANGELOG
Fixed: Relationship properties sort correctly by name in inspector


sfeng88:
LGTM

PR #1675, opened 2018-02-05
Always create the default workspace even if you have some shared to you

jharwig:
- [ ] joeferner

Points of Regression: workspace creation on login

CHANGELOG
Fixed: Default case is created even if user has some shared cases

### DIFF
--- a/web/web-base/src/main/java/org/visallo/web/routes/user/MeGet.java
+++ b/web/web-base/src/main/java/org/visallo/web/routes/user/MeGet.java
@@ -16,9 +16,7 @@ import org.visallo.webster.handlers.CSRFHandler;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.util.Comparator;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
+import java.util.Optional;
 import java.util.stream.StreamSupport;
 
 @Singleton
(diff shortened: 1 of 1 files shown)


web/web-base/src/main/java/org/visallo/web/routes/user/MeGet.java
```diff
@@ -60,14 +58,16 @@ public ClientApiUser handle(
             Iterable<Workspace> allWorkspaces = workspaceRepository.findAllForUser(user);
             Workspace workspace = null;
             if (allWorkspaces != null) {
-                Map<Boolean, List<Workspace>> userWorkspaces = StreamSupport.stream(allWorkspaces.spliterator(), false)
+                Optional<Workspace> first = StreamSupport.stream(allWorkspaces.spliterator(), false)
```
joeferner: Use the util method `StreamUtil.stream` and static import it so that this line becomes `stream(allWorkspaces)`

web/web-base/src/main/java/org/visallo/web/routes/user/MeGet.java
```diff
@@ -60,14 +58,16 @@ public ClientApiUser handle(
...
+                        .filter(workspace1 -> {
+                            String creator = workspaceRepository.getCreatorUserId(workspace1.getWorkspaceId(), user);
+                            return creator.equals(user.getUserId());
```
joeferner: Can you flip this equals or check for null creator. Non-system users will always have an id but a workspace could be created by the system which would result in a null creator.

web/web-base/src/main/java/org/visallo/web/routes/user/MeGet.java
```diff
@@ -60,14 +58,16 @@ public ClientApiUser handle(
...
+                        })
                         .sorted(Comparator.comparing(w -> w.getDisplayTitle().toLowerCase()))
-                        .collect(Collectors.partitioningBy(userWorkspace ->
-                                workspaceRepository.getCreatorUserId(userWorkspace.getWorkspaceId(), user).equals(user.getUserId())));
+                        .findFirst();
 
-                List<Workspace> workspaces = userWorkspaces.get(true).isEmpty() ? userWorkspaces.get(false) : userWorkspaces.get(true);
-                if (!workspaces.isEmpty()) {
-                    workspace = workspaces.get(0);
+                if (first.isPresent()) {
```
joeferner: `if` Could probably be replaced with `orElse(null)`

jharwig:
@joeferner made those changes

PR #1676, opened 2018-02-05
Use "Untitled" instead of "No title available"

jharwig:
- [ ] joeferner

Points of Regression: untitled vertices


sfeng88:
Still seeing `No Title Available` when creating new entities

PR #1679, opened 2018-02-06
Add support for map layer configuration components + base layer provider configuration

joeybrk372:
- [ ] joeferner

* Added support for layers to specify configuration components. When creating a new layer you can provide a `config` object in the layer's properties e.g:
```js 
new ol.layer.*({
    ...options,
    config: {
        height: Number // height of the container the config component will be rendered in,defaults to 40
        componentPath: String
    }
})
```

map-layer-satellite2

Points of Regression: rendering of layers, toggling visibility of layers

CHANGELOG
Added: Configuration can now specify multiple `web.ui.map.provider.<providerName>` for the same source. This lets you, for example, switch between a street and satellite imagery set from a single tile provider. You must specify a `web.ui.map.provider.<providerName>.<source>` in this case.
Added: The base map layer will provide a dropdown to toggle a product's map provider based on configured providers.
Fixed: Activity popover had boxy corners behind it's rounded edges.


web/plugins/map-product/src/main/resources/org/visallo/web/product/map/layers/BaseLayerConfig.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+            onUpdateLayerConfig(nextConfig, layer);
+        }
+    });
+
+    return redux.connect(
+        (state, props) => {
+            const properties = state.configuration.properties;
+            const defaultProvider = properties['map.provider'] || 'osm';
+            const mapProviders = _.chain(properties)
```
jharwig: I'd move this into the connected component and just pass configuration. doing it here means that mapProviders is always a new object and will always force re-render even if `properties` never changes.

Best yet, create a configuration selectors file to do this for you

web/plugins/map-product/src/main/resources/org/visallo/web/product/map/messages.properties
```diff
@@ -6,4 +6,6 @@ org.visallo.web.product.map.MapWorkProduct.layer.no.title=No title available
 org.visallo.web.product.map.MapWorkProduct.layer.error=Unable to render layer
 org.visallo.web.product.map.MapWorkProduct.layer.error.data.format=Unable to render file with incorrect or unsupported format
 org.visallo.web.product.map.MapWorkProduct.layers.empty=No layers available
-org.visallo.web.product.map.MapWorkProduct.layers.sort.help=Drag to order layers
\ No newline at end of file
+org.visallo.web.product.map.MapWorkProduct.layers.sort.help=Drag to order layers
+
+org.visallo.web.product.map.MapWorkProduct.layers.base.provider=Map Provider
```
jharwig: Don't think we should present the configuration property name to the user. Just `Base Layer` or something

joeybrk372: The layer name "Base" is right above it so I wouldn't want "Base \n Base Layer". I could change the wording to "Source" or maybe remove the select label altogether. Which do you think is better?

jharwig: Just remove it

PR #1680, opened 2018-02-06
Allow multiple values in structured ingest

jharwig:
- [x] joeferner

Each property key is a hash of the value

* Fix initial width to minimum level
* Change unique identifier wording if appending
* Fix some js errors
* Fix multiple unnecessary renders

Points of Regression: Structured import

CHANGELOG
Added: Structured file import (CSV,Excel) can now import multiple values of the same property for an entity
Changed: For structured import with few columns, make the modal window big enough to work with


joeybrk372:
LGTM

PR #1681, opened 2018-02-06
Add default error handler to web requests

joeybrk372:
- [x] joeferner

Makes remote debugging a little easier by returning stack traces in dev mode

Points of Regression: any web requests

no CHANGELOG


web/web-base/src/main/java/org/visallo/web/VisalloDefaultErrorHandler.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.IOException;
+
+public class VisalloDefaultErrorHandler extends HttpServlet {
+    public static final String GENERIC_ERROR_MESSAGE = "Unable to process request. Please contact your system administrator for more details.";
+
+    private boolean devMode;
+
+    @Override
+    public void init() {
+        final Injector injector = (Injector) getServletContext().getAttribute(Injector.class.getName());
```
mwizeman: Is it possible to us [InjectHelper#getInstance](https://github.com/visallo/visallo/blob/master/core/core/src/main/java/org/visallo/core/bootstrap/InjectHelper.java#L49) here instead of looking the injector up in the context?

sfeng88:
LGTM

PR #1683, opened 2018-02-07
Fix string property filters setting '[]' as an initial value

joeybrk372:
- [x] joeferner

Points of Regression: property fields when adding/editing values, search filter properties (manual and loading from saved search)

CHANGELOG
Fixed: Adding a string property filter in search would start with an initial value of '[]' instead of being empty.
Fixed: Adding a property with possible values as a search property filter would add an empty value filter immediately.


sfeng88:
LGTM

PR #1684, opened 2018-02-08
Save structured ingest mapping

joeybrk372:
- [ ] joeferner

* Saves the current mapping per sheet when starting ingest as a property on the structured file vertex.
* Reset button clears the current mapping for the sheet (client-side only)

CHANGELOG
Added: Structured file import saves the last import configuration used for the file.


sfeng88:
```
On the preview view, the `will create` column doesn't update when you change a mapping
1) Drag and drop coffee.csv
2) Map column 1 to Person concept and title property
3) Import
4) Reset the mapping and change column 1 to Company and city property

Will create column shows 0 to create -> Should it show 3?
```

```
Remapping doesn't work unless you click the reset button first
1) Drag and drop coffee.csv
2) Map column 1 to Person concept and title property
3) Import
4) Remap column 1 to a company concept and city property without clicking the reset button
If you click import, you get 3 person and 3 company vertices
```
screen shot 2018-02-14 at 4 53 58 pm

Should we save on preview rather than ingest?

PR #1686, opened 2018-02-09
Don't clear search hits count when closing the search pane

joeybrk372:
- [x] joeferner

CHANGELOG
Fixed: Number of search results would disappear when closing and reopening the search pane, but the results would still be displayed.


sfeng88:
LGTM

PR #1688, opened 2018-02-09
Don't create detail previews for vertices inside collapsed nodes

joeybrk372:
- [x] joeferner

CHANGELOG
Fixed: If you showed vertex previews with alt+P while collapsed nodes were in your selection, the previews for the contents of the collapsed nodes would all appear at the top left corner of the screen.


sfeng88:
LGTM

PR #1690, opened 2018-02-10
Fixes multiple iteration over same vertexium query.

joeferner:
- [x] joeferner

This fixes two issues. Calling iterator on a vertexium object will cause multiple queries on the data source. Calling iterator on a vertexium object can cause a Accumulo not closed exception.

Testing:
Search for edges. Before this change you will occasionally get an Accumulo not closed exception.


sfeng88:
LGTM

PR #1691, opened 2018-02-12
Cache ontology properties by IRI

joeferner:
- [x] joeferner

Fixes performance issues with repeated calls to OntologyRepositoryBase#getPropertyByIRI. One particular offender is PropertyValueHelper#getValueAsDisplayString called from CSV and excel export.

Points of Regression:
- Run CSV export on 10k rows with and without the PR. Performance is greatly enhanced.
- Test ontology related operations (dynamic) and ensure the cache is flushed properly


joeferner:
Yeah I was going to set an expire time on the property cache but in the end didn't really see the need. This PR should be good assuming what is there makes sense. I'll remove the cache changes from the other PR.

sfeng88:
LGTM 

Testing on the same dataset, exporting about ~7k rows
Without PR: 1.39 mins
With PR: 48 seconds

PR #1692, opened 2018-02-12
Upgrade to Vertexium 3.1.0

joeferner:
- [x] joeferner

Points of Regression:
- Be sure to install root/pom.xml
- Big change comes with aggregation on exact matches and search of exact matches with case insensitive searching.
- See https://github.com/visallo/vertexium/releases/tag/vertexium-root-3.1.0 for other changes 

CHANGELOG
Changed: Upgrade to Vertexium 3.1.0


core/core/src/main/java/org/visallo/core/cache/CacheOptions.java
```diff
@@ -1,7 +1,11 @@
 package org.visallo.core.cache;
 
+import java.util.concurrent.TimeUnit;
+
 public class CacheOptions {
     private Long maximumSize;
+    private Long expireAfterAccessDuration;
```
mwizeman: Did you mean for this change to be part of this PR?

joeferner: No, those must have made their way in when I was trying to fix the ontology cache. I'll remove them tonight.

joeferner:
Needed to make the following change to Vertexium... https://github.com/visallo/vertexium/commit/64db7c34e686ac72c6730489fc093dbbdf12d1aa

joeybrk372:
LGTM

PR #1685, opened 2018-02-08
Require edit privilege to perform structured file import

joeybrk372:
- [x] joeferner
- [x] mwizeman joeybrk372 jharwig sfeng88

CHANGELOG
Fixed: Users without edit privileges could still import structured files.

PR #1694, opened 2018-02-13
Auto publish structured file

sfeng88:
- [ ] joeferner

Merge after: https://github.com/visallo/visallo/pull/1684

Points of Regression: Structure Ingest, Publishing with dynamic ontology

CHANGELOG
Added: 
Changed: Publishing to fail on first error if dynamic ontology publishing fails
Documentation: 
Deprecated:
Removed:
Fixed:
Security:


web/plugins/structured-ingest/core/src/main/java/org/visallo/web/structuredingest/core/routes/Ingest.java
```diff
@@ -161,6 +164,10 @@ public void finishedRow(long row, long totalRows) {
         parserHandler.dryRun = true;
         ParseOptions parseOptions = new ParseOptions(optionsJson);
 
+        boolean canPublishImmediately = privilegeRepository.hasPrivilege(user, Privilege.ONTOLOGY_ADD)
```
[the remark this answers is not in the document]
sfeng88: I didn't check it here because this only get called after the ontology has already been added. I can add a check for `ONTOLOGY_ADD` here too

sfeng88:
@mwizeman done making the changes! good catch

web/plugins/structured-ingest/core/src/main/java/org/visallo/web/structuredingest/core/routes/Ingest.java
```diff
@@ -184,4 +191,43 @@ private void parse(Vertex vertex, StreamingPropertyValue rawPropertyValue, Parse
             structuredIngestParser.ingest(in, parseOptions, parserHandler);
         }
     }
+
+    private boolean canPublishImmediately(ParseMapping parseMapping,
+                                          User user,
+                                          String workspaceId
+    ) {
+        boolean canPublishOntology = privilegeRepository.hasAllPrivileges(user, Sets.newHashSet(Privilege.ONTOLOGY_PUBLISH, Privilege.ONTOLOGY_ADD));
```
joeferner: Should we also check `Privilege.PUBLISH` and `Privilege.EDIT` here or are these handled elsewhere.

sfeng88: Users without edit privileges can't perform structured file import (https://github.com/visallo/visallo/pull/1685/files)

If the user doesn't have `Privilege.PUBLISH` privilege the front end won't give the user to publish immediately

joeybrk372:
Undoing first import then publishing second blocks entities from appearing in case
1) Import some entities without publishing immediately
2) Undo all of the imported entities
3) Import different entities but publish immediately
The new entities are imported and published but you can't see them on the case you imported them from

sfeng88:
@joeybrk372 I created an issue for your comment above. The fix required a vertexium change. Lets go ahead and merge this PR.

PR #1695, opened 2018-02-13
fix mimetype ontology mapper test not running

sfeng88:
- [x] joeferner

Points of Regression: mime type ontology mapper

NO CHANGELOG


joeybrk372:
LGTM

PR #1696, opened 2018-02-13
fix visibility json not returning to the front end as a json object

sfeng88:
- [x] joeferner

Points of Regression: setting visibility and visibility viewer

NO CHANGELOG


joeybrk372:
LGTM

PR #1697, opened 2018-02-15
Allow ontology formula's to use F. functions

joeferner:
- [x] joeferner

Points of Regression:
Test formulas with F. functions on the front end and back end (unit test provided)

CHANGELOG
Added: Allow ontology formula's to use F. functions


joeybrk372:
LGTM

PR #1698, opened 2018-02-21
Api Tokens and Ontology Listeners

mwizeman:
- [ ] joeferner

In support of the enterprise GraphQL plugin, the following features were added:
1. Authentication via API tokens in addition to web tokens. API tokens must be passed as request headers. They are also stored on the Visallo user object, which means that they can be revoked simply by deleting them from the user.
2. Allow multi-value properties on user objects in order to store multiple API tokens on a user.
3. Added the notions of cache listener and ontology listener to support plugins that need to track ontology changes.

Points of Regression:
Authentication and Ontology

CHANGELOG
Added: CacheListener interface to allow interested code to be notified when CacheRepository keys are invalidated.
Added: OntologyListener interface to allow interested code to be notified when the OntologyRepository cache is cleared.
Added: API tokens can now be stored on a Visallo User and passed as request headers for authentication.
Changed: Visallo User objects now support storing multi-valued properties


core/core/src/main/java/org/visallo/core/user/SystemUser.java
```diff
@@ -102,7 +102,17 @@ public Object getProperty(String propertyName) {
...
 
     @Override
-    public Map<String, Object> getCustomProperties() {
+    public <PROP_TYPE> PROP_TYPE getProperty(String key, String propertyName) {
+        return null;
+    }
+
+    @Override
+    public <PROP_TYPE> Map<String, PROP_TYPE> getProperties(String propertyName) {
+        return null;
```
joeferner: Should this method return a new empty hash map? Would make it easier on the caller and prevent NPEs.

core/core/src/main/java/org/visallo/core/user/SystemUser.java
```diff
@@ -102,7 +102,17 @@ public Object getProperty(String propertyName) {
     }
 
     @Override
-    public Map<String, Object> getCustomProperties() {
+    public <PROP_TYPE> PROP_TYPE getProperty(String key, String propertyName) {
```
joeferner: Can you change all the `PROP_TYPE` to `TPropType`

core/core/src/main/java/org/visallo/core/user/SystemUser.java
```diff
@@ -102,7 +102,17 @@ public Object getProperty(String propertyName) {
...
+        return null;
+    }
+
+    @Override
+    public <PROP_TYPE> Map<String, PROP_TYPE> getProperties(String propertyName) {
+        return null;
+    }
+
+    @Override
+    public Map<String, Map<String, Object>> getCustomProperties() {
```
joeferner: I use this to store strings on my project. Any reason for the return type change? I believe custom properties will be passed to the UI so things like passwords and other sensitive information should be kept out of here.

core/core/src/main/java/org/visallo/core/user/SystemUser.java
mwizeman: Fixed: https://github.com/visallo/visallo/pull/1698/commits/968f9f72c2da08bfec5865acf872c6f43742d36d

core/core/src/main/java/org/visallo/core/user/SystemUser.java
mwizeman: Fixed: https://github.com/visallo/visallo/pull/1698/commits/968f9f72c2da08bfec5865acf872c6f43742d36d

joeferner:
Should also this be protected with a privilege as well. If I were to turn this on, on my project I would only want certain types of users to be able to do this.

core/core/src/main/java/org/visallo/core/user/SystemUser.java
mwizeman: Previously, the custom user properties only supported a single value and the map that was returned was indexed by property name. In order to store multiple API tokens on a User, I updated the User to support multiple values. The outer map is indexed by property name and the inner map is indexed by property key.

I worked with Jason to ensure that the UI doesn't use that portion of the user response. I also tested the UI manually to ensure that I didn't cause any errors.

@joeferner If it will impact your project, an alternative to changing this method is I could store the API tokens as a single JSON encoded property. That wouldn't be a difficult adjustment for me to make and there would be far fewer changes. Let me know.

mwizeman:
@joeferner These are just the infrastructure changes to support authenticating with API tokens. Without the API plugin installed the user's wouldn't have any means to actually create API tokens.

That being said, even if they could the API token authenticates as that user and wouldn't have any privileges above what they normally have.

core/core/src/main/java/org/visallo/core/user/SystemUser.java
```diff
@@ -102,7 +103,17 @@ public Object getProperty(String propertyName) {
     }
 
     @Override
-    public Map<String, Object> getCustomProperties() {
+    public <TPROP_TYPE> TPROP_TYPE getProperty(String key, String propertyName) {
```
joeferner: I was thinking `<TPropType>` or to follow Java's naming convention just `<T>`

joeferner:
Right makes sense, I was looking at the other PR at the same time.

core/core/src/main/java/org/visallo/core/user/SystemUser.java
mwizeman: I understand now. Fixed: https://github.com/visallo/visallo/pull/1698/commits/7ac7ed5593ccc63a8584840e57bc00d28896a116

core/core/src/main/java/org/visallo/core/user/SystemUser.java
joeferner: It looks like this should be fine, it doesn't look like it changes how the data gets stored just how it's returned. All my users in the system have custom properties I need to retain.

PR #1699, opened 2018-02-22
Allow react components for org.visallo.visibility extension + add new visibility selector as default editor

joeybrk372:
- [ ] joeferner

Points of Regression: any places with visibility viewing/editing (properties, comments, diffs, create connection, structured import, term resolution, 

CHANGELOG
Added: `org.visallo.visibility` extension point now supports React components.
Changed: The default visibility editor is now a selection dropdown that lists the current user's authorizations.


docs/extension-points/front-end/visibility/index.md
```diff
@@ -2,56 +2,35 @@
```
joeybrk372: This is not complete, will update with links once https://github.com/visallo/doc-examples/pull/4 is merged

web/war/src/main/webapp/js/util/component/attacher.js
```diff
@@ -44,9 +44,6 @@ define([
             throw new Error('Valid component or path is required')
         }
         if (this._params) {
-            if ('visalloApi' in this._params) {
```
joeybrk372: Removed this check because we're using visibility/util to attach components from Flight parents as well as directly using VisibilityEditor VisibilityViewer to attach from React parents. This errors because the first attach passes those props to the second attach.

